PR #23, opened 2015-10-14
Fix error in synchronisation logic

### LINKED ISSUE #22: Problem with syncing record changes in v0.10.0
After updating to v0.10.0 Padlock Cloud synchronization works only when new records are created.
Deleting and updating records (adding/removing/updating fields) actions are not synchronized.
This problem is manifesting both ways on following setup:
Chrome app on PC (Win8.1, Chrome 45.0.2454.101) with iPhone app (iOS 9.0.2).

MaKleSoft:
Fixes #22

### DIFF
--- a/src/Collection.js
+++ b/src/Collection.js
@@ -82,7 +82,7 @@ padlock.Collection = (function(util) {
          * @param {Object}  rec A record object or an array of record objects to be added to the collection
          */
         add: function(rec) {
-            var records = [];
+            var newRecords = [];
 
             rec = util.isArray(rec) ? rec : [rec];
             rec.forEach(function(r) {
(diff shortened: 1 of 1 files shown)


MaKleSoft:
@inikolic This should fix #22 . Would you be so nice and verify this?

inikolic:
how to check? update app, or I could run fixed version online?
I've tried with current installed apps, but bug is still occurring.

MaKleSoft:
@inikolic Oh, sorry, I was just assuming you had a development version running locally. Never mind... I've tested it locally and it seems to solve the problem.

inikolic:
@MaKleSoft Np, I've been in IT development for 20 years but got overburned and 18 mths ago changed career. So now I'm just a simple user (with some insight into the process).   ;-)

PR #26, opened 2015-10-30
Add `bower install` to `postinstall` script

lewiscowper:
This PR makes use of the https://docs.npmjs.com/cli/run-script property in your `package.json`, and now whenever someone runs `npm install` in the project directory, once it is completed, `bower install` will be run as well. This means we can go from git clone to installed project in only two steps, instead of 3, and is part of a longer series of PRs I intend to propose to make getting set up easier, and less reliant on external dependencies outside of node and npm.

Stage 1 of #27

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,6 @@ Install these if you haven't yet:
 Now from inside the project folder, install the local requirements:
 
     npm install
-    bower install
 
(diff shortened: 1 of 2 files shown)


MaKleSoft:
This only works if bower is already installed (globally or locally) which can only be assumed if bower is added to the npm dependencies. See my comment in https://github.com/MaKleSoft/padlock/pull/29 on why I think this a bad idea.

Issue #27, opened 2015-10-30
Build, install and serve processes relies on external dependencies.


lewiscowper:
The sooner we can allow people to go from git clone, to running the software, the more likely we are to bring people on board with the project.

I'm more than happy to put in the work to close this issue, making more use of npm scripts, which allow us to define a version of `gulp` or `bower` to use within the project, by adding them to our devDependencies. This should help reduce the number of bugs people hit if `gulp` or `bower` or similar update with a breaking change and prevent people from running the software.

We should also include a serve script + a dependency that can serve this, then we can have a default start script that will serve the content without any extra downloads or effort required.

I've opened #26 to get the install down to one command instead of two, and will open further PRs when I have time.

- [x] Reduce complexity of installation. (Fixed with #26)
- [x] Add bower to devDependencies (Fixed with #28)
- [x] Alias gulp scripts to npm scripts, so we use the version of gulp defined locally (Fixed with #29)
- [x] Add a serve script (Fixed with #30)

Sorry if I've put too many commits into any 1 PR. I have a habit of getting my branches mixed up and accidentally make a branch from a branch instead of master. Feel free to let me know what you think, and discuss these proposals either inside each PR, or in this issue.

MaKleSoft:
Thanks for your contributions, @lewiscowper! I"ll look at the PRs as soon as possible.

MaKleSoft:
@lewiscowper I've given your PRs a lot of thought (see my comments on each of them) but right now I'm leaning towards not merging any of them. It breaks my heart to reject a PR every time but keeping a straight course for an open source project is hard and staying consistent usually means rejecting the majority of contributions. Please note that I appreciate every single contribution and suggestion regardless of whether I accept it or not! I'm keeping all of these open for now for you to be able to plead your case.

PR #28, opened 2015-10-30
Add bower to devDependencies

lewiscowper:
This PR specifies the version of bower that someone who wants to build the project will use, and combined with #26, it allows us to drop telling people that bower needs to be installed globally, and should make troubleshooting newer versions of bower easier, because we can test them in isolation before updating the `package.json`.

Stage 2 of #27

### DIFF
--- a/package.json
+++ b/package.json
@@ -2,6 +2,7 @@
   "name": "Padlock",
   "version": "0.0.0",
   "devDependencies": {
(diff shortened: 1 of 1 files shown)


MaKleSoft:
See my comment on https://github.com/MaKleSoft/padlock/pull/29 about managing bower and gulp locally.

PR #29, opened 2015-10-30
add npm script mirrors of gulp commands

lewiscowper:
Part 3 of #27

This allows us to use the specified version of gulp, and bower, and generally makes the development environment more predictable, with a new task called via `npm run dev` where you can run the watch-able tasks in parallel, so linting and testing all at once. :)

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,9 +14,18 @@
     "mkdirp": "^0.5.0",
     "ncp": "^0.6.0",
     "nib": "^1.0.3",
+    "npm-run-all": "^1.2.12",
     "q": "^1.0.1",
     "rimraf": "^2.2.8",
     "yargs": "^1.3.1"
   },
-  "dependencies": {}
+  "dependencies": {},
+  "scripts": {
+    "test": "gulp test",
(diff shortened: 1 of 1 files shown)


MaKleSoft:
I am generally a fan of managing dependencies locally wherever possible but in the case of command line tools like gulp or bower this makes little sense in my opinion. The whole point of installing gulp globally is to make it available anywhere via the `gulp` command. If gulp is only installed locally, running 'gulp deploy' for example will not work, even if run from the project directory directly. This means that every time you use gulp you have to either specify the path of the locally installed instance or pipe the command through `npm run`. The same applies to bower.

Yes, this would remove a couple of initial setup steps like globally installing bower and gulp and running `bower install` but I would argue that doing these things explicitly rather than implicitly is a good thing. Abstracting all these things away behind npm might make things easier in the beginning, but will likely result in contributors being less familiar with the dev environment and the tools involved.

I also understand your point about how managing bower and gulp locally gives the advantage of a more predictable dev environment, but to be realistic, the versions of neither of those tools have any significant impact on the stability of the code base.

PR #30, opened 2015-10-30
add serve script + dependency

lewiscowper:
This script adds a default, live-reloading server, and aliases the command to start it to `npm run server`. I'm no gulp expert, so I'm more than happy to defer this one, it's a little nicety I've found helpful when I've been running the project locally.

The final step in #27

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,6 +11,7 @@
     "gulp-vulcanize": "^6.0.1",
     "gulp-watch": "^0.6.9",
     "jshint-stylish": "^0.4.0",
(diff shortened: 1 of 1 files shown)


MaKleSoft:
Running a local web server is definitely an important part of the development workflow but I feel the choice of which one to use should be left to the individual developer. I know that adding this to the gulpfile doesn't necessarily hurt anyone but I would like to keep the project as lean as possible which means avoiding any external dependencies that are not strictly necessary.

PR #42, opened 2016-05-17
Adding KeePassX XML import option

bhavesh-k:
KeePassX password manager allows the user to export their database as an XML file. Functions have been added to allow a user to migrate from KeePassX by importing the XML database into Padlock.

### DIFF
--- a/src/import.js
+++ b/src/import.js
@@ -250,6 +250,69 @@ padlock.import = (function(crypto, DisposableSource, Papa) {
         return data.split("\n")[0] == "url,username,password,extra,name,grouping,fav";
     }
 
+    /**
+     * Checks if a given string represents a KeePassX exported XML file
+     */
+    function isKeePassXML(xmldata) {
+        return xmldata.split("\n")[0] == "<!DOCTYPE KEEPASSX_DATABASE>";
+    }
+
+    /**
+     * Imports data from the XML file exported by KeePassX
+     */
+     function importKeePassXML(xmldata) {
+
+         // parse the xmldata string into actual XML format
+         parser = new DOMParser();
+         xmlDoc = parser.parseFromString(xmldata,"text/xml");
+
+         // initialize list of all records in this XML file
+         var records = [];
+
+         // get all the "groups" (aka categories) from xmlDoc and loop through them
+         var groups = xmlDoc.getElementsByTagName("group");
+         for (group = 0; group < groups.length; group++) {
+             var category = groups[group].getElementsByTagName("title")[0];
+             var category = category.childNodes[0].nodeValue;
+
+             // get all the entries (records) within this category and loop through them
+             var entries = groups[group].getElementsByTagName("entry");
+             for (entry = 0; entry < entries.length; entry++) {
+                 var name = entries[entry].getElementsByTagName("title")[0];
+                 var username = entries[entry].getElementsByTagName("username")[0];
+                 var password = entries[entry].getElementsByTagName("password")[0];
+                 var url = entries[entry].getElementsByTagName("url")[0];
+                 var notes = entries[entry].getElementsByTagName("comment")[0];
+
+                 // row will have name,username,password,url,notes or empty spaces if any of these fields are empty
+                 var row = [name,username,password,url,notes];
+                 for (field=0; field < 5; field++) {
+                     if (row[field].hasChildNodes()) {
+                         row[field] = row[field].childNodes[0].nodeValue;
+                     } else {
+                         row[field] = "";
+                     }
+                 }
+
+                 // create a record from the above info and add to records
+                 var record = {
+             		name: row[0],
+             		category: category,
+             		fields: [
+                 		{name: "url", value: row[3]},
+                 		{name: "username", value: row[1]},
+                 		{name: "password", value: row[2]},
+                 		{name: "notes", value: row[4]}
+             		]
+         		};
+         		records.push(record);
+             }
+         }
+
+         collection.add(records);
+         return records;
+     }
+
     return {
         isSecuStoreBackup: isSecuStoreBackup,
         isPadlockBackup: isPadlockBackup,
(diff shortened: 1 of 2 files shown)


src/import.js
```diff
@@ -250,6 +250,69 @@ padlock.import = (function(crypto, DisposableSource, Papa) {
         return data.split("\n")[0] == "url,username,password,extra,name,grouping,fav";
     }
 
+    /**
+     * Checks if a given string represents a KeePassX exported XML file
+     */
+    function isKeePassXML(xmldata) {
+        return xmldata.split("\n")[0] == "<!DOCTYPE KEEPASSX_DATABASE>";
```
MaKleSoft: Why not just `xmldata.indexOf("<!DOCTYPE KEEPASSX_DATABASE>") !== -1`? Should be a little more reliable I think.

src/import.js
```diff
@@ -250,6 +250,69 @@ padlock.import = (function(crypto, DisposableSource, Papa) {
...
+    }
+
+    /**
+     * Imports data from the XML file exported by KeePassX
+     */
+     function importKeePassXML(xmldata) {
+
+         // parse the xmldata string into actual XML format
+         parser = new DOMParser();
+         xmlDoc = parser.parseFromString(xmldata,"text/xml");
```
MaKleSoft: What if the parsing fails? Should this maybe be in a `try {} catch {}` block?

src/import.js
```diff
@@ -250,6 +250,69 @@ padlock.import = (function(crypto, DisposableSource, Papa) {
...
+
+         // initialize list of all records in this XML file
+         var records = [];
+
+         // get all the "groups" (aka categories) from xmlDoc and loop through them
+         var groups = xmlDoc.getElementsByTagName("group");
+         for (group = 0; group < groups.length; group++) {
```
MaKleSoft: The `group` variable should be initialised as a local variable, i.e. `var group = 0`. Also, maybe use a variable name that indicates its an index, not the actual xml node. How about just using `i`?

src/import.js
```diff
@@ -250,6 +250,69 @@ padlock.import = (function(crypto, DisposableSource, Papa) {
...
+             var category = groups[group].getElementsByTagName("title")[0];
+             var category = category.childNodes[0].nodeValue;
```
MaKleSoft: Here you declare the `category` variable twice. You can skip the `var` the second time. Or better yet, use two different variables to make the code more readable. E.g.:
```js
var titleNodes = groups[group].getElementsByTagName("title")[0];
var category = titleNodes.childNodes[0].nodeValue;
```

src/import.js
```diff
@@ -250,6 +250,69 @@ padlock.import = (function(crypto, DisposableSource, Papa) {
...
+
+             // get all the entries (records) within this category and loop through them
+             var entries = groups[group].getElementsByTagName("entry");
+             for (entry = 0; entry < entries.length; entry++) {
```
MaKleSoft: Same as in the previous loop: Use `var` and a name that indicates an index

src/import.js
MaKleSoft: Actually, shouldn't it just be `titleNodes[0].nodeValue`?

src/import.js
```diff
@@ -250,6 +250,69 @@ padlock.import = (function(crypto, DisposableSource, Papa) {
...
+             		name: row[0],
+             		category: category,
+             		fields: [
+                 		{name: "url", value: row[3]},
+                 		{name: "username", value: row[1]},
+                 		{name: "password", value: row[2]},
+                 		{name: "notes", value: row[4]}
+             		]
+         		};
+         		records.push(record);
```
MaKleSoft: The indentation is off here

src/import.js
```diff
@@ -250,6 +250,69 @@ padlock.import = (function(crypto, DisposableSource, Papa) {
...
+                 var record = {
+             		name: row[0],
+             		category: category,
+             		fields: [
+                 		{name: "url", value: row[3]},
+                 		{name: "username", value: row[1]},
+                 		{name: "password", value: row[2]},
+                 		{name: "notes", value: row[4]}
+             		]
+         		};
```
MaKleSoft: That's a weird way of building the record object. A much more concise and readable variant would be

```js
var name = entries[entry].getElementsByTagName("title")[0].nodeValue;
var url = entries[entry].getElementsByTagName("url")[0].nodeValue;
var username = entries[entry].getElementsByTagName("username")[0].nodeValue;
var password = entries[entry].getElementsByTagName("password")[0].nodeValue;
var notes = entries[entry].getElementsByTagName("notes")[0].nodeValue;
var record = {
    name: name,
    category: category,
    fields: [
        {name: "url", value: url},
        {name: "username", value: username},
        {name: "password", value: password},
        {name: "notes", value: notes}
    ]
};
```

MaKleSoft:
Looks pretty good overall, although there are definitely quite a few changes to be made. I'll taker a closer look tomorrow.

MaKleSoft:
Btw, please always run the javascript linter on your code before you submit a pull request. See https://github.com/MaKleSoft/padlock#linting

src/import.js
```diff
@@ -250,6 +250,69 @@ padlock.import = (function(crypto, DisposableSource, Papa) {
...
+        return xmldata.split("\n")[0] == "<!DOCTYPE KEEPASSX_DATABASE>";
+    }
+
+    /**
+     * Imports data from the XML file exported by KeePassX
+     */
+     function importKeePassXML(xmldata) {
+
+         // parse the xmldata string into actual XML format
+         parser = new DOMParser();
```
MaKleSoft: You need a `var` here to define a local variable.

src/import.js
MaKleSoft: This line also misses a `var` btw.

MaKleSoft:
Where are you getting the keepassx xml data from? The osx client only gives me the option to export to csv...

src/import.js
MaKleSoft: Nevermind... what you're doing is right. However, a more concise way of writing it would be
```js
var titleNode = groups[group].querySelector("title"); // Will only return the first title node
var category = titleNode ? titleNode.innerHTML : "";
```
Actually, not much more concise but a little safer and easier to read imo.

MaKleSoft:
Seems this PR has gone stale so I'm closing it. If you want to take another shot at this at some point just reopen it or submit a new PR. Thanks!

PR #59, opened 2016-11-12
Fixed bias in random

Sc00bz:
I didn't find it, but I fixed it. https://twitter.com/dchest/status/796819713162641409

### DIFF
--- a/src/rand.js
+++ b/src/rand.js
@@ -21,15 +21,38 @@ padlock.rand = (function() {
         hexa: chars.numbers + "abcdef"
     };
 
+    //* Returns a random uniform value 0 to max, or -1 if max < 0 or max > 0x7fffffff
+    function secureRandom(max) {
+        var rand = new Uint32Array(1);
+        var result;
+
+        max = max | 0;
+        if (max < 0 || max > 0x7fffffff) {
+            return -1;
+        }
+
+        if (((max + 1) & max) === 0) {
+            window.crypto.getRandomValues(rand);
+            result = rand[0] & max;
+        } else {
+            var skip = 0x7fffffff - 0x7fffffff % (max + 1);
+
+            do {
+                window.crypto.getRandomValues(rand);
+                result = rand[0] & 0x7fffffff;
+            } while (result >= skip);
+            result %= max + 1;
+        }
+        return result;
+    }
+
     //* Creates a random string with a given _length_ comprised of given set or characters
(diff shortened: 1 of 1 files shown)


MaKleSoft:
Wow. Seems like for some reason I thought `Math.floor(rnd[i] * charSet.length / 256)` would be more evenly distributed than `rnd[i] % charSet.length` which obviously isn't true. I blame it on too much coffee (or too little).

Some things to point out here:

- `randomString()` is used exclusively for the 'password generator' tool within the app which produces random passwords for use with other services, sites etc. While the bias is reducing the entropy of the produced passwords, it doesn't strictly make them 'insecure'. This is no excuse for sloppiness of course.

- Since this is only used by the password generator tool, we have some implicit constraints here. Most notably, `charSet.length <= charSets.full.length == 94`, so we don't need to cover edge cases like `max < 0 || max > 0x7fffffff`

While I appreciate your fix, I think it is slightly 'overkill'. Rejection-based sampling is probably the more straighforward solution here:

```js
//* Creates a random string with a given _length_ comprised of given set or characters
function randomString(length, charSet) {
    length = length || 32;
    charSet = charSet || charSets.full;
    var rnd = new Uint8Array(1);
    var str = "";
    while (str.length < length) {
        window.crypto.getRandomValues(rnd);
        // Prevent modulo bias by rejecting values larger than
        // the highest muliple of `charSet.length`
        if (rnd[0] > 255 - 256 % charSet.length) {
            continue;
        }
        str += charSet[rnd[0] % charSet.length];
    }
    return str;
}
```

What do you think?

dchest:
@MaKleSoft your code looks good. I'd use an estimated length for rnd array to get most of the characters in one call from getRandomValues, but it complicates things, and performance is not critical here, so it's fine.

If you'd like to add tests, here's a test I made for biases for *node-randomstring* https://github.com/klughammer/node-randomstring/commit/15ce2fd0841f46141cb2a3c7bf333e3ab064c416#diff-910eb6f57886ca16c136101fb1699231R79

MaKleSoft:
Thanks for your feedback, @dchest!

Fixed it using rejection sampling and added tests: https://github.com/MaKleSoft/padlock/commit/cc910eca5db614e4d6c0b68c39815c34e17efada

Thanks again @Sc00bz for your contribution, even if we didn't end up merging it!

PR #84, opened 2017-05-14
Use `npm run <cmd>` for various dev actions

azeemba:
Looks like `gulp eslint` was already broken. While fixing that, I also
removed the need to have gulp installed globally. Instead we can use
npm scripts section to run the same commands.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,13 +6,11 @@ A minimal open source password manager built with [Polymer](http://www.polymer-p
 Install these if you haven't yet:
 
 - [Node.js and npm](http://nodejs.org/)
-- [Gulp](http://gulpjs.com/)
-- [Bower](http://bower.io/)
 
 Now from inside the project folder, install the local requirements:
 
     npm install
-    bower install
+    npm run bower
 
 For the [HTML Imports](http://www.polymer-project.org/platform/html-imports.html) polyfill of Polymer to work, the app has to be served on a local web server. You can use whatever web server you prefer to serve the files. For example:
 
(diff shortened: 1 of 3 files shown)


MaKleSoft:
This is generally not a bad idea but I'm hesitant to merge this since the build process has changed significantly in the `v1.2` and `v2.0` branches. I'll definitely look at applying the same idea to the new build system. Or let you do it if you want. Well, 2.0 is coming along nicely and will probably be merged into master soon. Let's revisit this then.

MaKleSoft:
@azeemba So after merging the v2.0 branch I was actually pretty close to implementing your suggestions but after thinking about it some more I realised that installing these tools als local dependencies should be considered an anti-pattern. Tools like bower and gulp are designed to be run as global commands and abstracting away their use is unnecessary and frankly not very helpful. Specifically:

1. Adding build tools als local dependencies slows down `npm install` and unnecessarily inflates the size of the project. There is a reason why these tools are usually installed globally.
2. Running the commands through `npm run` adds an unnecessary layer of indirection and makes it harder to understand what's really going on. I think it's better to have a few additional commands if that makes the tooling more transparent.

Any thoughts?

azeemba:
I totally see your point and I have seen many projects on both sides.
Here are the arguments for the other side:

1. If you have a dependency, then it is better to have an explicit dependency than an implicit one. That is the main reason the `devDependency` section exists in `package.json`.
2. `npm run` provides a common consistent interface by leveraging the indirection. This way I can pick up a new project and reasonably guess that `npm run test`, `npm run lint` and `npm run build` will do reasonable things without really needing to know what tools are being used for testing, linting or building.

As I said, I don't think there is a strict right and wrong here. You are probably just going to have to pick one.

MaKleSoft:
I'll probably end up using a compromise. Gulp for example, needs to be installed locally anyway so we might as well use the local copy for the commands. I agree that having `npm run` as a consistent interface would be nice...

PR #133, opened 2017-10-16
Mask Field Values

MaKleSoft:
* Mask field values by default
* Add button for masking/unmasking value
* Add field 'info' dialog for displaying advanced options, like 'delete'

Fixes https://github.com/MaKleSoft/padlock/issues/126

### DIFF
--- a/app/src/ui/dialog/dialog-mixin.html
+++ b/app/src/ui/dialog/dialog-mixin.html
@@ -61,6 +61,10 @@ padlock.DialogMixin = (superClass) => {
             return lineUpDialog(() => dialog.generate());
         }
 
+        getSingleton(elName) {
+            return getDialog(elName);
+        }
+
     };
 };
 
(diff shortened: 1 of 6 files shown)


MaKleSoft:
@pwasiewicz @TizianoPerrucci Here is my implementation. I did end up introducing a dedicated button after all since it's simply the easiest to understand for users and works well both on desktop and mobile. The new button replaces the copy button which moved one slot down. The buttons for deleting the field and generating random values have moved into a new 'info' dialog, which will eventually hold other information and functionality as well (the 'generate' option is still missing from the dialog. I'll add it later). Would love to get your feedback on this. Does this suit your needs?

screen shot 2017-10-16 at 19 22 13

MaKleSoft:
@pwasiewicz thanks for your feedback! I'll merge this now.

TizianoPerrucci:
@MaKleSoft yes it does, thanks for your work!
I'll give it a shot later with the latest release later

TizianoPerrucci:
@MaKleSoft I actually have only one remark, where I agree with @pwasiewicz, that the number of "dots" should not depend on value length

PR #190, opened 2018-06-19
Dont give last sync warning if autosync is already enabled

suihkulokki:
I use padlock from multiple machines. On some machines I dont
use padlock daily, so I sometimes get the alert that I haven't
synchronized for X days. But I already have autosync enabled,
and padlock is already synchronizing behind the "please enable
autosync" warning dialog.


MaKleSoft:
@suihkulokki Looks good! Thanks for your contribution!

PR #237, opened 2019-11-07
Add Storage Support for Amazon DynamoDB

mandusm:
Adding support for using Amazon DynamoDB as a storage provider to provide persistence across multiple instances of `padloc-server` for Elasticity and Availability. 

#### Environment Variables
| VARIABLE NAME  | TYPE |  ACCEPTED VALUES |DEFAULT|
| ----------- | ----------- |----------- |----------- |
| STORAGE_PROVIDER      |  Storage Provider       | LevelDB, DynamoDB  |LevelDB |
| AWS_REGION   | String        | aws region name|us-west-2 |
| AWS_DDB_TABLE   | String        | pre-created DynamoDB table name |padloc |

#### Requirements. 
A pre-created DynamoDB table with `Primary partition key` named `storeable` of type `String`

### DIFF
--- a/packages/server/src/storage.ts
+++ b/packages/server/src/storage.ts
@@ -2,6 +2,9 @@
 import level from "level";
 import { Storage, Storable, StorableConstructor } from "@padloc/core/src/storage";
 import { Err, ErrorCode } from "@padloc/core/src/error";
+import { AwsRegion } from "aws-sdk/clients/servicequotas";
+import DynamoDB from "aws-sdk/clients/dynamodb";
+import { CredentialProviderChain } from "aws-sdk";
 
 export class LevelDBStorage implements Storage {
     private _db: any;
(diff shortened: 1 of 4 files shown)


MaKleSoft:
Thanks for the PR! Looks good overall, apart from some formatting inconsistencies (please use [Prettier](https://prettier.io/) to format code before committing).

However, we'll have to figure out a better way to load alternative providers for storage, email, crypto etc, since we want to keep the dependencies of the official packages to a minimum and adding them all to the default init script is simply not feasible. I'll have to think about this, but I can already tell you that we probably won't merge this into this repo (or at least not into the `@padloc/server` package). We'll definitely find a place for this though so we'll keep the PR around for now 😊

packages/server/src/storage.ts
```diff
@@ -36,3 +39,73 @@ export class LevelDBStorage implements Storage {
...
+            // Checking whether the _item_ property is defined to
+            // determine whether the Item is not found.
+
+            if( typeof(raw.Item) === 'undefined') {
+                throw new Err(ErrorCode.NOT_FOUND);
+            }else{
+                return res.fromJSON(raw.Item.secret.S);
+            };     
+        } catch (e) {
+            throw e;
```
MaKleSoft: If you're just going to re-throw the error, you may as well just omit the catch block :)

mandusm:
Awesome. Thanks for the feedback. 
Yeah, you’re right, the storage selecter might be better suited to be a webpack function. I’ll think about a way to do it too. 

I have some code lying around for AWS SES support too, so I’ll see if I can fiddle with webpack to make it work.

MaKleSoft:
How do you imagine webpack to solve this? We're not using webpack for the backend right now  and I'd prefer to keep it that way unless it provides a clear value. Right now, I'm thinking that if people want to use different providers for email, storage etc. that they should simply write a custom init script. That way alternative providers could be maintained outside of the main repo and simply imported from a node module.

mandusm:
I don't know what I meant by webpack... I was obviously delirious/not paying attention when I typed that.
A custom storage module sounds like the right way to go, that way you just do like a `npm install --save padloc-storgage-dynamodb` 

I'll roll a standalone package sometime this week with only the DynamoDB code in it.

MaKleSoft:
> A custom storage module sounds like the right way to go, that way you just do like a npm install --save padloc-storgage-dynamodb

Yes, exactly. You'll need the `@padloc/core` package as a dependency though so I'll have to publish it to npm first. I'll let you know once I've done that.

MaKleSoft:
@mandusm `@padloc/core` is now public on npm so you can add it as a dependency. Let me know once you've published your package and we'll find a place to mention it in the documentation somewhere 😊Closing this now.

PR #258, opened 2020-01-03
Russian translation

avvolodin:
Just UI translation.

### DIFF
--- a/packages/locale/res/translations/ru.json
+++ b/packages/locale/res/translations/ru.json
@@ -1,0 +1,1574 @@
+[
+  [
+    "OK",
+    "OK"
+  ],
+  [
+    "o f f l i n e",
+    "о ф л а й н"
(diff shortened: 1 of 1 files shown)


MaKleSoft:
@avvolodin Thats awesome, thank you so much! Have you tested this? Not sure if the current font even supports cyrillic 😅

avvolodin:
> @avvolodin Thats awesome, thank you so much! Have you tested this? Not sure if the current font even supports cyrillic 😅

It supports cyrillic, but i didn't test it deeply. I have some doubts about context of several phrases but didn't have time yet 
2020-01-03 21_52_03-Padloc
to set up and examine that pages.

MaKleSoft:
Ok, that's good enough for now! You can always come back and add improvements once you've had a chance to test it more.

Issue #248, opened 2019-12-05
HELP WANTED: Translate Padloc to other languages 


MaKleSoft:
Title says it all. The Padloc app is currently available in English and German, but we'd love to support more languages! If you'd like to help translate Padloc into your language (or any language you are sufficiently fluent in), here is how you can do it:

1. Fork this repo
2. Create a copy of the [empty translations template](../blob/master/packages/locale/res/translations/_template.json)
3. Change the name to the appropriate [country code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-2#Officially_assigned_code_elements)
4. Fill in the translations
5. Create a pull request

Feel free to use this thread to let others know you're working on a translation or to coordinate your efforts.

That's it!  For more information, check out the [locale package](../blob/master/packages/locale/). Looking forward to your contributions 😊

moigarmo:
Hi guys! I'm happy to help here. 
I'm translating into spanish.

Cheers.

MaKleSoft:
@moigarmo Great! Let me know if you have any questions.

moigarmo:
I've created a pull request with the translations in spanish.
I hope it helps.

MaKleSoft:
@moigarmo Wow, that was fast. Thanks!

moigarmo:
Hi @MaKleSoft, I've been playing around with the app a bit and I found a couple of errors in the translations I'd like to fix. But I don't know how to handle one kind of them. 
In spanish there are some sentences that in english are written the same but depending on the context they may have a different meaning and because of the way translations work in this project it is not easy to differentiate between them.
For instance, "log in" in the main screen means that you have to log in to perform further actions, whereas when you are creating a new item, the option log in means that you want to create a log in information item, which would be translated a bit diffetent.
I don't think it's a big deal but it seems awkard the way it is displayed in that screen. And the way translations work in this project don't allow, that I know of, a way to translate the same sentence differently.

If you know a way to do that it'd be great.

Thanks,
Moi.

MaKleSoft:
@moigarmo Yes, I've noticed the same issue with German, specifically about the different meanings of "Login". Would you mind creating a separate issue listing all problems and where in the app they occur (screenshots would be amazing)? I'll figure out a strategy based on that.

avvolodin:
I've added PR for russian ui translation.
[https://github.com/padloc/padloc/pull/258](https://github.com/padloc/padloc/pull/258)

jhjacobs81:
[...]

PR #294, opened 2020-10-31
French translation

detobel36:
Added french translation.   

Notice:   
In french there is not a "real" difference between "download" and "upload" (which could be translate by "téléchargement"). So, it's "usual" in french to say "upload".

Linked to issue #248

### DIFF
--- a/packages/locale/res/translations/fr.json
+++ b/packages/locale/res/translations/fr.json
@@ -1,0 +1,1574 @@
+[
+  [
+    "OK",
+    "OK"
+  ],
+  [
+    "o f f l i n e",
+    "hors ligne"
+  ],
+  [
+    "Could not find invite! Did you use the correct link?",
+    "Impossible de trouver une invitation ! Avez-vous utilisé le bon lien ?"
+  ],
+  [
+    "Get Premium",
+    "Devenir Premium"
+  ],
+  [
+    "Cancel",
+    "Annuler"
+  ],
+  [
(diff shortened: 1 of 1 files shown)


MaKleSoft:
Good stuff! Thank you so much! 🥳👍

LaunchedBerry07:
Hi! I'm a newbie on GitHub, so please redirect me, if necessary. I would like to help for the translation into French. My main career is as a translator.

(Btw, the term "téléverser" for "upload" is more accurate than "télécharger" :).)

detobel36:
> (Btw, the term "téléverser" for "upload" is more accurate than "télécharger" :).)
Yes but nobody use it xD

LaunchedBerry07:
> > (Btw, the term "téléverser" for "upload" is more accurate than "télécharger" :).)
> 
> 
> Yes but nobody use it xD       

Indeed, ahah!

MaKleSoft:
@LaunchedBerry07 thanks for offering to help! Unfortunately @detobel36 beat you to it this time but there are new things being added to the app all the time that will require translations as well. I'll be sure to create a new issue when a new batch comes in. Apart from that, reviewing existing translations can never hurt either, especially in the context of the actual UI (I know the texts from the translation file by themselves and without context can be quite ambiguous and hard to translate). Finally another thing that would be helpful would be to curate the [word lists](https://github.com/padloc/padloc/blob/master/packages/locale/res/wordlists/fr.json) used to generate passphrases to make sure there are no strange or (too) offensive words in there (funny/weird is ok). But to be honest, that's such a shitty job that I don't really want to ask anyone to do it 😅

LaunchedBerry07:
@MaKleSoft Ahah! Well, honestly, those tasks don't seem too boring to me! I could help with those, if you want to.

PR #306, opened 2021-04-28
Improve User Credential Quality

justinkonecny:
- Add check for how many times a user's password is reused in their main vault when saving or updating a vault item
 - Add check for how strong a user's password is when creating a new vault item
 - Add check against HaveIBenPwned API's for how many times a user's password has been detected in a data breach after entering a password in the item dialog
 - Add check against HaveIBeenPwned's API for whether or not a user's email has been detected in a data breach for a given domain when entering credentials in the item dialog
 - Add HaveIBeenPwned key to README for server setup


packages/app/src/elements/field.ts
```diff
@@ -8,6 +8,9 @@ import { Textarea } from "./textarea";
 import "./input";
 import "./textarea";
 import "./totp";
+import { app } from "../globals";
+
+const crypto = require('crypto');
```
MaKleSoft, line 13: Please always use the es6-style module imports. Also the `@padloc/app` package is designed to be run in a browser environment, so node modules should not be expected to be available in this context. Finally, for any cryptographic functionality the `@padloc/core/src/crypto` module should be used.

packages/app/src/elements/field.ts
```diff
@@ -23,6 +26,12 @@ export class FieldElement extends BaseElement {
     @property()
     type: FieldType = FieldType.Note;
 
+    @property()  // number of times user password has been detected in a data breach
+    _passwordBreachCount: number = -1;
```
MaKleSoft, line 30: Should be private.

packages/app/src/elements/field.ts
```diff
@@ -23,6 +26,12 @@ export class FieldElement extends BaseElement {
...
+
+    @property()  // list of missing properties that would make the user password strongest
+    _missingList: string[] = [];
```
MaKleSoft, line 33: Should be private

packages/app/src/elements/field.ts
```diff
@@ -200,6 +228,123 @@ export class FieldElement extends BaseElement {
...
+     */
+    private async _checkPasswordBreach() {
+        if (!this.value) {
+            this._passwordBreachCount = -1;
+            return;
+        }
+
+        const sha1Hash = crypto.createHash("sha1")
+            .update(this.value)
+            .digest("hex");
```
MaKleSoft, line 243: Again, please use the `@padloc/core/src/crypto` module for all cryptographic needs.

packages/app/src/elements/field.ts
```diff
@@ -200,6 +228,123 @@ export class FieldElement extends BaseElement {
...
+        if (strength >= 12) {
+            return "value-input value-strong";
+        } else if (strength > 8) {
+            return "value-input value-good";
+        } else if (strength > 6) {
+            return "value-input value-okay";
+        }
+
+        return "value-input value-weak";
+    }
```
MaKleSoft, line 318: This is a very simplistic way of "calculating" password strength and while it's better than nothing, it's certainly not an accurate or reliable way to measure password entropy. Besides, we already have a [much better way](https://github.com/padloc/padloc/blob/master/packages/app/src/lib/util.ts#L25) to do this, which based on the [zxcvbn](https://github.com/dropbox/zxcvbn) library.

packages/core/src/server.ts
```diff
@@ -1174,6 +1179,78 @@ export class Controller extends API {
...
+        this.log("passwordBreachStatus.get", {
+            passwordHash: sha1Hash
+        });
+
+        return new Promise<PasswordBreachResult>((resolve, reject) => {
+            const sha1HashPrefix = sha1Hash.substring(0,5);
+            const sha1HashSuffix = sha1Hash.substring(5);
+            const url = `https://api.pwnedpasswords.com/range/${sha1HashPrefix}`;
+
+            axios.get(url)
```
MaKleSoft, line 1199: Please don't import an external packages just for making network requests. We're trying to keep external dependencies to an absolute minimum and the core package in particular should stay completely dependency-free.

MaKleSoft (commented):
Thank you for your pull request! Auditing credentials within vaults is definitely a worthwhile endeavour and you have made a decent effort at implementing this. However I am afraid we can't merge this in its current state. In addition to the things I pointed out in my inline comments there are a couple of other issues that prevent us from merging this:

- External pull requests should be as small as possible and should only address one specific issue at a time. You've mixed several concerns here (documentation, password strength check, check for leaked credentials). Separating these into smaller, individual pull requests would make them easier to review and more likely to be merged.
- We try to maintain a **very** high standard for things like consistent formatting, a concise and clean code style and clearly structured dependency graph. While most of your deviations from those standards are small, they still prohibit us from merging your PR. Honestly, the fault here is mostly on our side - we need to better document our coding standards and tooling to make contributions like this easier.
- For contributions other than bug fixes or smaller, non user-facing features, please discuss your plans with us beforehand to make sure your changes align with our product vision and don't collide with our own development plans (which is actually the case here).

Again, thanks a lot for taking a shot a this - we really appreciate the effort! However for reasons mentioned above I'm going to reject this PR. If you would like to give this another shot, I would ask you to wait for us to give you the go-ahead first. We're currently in the process of implementing a major update and any changes to the UI part of the project will likely cause conflicts with our own efforts.

Issue #320, opened 2021-12-15
Lock dependencies' versions in package.json files


BrunoBernardino:
This is something that I generally recommend (always add dependencies with `--save-exact` and use a separate and intentional process to update them, rather than having them be updated unintentionally), because otherwise there can be unexpected side effects (and reproducible builds are harder). It's less important for `devDependencies`, but I still recommend for that.

Some interesting links on the subject:

- https://stackoverflow.com/questions/45022048/why-does-npm-install-rewrite-package-lock-json
- https://docs.npmjs.com/cli/v6/configuring-npm/package-locks
- https://the-guild.dev/blog/how-should-you-pin-dependencies-and-why

Additionally, adding `.nvmrc` and `.npmrc` files to lock the recommended `node` version, the `--save-exact`, and also add `npm` and `node` to `engines` in `package.json` helps with reproducible builds, and development in general.

PR #324, opened 2021-12-15
Add CI - Run Tests (v4)

### LINKED ISSUE #321: Implement basic CI for running tests
A basic CI should be implemented that at a minimum runs tests, to prevent issues like #318. Later it can be extended for building multiple artifacts.

GitHub Actions is a great option, and is already enabled for this repo.

BrunoBernardino:
This makes it so that on every push (includes PRs), dependencies and tests run on GitHub Actions.

It also adds a small badge reporting the latest push/run on the `v4` branch.

Fixes #321


packages/core/test/account.ts
```diff
@@ -35,6 +35,6 @@ suite("Account", () => {
         await account.unlock(password);
 
         assert.equal(bytesToHex(publicKey), bytesToHex(account.publicKey));
-        assert.equal(bytesToHex(privateKey), bytesToHex(account.privateKey));
+        assert.equal(bytesToHex(privateKey as Uint8Array), bytesToHex(account.privateKey as Uint8Array));
```
BrunoBernardino, line 38: This fixes one of the errors I was getting.

MaKleSoft (approved):

BrunoBernardino:
The CI failures probably need #320 in order to not continue.

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,16 @@
...
+on: [push]
+
+jobs:
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
```
MaKleSoft, line 12: Let's use node v16 since that is what production will run on

MaKleSoft (changes requested):

.github/workflows/run-tests.yml
BrunoBernardino: If you tell me which specific version of `node` and `npm` you're on, I can work on #320 next with those set.

MaKleSoft: Like I said lets use 16.x + the latest version of npm that goes with it 👍

PR #325, opened 2021-12-15
Lock dependencies' versions

### LINKED ISSUE #320: Lock dependencies' versions in package.json files

BrunoBernardino:
This implements `.nvmrc` and `.npmrc` files, as well as locking `package.json` versions, in order to get us closer to reproducible builds and a more stable environment.

Closes #320


BrunoBernardino:
Alright, you can see in https://github.com/padloc/padloc/runs/4536694040?check_suite_focus=true that the `npm ci` (Install dependencies) step passes now, but the tests still fail, with the same result as local, which is to be tackled separately.

MaKleSoft (approved):

PR #327, opened 2021-12-16
Allow running dev server without explicit configuration

MaKleSoft:


BrunoBernardino (approved):
I've pulled this down and can confirm it works as expected! The changes are also clear and readable. Thanks!

Issue #322, opened 2021-12-15
Add prettier + linting to tests


BrunoBernardino:
To make sure others feel more confident contributing, it would be a good idea to have something like `prettier` and `eslint` checks running on `npm test`, and let people know if they need to change something. Additionally, having a `npm run prettier` to fix things would be helpful.

BrunoBernardino:
Since we don't want to implement `eslint` as `tsc` is enough (and it runs via the build commands in CI), and `prettier` was added to the CI as part of #328  this can be marked as complete.

PR #328, opened 2021-12-16
Feature: 1pux file import

BrunoBernardino:
This allows importing [1pux files from 1Password](https://support.1password.com/1pux-format/).

Everything is working except for tests (still fail on `v4`).

This PR/branch includes other changes for version locking that are more future-proof, while keeping the build passing (except for the aforementioned tests).


lerna.json
```diff
@@ -2,5 +2,6 @@
   "packages": [
     "packages/*"
   ],
-  "version": "4.0.0"
+  "version": "4.0.0",
+  "exact": true
```
BrunoBernardino, line 6: This makes it so that when we run `lerna add`, it defaults to the exact version of the package.

package.json
```diff
@@ -33,6 +33,7 @@
     "tauri:dev": "lerna run --parallel --scope '@padloc/{server,tauri}' --parallel dev",
     "repl": "cd packages/server && npm run repl && cd ../..",
     "test": "lerna run test",
-    "locale:extract": "lerna run extract --scope '@padloc/locale'"
+    "locale:extract": "lerna run extract --scope '@padloc/locale'",
+    "add": "lerna add"
```
BrunoBernardino, line 37: We don't need to have this, but made it easier for me to add a new dependency (`npm run add -- 1pux-to-csv --scope=@padloc/app` vs `./node_modules/.bin/lerna add 1pux-to-csv --scope=@padloc/app`).

packages/app/package.json
```diff
@@ -40,7 +41,7 @@
     "event-target-shim": "6.0.2",
     "http-server": "0.12.3",
     "jsqr": "1.4.0",
-    "lit": "*",
+    "lit": "2.0.0-rc.2",
```
BrunoBernardino, line 44: This forces the version (better), without breaking (yay). Previously I had attempted something more recent, which broke.

packages/app/src/elements/settings-tools.ts
```diff
@@ -29,11 +29,15 @@ export class SettingsTools extends StateMixin(LitElement) {
...
+            await this._importDialog.show(reader.result as string | Uint8Array);
             this._fileInput.value = "";
         };
 
-        reader.readAsText(file);
+        if (file.name.endsWith('.1pux')) {
+            reader.readAsArrayBuffer(file);
+        } else {
+            reader.readAsText(file);
+        }
```
BrunoBernardino, line 40: Basically because `.1pux` is a zip file, it can't be read as text/string, which requires the type additions above and below.

packages/app/src/lib/import.ts
```diff
@@ -244,9 +251,113 @@ export async function asLastPass(data: string): Promise<VaultItem[]> {
  * Checks if a given string represents a LastPass CSV file
  */
 export function isLastPass(data: string): boolean {
-    return data.split("\n")[0] === "url,username,password,extra,name,grouping,fav";
+    try {
+        return data.split("\n")[0] === "url,username,password,extra,name,grouping,fav";
+    } catch (error) {
+        return false;
+    }
```
BrunoBernardino, line 258: This would fail for the 1pux/zip file.

packages/core/src/error.ts
```diff
@@ -50,6 +50,7 @@ export enum ErrorCode {
 
     NOT_FOUND = "not_found",
     INVALID_CSV = "invalid_csv",
+    INVALID_1PUX = "invalid_1pux",
```
BrunoBernardino, line 53: I didn't see any translation/text for this. Not sure if I missed something.

packages/core/src/messenger.ts
```diff
@@ -104,6 +104,6 @@ export class StubMessenger implements Messenger {
 
 export class ConsoleMessenger implements Messenger {
     async send(recipient: string, message: Message<any>) {
-        console.log(`Message send to ${recipient}: ${message.data}`);
+        console.log(`Message sent to ${recipient}: ${JSON.stringify(message.data)}`);
```
BrunoBernardino, line 107: I was getting `([object Object])` in the logs, which made it difficult to see the data.

.github/workflows/run-tests.yml
```diff
@@ -16,5 +16,5 @@ jobs:
         run: npm run pwa:build
       - name: Test starting zero-config server
         run: npm run server:start-dry
-      - name: Run tests
-        run: npm test
+      #- name: Run tests
+      #  run: npm test
```
BrunoBernardino, line 20: I've disabled these for now since they're out of date and we can at least trust green === it builds. A separate PR will bring them back.

packages/app/src/lib/import.ts
```diff
@@ -244,9 +251,113 @@ export async function asLastPass(data: string): Promise<VaultItem[]> {
...
+                const secret = new URL(extraField.value).searchParams.get('secret');
+                if (secret) {
+                    fields.push(new Field({ name: extraField.name, value: secret, type: FieldType.Totp }));
+                }
+            } catch (error) {
+                // Do nothing
+            }
+        } else {
+            // @ts-ignore All of extraField.type possibilities match FieldType.*
+            fields.push(new Field({ name: extraField.name, value: extraField.value, type: extraField.type }));
```
MaKleSoft, line 302: Let's try to avoid `@ts-ignore` as much as possible. In this case a simple type cast would be better: `extraField.type as FieldType`.

packages/app/src/lib/import.ts
```diff
@@ -244,9 +251,113 @@ export async function asLastPass(data: string): Promise<VaultItem[]> {
...
+}
+
+export async function as1Pux(file: string | Uint8Array): Promise<VaultItem[]> {
+    try {
+        const dataExport = await parse1PuxFile(file);
+
+        const items = [];
+        
+        for (const account of dataExport.data.accounts) {
+            for (const vault of account.vaults) {
```
MaKleSoft, line 317: Idea: If there are multiple vaults, add the vault name as a tag? That way the items can easily be split up into multiple vaults later.

MaKleSoft: I.e.: (further down)
```ts
if  (account.vaults.length) {
    parsedItem.tags.push(vault.attrs.name);
}
```

packages/app/src/lib/import.ts
```diff
@@ -244,9 +251,113 @@ export async function asLastPass(data: string): Promise<VaultItem[]> {
...
+        throw new Err(ErrorCode.INVALID_1PUX);
+    }
+}
+
+/**
+ * Checks if a given string/Uint8Array represents a 1Password 1pux file
+ */
+export async function is1Pux(file: string | Uint8Array): Promise<boolean> {
+    try {
+        const dataExport = await parse1PuxFile(file);
```
MaKleSoft, line 340: I'm thinking it would probably be easer if we simply check the file extension. For that we'd need a reference to the file object of course, but I'm thinking maybe it would be better to keep a reference of the File object instead of `string | Uint8Array` anyway? The file would then only be read / parsed when necessary. The big advantage of this would be that we could load the relevant code on demand, instead of having to load it for every file type.

packages/app/src/lib/import.ts
```diff
@@ -244,9 +251,113 @@ export async function asLastPass(data: string): Promise<VaultItem[]> {
...
+        return Boolean(dataExport.attributes && dataExport.data);
```
MaKleSoft, line 341: I usually use `!!` when I want to convert a truthy/falsy value into a boolean. But I think I like the explicit use of `Boolean` better...

packages/core/src/error.ts
MaKleSoft: The error codes are kind of arbitrary and not really used consistently. I'm actually thinking about using more generic error codes in general. `ErrorCode.ENCODING_ERROR` would do here as well for example.

packages/app/src/elements/settings-tools.ts
MaKleSoft: I think the better option would be to defer reading the file at this point and pass around the file object instead. `guessFormat` would then accept a file object for example, and so would `as1Pux` and all the other import functions. This would also allow us to use the file extension and mime type for guessing the format. What do you think?

MaKleSoft (changes requested):
Looks great for a first pass! I've been thinking about the external package vs internal module question for the 1pux parser though and I think I'd prefer having it in this code base directly. Like I mentioned elsewhere I'd like to keep external dependencies to a minimum and having it in the same code base would make maintenance a lot easier. Of course you can still maintain the 1pux-to-csv package separately if you like!

package.json
BrunoBernardino: Almost seems like lerna should be replaced with [`npm workspaces`](https://docs.npmjs.com/cli/v7/using-npm/workspaces). I could add a command to do this, but would need to be in something like a `Makefile` or `.sh` script. If that doesn't sound like a terrible idea, I can implement it in a separate PR.

packages/app/src/elements/settings-tools.ts
BrunoBernardino: I'm happy to make that change, I just wasn't sure if you'd be fine with changing the API too much.

package.json
MaKleSoft: Let's put a pin in it for now - the pain definitely isn't big enough to warrant spending too much time on this right now.

packages/core/src/error.ts
BrunoBernardino: Ok. I'll keep it. Let me know if you'd prefer I replace it with `ENCODING_ERROR`.

packages/app/src/lib/import.ts
BrunoBernardino: In this line, the vault name is being passed along to be added as a tag already: https://github.com/padloc/padloc/pull/328/files/feb7a3ce532bf85ce2a4dfb5c8e25fbad5c0f4b2#diff-f83e7b908e8e3f2938f2ff2d2b0e7d15930ddcc1065909bdb3a0de0251aa7e01R320 

Let me know if something isn't clear, or if you meant something else.

packages/app/src/lib/import.ts
BrunoBernardino: That makes sense. I'll make that change.

packages/app/src/lib/import.ts
BrunoBernardino: Yep, I'm a fan of readability (`error` vs `e` or `err`, or example, and `Boolean()` instead of `!!`). So much clearer and harder to miss!

BrunoBernardino:
> Looks great for a first pass! I've been thinking about the external package vs internal module question for the 1pux parser though and I think I'd prefer having it in this code base directly.

Sounds good and makes sense. I'll create a separate file for the parser stuff, let me know if you'd prefer I just add it all in the `import.ts` file, like for the other formats.

README.md
```diff
@@ -61,6 +61,8 @@ For more configuration options, see [Configuration](#configuration)
 | `npm run dev`          | Starts backend server and client app in dev mode, which watches for changes in the source files and automatically rebuilds/restarts the corresponding components. |
 | `npm test`             | Run tests.                                                                                                                                                        |
 
+To add dependencies, you can use `scope=[scope-without-@padloc/] npm run add [package]` and to remove them, run `scope=[scope-without-@padloc/] npm run remove [package]`.
```
BrunoBernardino, line 64: So now we have these two commands. They're not as "sturdy" as if they were `Makefile` or `.sh`, but they work and make these actions simpler and easier to make.

For reference, here's what I had to do to remove `1pux-to-csv` and add `jszip` to `@padloc/app`:

```bash
scope=app npm run remove 1pux-to-csv
scope=app npm run add jszip
```

packages/app/src/elements/import-dialog.ts
```diff
@@ -67,13 +67,24 @@ export class ImportDialog extends Dialog<string, void> {
...
-    async show(input: string) {
+    async show(file: File) {
         await this.updateComplete;
         const result = super.show();
-        this._rawData = input;
-        this._formatSelect.value = (imp.guessFormat(input) || imp.CSV).value;
-        this._parseString();
-        this._vaultSelect.value = app.mainVault!;
+
+        const reader = new FileReader();
```
BrunoBernardino, line 74: I experimented with, but ended up not liking passing the `FileReader` down the line. It meant we had to keep a reference of the file inside the `ImportDialog`, and there are no performance gains here, since we're not parsing 1pux/zip yet at this point, now.

packages/app/src/elements/import-dialog.ts
```diff
@@ -88,7 +99,7 @@ Github,"work,coding",https://github.com,john.doe@gmail.com,129lskdf93`)
         );
     }
 
-    private async _parseString(): Promise<void> {
+    private async _parseData(): Promise<void> {
```
BrunoBernardino, line 102: Just making it more "broad", now that it isn't just strings.

BrunoBernardino:
@MaKleSoft implemented the changes you requested, and tested again. Let me know what you think, now!

MaKleSoft (changes requested):
Sorry if it seems like I'm doing a lot of bike shedding here but I think it's worthwhile trying to learn as much as possible about our respective coding styles and preferences early on so we can spend less times discussing implementation details later. Code consistency is quite important to me (bordering a bit on OCD 😅) so please forgive obsession with details. I'll try to be a bit more hand-off in the future 😊

packages/app/src/lib/1pux-parser.ts
```diff
@@ -0,0 +1,308 @@
...
+    }
+  });
+
+  // Extract some more extraFields
+  item.details.sections.forEach((section) => {
+    section.fields.forEach((field) => {
+      let value = '';
+      let type: ExtraFieldType = 'text';
+
+      if (Object.prototype.hasOwnProperty.call(field.value, 'concealed')) {
```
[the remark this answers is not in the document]
BrunoBernardino: I generally go with this safer-if-more-verbose approach on frontend code (more easily tampered with), but we don't have to. You can see some info on why at https://eslint.org/docs/rules/no-prototype-builtins

packages/app/src/lib/1pux-parser.ts
```diff
@@ -0,0 +1,308 @@
...
+      rowData.extraFields.push({
+        name: field.title || field.id,
+        value,
+        type,
+      });
+    });
+  });
+
+  return rowData;
+};
```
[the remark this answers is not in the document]
BrunoBernardino: Sorry, I'll do this while implementing a command for it adding it to CI as well (addressing #322 partially), if that's OK with you.

packages/app/src/lib/import.ts
```diff
@@ -298,18 +298,21 @@ async function parse1PuxItem(accountName: string, vaultName: string, item: OnePu
...
 
     }
 
     return createVaultItem(itemName, fields, tags);
 }
 
-export async function as1Pux(file: string | Uint8Array): Promise<VaultItem[]> {
+export async function as1Pux(data: null | string | ArrayBuffer): Promise<VaultItem[]> {
+    if (!data) {
+        throw new Err(ErrorCode.INVALID_1PUX);
```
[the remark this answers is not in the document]
BrunoBernardino: Sounds perfect. I didn't an example for it before, thanks.

packages/app/src/lib/import.ts
```diff
@@ -8,6 +6,8 @@ import { Err, ErrorCode } from "@padloc/core/src/error";
 import { uuid } from "@padloc/core/src/util";
 import { translate as $l } from "@padloc/locale/src/translate";
 
+import { parse1PuxFile, parseToRowData, OnePuxItem } from "./1pux-parser";
```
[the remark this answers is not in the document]
BrunoBernardino: Great! I hadn't noticed dynamic imports used in this package before, so I wasn't sure we were properly supporting them.

packages/app/src/elements/import-dialog.ts
BrunoBernardino: The problem was that the `FileReader` and reading of the file had to be implemented/called in many more places (for each type of import). It just seemed less clear/readable to me, but I'm happy to implement it that way, if you prefer.

BrunoBernardino:
> Sorry if it seems like I'm doing a lot of bike shedding (..).

You're not. These things are important to get right initially, especially as we still don't have the automations in place to do it for us! I'll get on these next week.

packages/app/src/elements/import-dialog.ts
MaKleSoft: There are actually convenience functions for reading files already implemented in https://github.com/padloc/padloc/blob/master/packages/core/src/attachment.ts. You can move those to `@padloc/core/src/util` for reuse.

packages/app/src/lib/1pux-parser.ts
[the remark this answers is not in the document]
MaKleSoft: In this case (and probably in most other cases as well) I would prefer the more terse, readable version unless there is a concrete failure mode we're trying to avoid. In this case I don't see any real risk in simply checking for thruthyness.

packages/app/src/elements/import-dialog.ts
MaKleSoft: And while you're at it maybe also give them all consistent names (i.e. `readFileAs...`).

README.md
MaKleSoft: I think I'd prefer
 ```sh
npm run add/remove <package> [scope]
```
What do you think?

Comment (BrunoBernardino) [reply README.md:64]: Yeah, I tried that initially, but `npm` messes up `$2`, so it needs to be set as a variable before the script runs, unfortunately.

**Comment by BrunoBernardino** [comment]
Alright @MaKleSoft I've pushed another update, addressing all the requests and suggestions. The diff looks huge because `prettier` ran on it, and it had many files to change. Check it out.

File: package.json [hunk 20]
```diff
@@ -1,38 +1,43 @@
...
+        "electron:build": "cd packages/electron && npm run build && cd ../..",
+        "start": "npm run pwa:build && lerna run --scope '@padloc/{server,pwa}' --parallel start",
+        "dev": "lerna run --parallel --scope '@padloc/{server,pwa}' --parallel dev",
+        "tauri:dev": "lerna run --parallel --scope '@padloc/{server,tauri}' --parallel dev",
+        "repl": "cd packages/server && npm run repl && cd ../..",
+        "test": "lerna run test",
+        "locale:extract": "lerna run extract --scope '@padloc/locale'",
+        "add": "lerna add $1 --scope=@padloc/$scope",
+        "remove": "rm packages/$scope/package-lock.json && lerna exec \"npm uninstall $1\" --scope=@padloc/$scope",
+        "prettier": "prettier --write .",
```
BrunoBernardino, line 40: Now we can just run `npm run prettier` to get everything conforming.

packages/app/src/lib/import.ts
```diff
@@ -240,13 +251,118 @@ export async function asLastPass(data: string): Promise<VaultItem[]> {
...
-export function isLastPass(data: string): boolean {
-    return data.split("\n")[0] === "url,username,password,extra,name,grouping,fav";
+export function is1Pux(file: File): boolean {
+    return file.name.endsWith(".1pux");
 }
 
-export function guessFormat(data: string): ImportFormat | null {
-    return isPBES2Container(data) ? PBES2 : isPadlockV1(data) ? PADLOCK_LEGACY : isLastPass(data) ? LASTPASS : CSV;
+export async function guessFormat(file: File): Promise<ImportFormat> {
+    // Try to guess sync first (won't need parsing)
```
BrunoBernardino, line 353: So, one problem of moving to `file` being passed down and delaying the reading of files is that for ones that can't be distinguished from file extensions (like lastpass/csv), the parsing needs to happen multiple times, if it keeps failing to guess the format. It should not be slow or a big problem, though.

packages/app/src/lib/import.ts
```diff
@@ -139,16 +147,18 @@ export async function importLegacyContainer(container: PBES2Container) {
     return Promise.all(items);
 }
 
-export function isPBES2Container(data: string) {
+export async function isPBES2Container(file: File) {
```
BrunoBernardino, line 150: I could've made this one check the `.pbes2` file extension, but this is a better check and reading a file as text is pretty cheap. Let me know if you disagree, here.

packages/app/src/lib/import.ts
```diff
@@ -247,33 +251,37 @@ export async function asLastPass(data: string): Promise<VaultItem[]> {
     return Promise.all(items);
 }
 
-/**
- * Checks if a given string represents a LastPass CSV file
- */
-export function isLastPass(data: string): boolean {
+export async function isLastPass(file: File): Promise<boolean> {
     try {
```
MaKleSoft, line 255: Perhaps in the future we can check the mime type and extension of the file first before reading the file. For example in this case, if the mime type is anything other than `text/plain` or `text/csv`, we can safely discard this option. This can wait until we add more import formats though.

MaKleSoft (approved):
Boy, I didn't expect you to run the formatter on the entire code base 😅 There are fewer changes than I expected though plus it's good to have it out of the way. Looks good to merge!

Issue #330, opened 2021-12-21
Build automation using github actions


MaKleSoft:
Automatically build and sign targets, create release, attach signed and unsigned artifacts and (optionally) publish to respective stores.

- [x] Web Extension
- [x] Desktop (Electron)
- [x] Desktop (Tauri)
- [x] Android
- [x] iOS

PR #336, opened 2021-12-21
CI: Build Web Extension

BrunoBernardino:
This PR/branch implements the automatic building + archiving of the Web Extension (for Firefox and Chrome).

It won't publish to the stores just yet (I'd rather that be a separate, manually triggered GH action, that downloads one of the built artifacts and publishes it), and for now Firefox is disabled, until we have the secrets set in GitHub.

Related to #330

### DIFF
--- a/.github/workflows/build-web-extension.yml
+++ b/.github/workflows/build-web-extension.yml
@@ -45,5 +45,5 @@ jobs:
               uses: actions/upload-artifact@v2
               with:
                   name: unsigned-extension-chrome-${{ github.sha }}.zip
-                  path: chrome.zip
+                  path: packages/extension/chrome.zip
                   if-no-files-found: error
--- a/.github/workflows/build-web-extension.yml
+++ b/.github/workflows/build-web-extension.yml
@@ -47,7 +47,7 @@ jobs:
               uses: actions/upload-artifact@v2
               with:
                   name: signed-web-extension-firefox-${{ github.sha }}
-                  path: packages/extension/dist/*.xpi
+                  path: packages/extension/dist/web-ext-artifacts/*.xpi
                   if-no-files-found: error
             - name: Archive Unsigned Web Extension (Chrome)
               uses: actions/upload-artifact@v2
@@ -55,5 +55,7 @@ jobs:
                   name: unsigned-web-extension-chrome-${{ github.sha }}
                   path: |
                       packages/extension/dist
+                      !web-ext-artifacts
                       !*.xpi
+                      !*.zip
                   if-no-files-found: error
--- a/.github/workflows/build-web-extension.yml
+++ b/.github/workflows/build-web-extension.yml
@@ -38,13 +38,13 @@ jobs:
             - name: Archive Signed Web Extension (Firefox)
               uses: actions/upload-artifact@v2
               with:
-                  name: signed-web-extension-firefox-${{ github.sha }}
+                  name: padloc-web-extension-${{ github.sha }}-signed.xpi
                   path: packages/extension/dist/web-ext-artifacts/*.xpi
                   if-no-files-found: error
             - name: Archive Unsigned Web Extension (Chrome)
               uses: actions/upload-artifact@v2
               with:
-                  name: unsigned-web-extension-chrome-${{ github.sha }}
+                  name: padloc-web-extension-${{ github.sha }}-unsigned
                   path: |
                       packages/extension/dist
                       !packages/extension/dist/web-ext-artifacts
--- a/.github/workflows/build-web-extension.yml
+++ b/.github/workflows/build-web-extension.yml
@@ -21,7 +21,7 @@ on:
 jobs:
     archive:
         runs-on: ubuntu-latest
-        environment: ${{ github.event.inputs.environment || "Local" }}
+        environment: ${{ github.event.inputs.environment || "Local' }}
         steps:
             - uses: actions/checkout@v2
             - uses: actions/setup-node@v1
--- a/.github/workflows/build-web-extension.yml
+++ b/.github/workflows/build-web-extension.yml
@@ -55,7 +55,6 @@ jobs:
                   name: unsigned-web-extension-chrome-${{ github.sha }}
                   path: |
                       packages/extension/dist
-                      !web-ext-artifacts
-                      !*.xpi
-                      !*.zip
+                      !packages/extension/dist/web-ext-artifacts
+                      !packages/extension/dist/*.xpi
                   if-no-files-found: error
--- a/.github/workflows/build-web-extension.yml
+++ b/.github/workflows/build-web-extension.yml
@@ -3,15 +3,15 @@ name: Build Web Extension
 on:
     workflow_dispatch:
         inputs:
-            server_url_variable:
+            environment:
                 type: choice
-                description: "Server URL Variable"
-                default: "LOCAL_PL_SERVER_URL"
+                description: "Environment to build for"
+                default: "Local"
                 required: true
                 options:
-                    - "LOCAL_PL_SERVER_URL"
-                    - "BETA_PL_SERVER_URL"
-                    - "PRODUCTION_PL_SERVER_URL"
+                    - "Local"
+                    - "Beta"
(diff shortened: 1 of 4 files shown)


.github/workflows/build-web-extension.yml
```diff
@@ -0,0 +1,45 @@
...
+              with:
+                  node-version: 16.13.1
+            - name: Install dependencies
+              run: |
+                  npm ci
+                  npm i -g web-ext@6.6.0
+            - name: Build
+              run: npm run web-extension:build
+            # - name: Sign for Firefox
+            #   run: cd packages/extension/dist && web-ext sign --api-key=${{ secrets.WEB_EXT_API_KEY }} --api-secret=${{ secrets.WEB_EXT_API_SECRET }}
```
BrunoBernardino, line 31: @MaKleSoft can you please set these two secrets in GitHub so I can uncomment these two steps?

.github/workflows/build-web-extension.yml
```diff
@@ -0,0 +1,45 @@
+name: Build Web Extension
+
+on:
+    workflow_dispatch:
+
+    push:
+        branches:
```
BrunoBernardino, line 7: I'm not sure if you wanted to "restrict" automatic building to just the `master' (and `v4`, for now) branches, since it's a slow thing, and we have the option to build manually for any branch (via `workflow_dispatch` above).

.github/workflows/build-web-extension.yml
```diff
@@ -0,0 +1,45 @@
...
+            - "v4"
+            - "master"
+        paths:
```
BrunoBernardino, line 10: This makes it so that the automatic building only happens if any file in these paths changes. If you change something for the `cordova` or `server` packages, for example, there's no need to build the web extension again.

BrunoBernardino:
You can see an artifact for Chrome that works at https://github.com/padloc/padloc/actions/runs/1607275401 although it's trying to connect to a local web server. I'm guessing there are some environment variables I'm still missing. To be continued.

.github/workflows/build-web-extension.yml
```diff
@@ -0,0 +1,62 @@
...
+        inputs:
+            server_url_variable:
+                type: choice
+                description: "Server URL Variable"
+                default: "LOCAL_PL_SERVER_URL"
+                required: true
+                options:
+                    - "LOCAL_PL_SERVER_URL"
+                    - "BETA_PL_SERVER_URL"
+                    - "PRODUCTION_PL_SERVER_URL"
```
BrunoBernardino, line 14: @MaKleSoft I don't have permissions to add these secrets to GitHub, please do so.

.github/workflows/build-web-extension.yml
MaKleSoft: I think for a trigger semver tags make the most sense (e.g. `v4.0.0`). This gives us more control over when the build should be triggered plus it gives us a release to attach the artifacts to.

.github/workflows/build-web-extension.yml
MaKleSoft: If we trigger builds only on explicit releases/tags (as mentioned above), all targets should be rebuilt with a version bump regardless of whether they have changed or not. This gives us consistent versioning across targets and people only need to refer to the latest release to find whatever they're looking for.

.github/workflows/build-web-extension.yml
BrunoBernardino: Thanks for sharing your plan! A couple of things to note:

1. The artifacts are stored for up to 90 days, so they're not a good candidate/alternative for linking release/downloads. There's a download-artifact action which we can use to copy and store artifacts somewhere "permanently".
2. The best advantage of CI/CD is to really have these things be built to be tested easily on interim steps (the "continuous" part of the word), not just releases. I'd even say the release part is the least important here, and why I'd like to make it a separate GitHub action that only runs manually (unless we're releasing a new version "every day" or continuously -- in that case, it makes more sense to have it triggered automatically too).

So I think this will be good enough for us to test (when needed) the things we're working on, and then the release "button" can be pushed whenever there's a new release to be made.

Let me know if that makes sense or not, and if there's anything unclear about it.

.github/workflows/build-web-extension.yml
BrunoBernardino: Thanks for sharing your thought process here. As I mentioned above, I think it's best to keep this for the "continuous" part of CI/CD, and have a separate action (ideally just manual, for now, but can also be automatic, on tags) to release, which can do the process for all targets.

Let me know if you agree.

BrunoBernardino:
Alright, this is now working and building both extensions (signed for firefox and unsigned for chrome) automatically for local development, and can be manually triggered for another environment.

BrunoBernardino:
You can see the artifacts automatically built for local development in https://github.com/padloc/padloc/actions/runs/1610591969

BrunoBernardino:
To see/test the artifacts built for beta, download them from https://github.com/padloc/padloc/actions/runs/1614499755

BrunoBernardino:
Alright, I was trying to get the `workflow_dispatch` button to show in in the actions tab, but that won't work until we have a workflow with the same file name in the default branch.

We can still manually trigger this workflow (like I did for beta above) with:

```bash
curl --request POST \
  --url https://api.github.com/repos/padloc/padloc/actions/workflows/build-web-extension.yml/dispatches \
  --header 'Authorization: Basic [base64-encoded "username:personal-token"]' \
  --header 'Content-Type: application/json' \
  --data '{
	"ref": "feature/ci-build-web-extension",
	"inputs": {
		"environment": "Beta"
	}
}'
```

If that's too inconvenient, I can make a separate PR to add this workflow to the `master` branch, as a basic "echo" (that way, the button should show up, and upon choosing `v4` or a new branch branched from `v4`, run with the latest stuff).

MaKleSoft:
> If that's too inconvenient, I can make a separate PR to add this workflow to the master branch, as a basic "echo" (that way, the button should show up, and upon choosing v4 or a new branch branched from v4, run with the latest stuff).
Using the api is definitely good enough for now 👍

MaKleSoft (changes requested):
Looks great! I'd just like to make two minor changes in the naming of the artifacts:

First, the "raw", unsigned `.zip` archive should be compatible (as an unpacked extension) not just with chrome but with any browser supporting the web extension "standard". So let's simply call it `padloc-web-extension-${version}-unsigned.zip` (maybe we can also omit the "unsigned" since there is not going to be a signed version of the "raw" archive).

Second, if it's possible to upload the `.xpi` as-is without re-zipping it, let's do that. Then we can simply name it `padloc-web-extension-${version}-signed.xpi` since `.xpi` ending already indentifies it as a Firefox addon. If that's not possible, `padloc-web-extension-${version}-signed.xpi.zip` to make it clear the zip unarchives into a `.xpi` file.

Next step would be packaging the extension as a signed `.crx` file. Would you like to do that in a separate PR or keep this one open?

BrunoBernardino:
@MaKleSoft thanks! Unfortunately [the `upload-artifact` action only creates `zip` files for now](https://github.com/actions/upload-artifact). I'll make the name changes.

As for the `.crx` file, I can't seem to find a way to do it programmatically with [the latest chrome store documentation](https://developer.chrome.com/docs/webstore/publish), so if you have some information, please let me know, otherwise we might want to punt or create a specific issue to let users know it's not yet possible.

BrunoBernardino:
@MaKleSoft just pushed the name changes. Note I still kept the `commit.sha` instead of `version` because these artifacts are mostly for development. When we release them, they should be renamed + listed appropriately (which can still be done automatically).

MaKleSoft:
> As for the .crx file, I can't seem to find a way to do it programmatically with the latest chrome store documentation, so if you know how, please let me know, otherwise we might want to punt or create a specific issue to let users know it's not yet possible, but it seems unnecessary as it's more an administrative/devops thing and less user-facing.

This looks promising: https://github.com/cardinalby/webext-buildtools-chrome-crx-action

MaKleSoft (approved):
Good to merge 👍

PR #353, opened 2021-12-24
CI: Build Electron

BrunoBernardino:
This has the setup for building for Linux, macOS, and Windows, but still likely needs some more setup.

Related to #330

### DIFF
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -22,12 +22,12 @@ on:
             - "master"
             - "feature/**"
             - "fix/**"
-        paths:
-            - "packages/app/**"
-            - "packages/core/**"
-            - "packages/electron/**"
-            - "packages/pwa/**"
-            - "packages/locale/**"
+        # paths:
+        #     - "packages/app/**"
+        #     - "packages/core/**"
+        #     - "packages/electron/**"
+        #     - "packages/pwa/**"
+        #     - "packages/locale/**"
 
 env:
     GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -102,7 +102,6 @@ jobs:
               run: |
                   npm run electron:build
               env:
-                  CSC_IDENTITY_AUTO_DISCOVERY: false
                   CSC_LINK: ${{ secrets.PL_MACOS_SIGNING_CERT_BASE64 }}
                   CSC_KEY_PASSWORD: ${{ secrets.PL_MACOS_SIGNING_CERT_PASSWORD }}
                   PL_MACOS_NOTARIZE_APPLE_ID: ${{ secrets.PL_MACOS_NOTARIZE_APPLE_ID }}
--- a/.github/workflows/build-electron.yml
+++ b/.github/workflows/build-electron.yml
@@ -22,12 +22,12 @@ on:
             - "master"
             - "feature/**"
             - "fix/**"
-        # paths:
-        #     - "packages/app/**"
-        #     - "packages/core/**"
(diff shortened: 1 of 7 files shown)


BrunoBernardino:
Everything is being built. I can only test AppImage at the moment, but I suppose everything else is also working. Beta artifacts can be found at https://github.com/padloc/padloc/actions/runs/1620247852

packages/electron/package.json
```diff
@@ -13,7 +13,7 @@
         "npm": "8.2.0"
     },
     "scripts": {
-        "build": "rm -rf app && webpack && node ./prepare-build.js && electron-builder --config build/build.json",
+        "build": "rm -rf app build dist && webpack && node ./prepare-build.js && electron-builder --config build/build.json",
```
BrunoBernardino, line 16: These 2 directories are also ignored and rebuilt on building.

packages/cordova/package.json
```diff
@@ -78,9 +78,9 @@
     "scripts": {
         "build": "webpack",
         "start": "npm run build && cordova run",
-        "prepare": "npm run build && cordova prepare",
-        "build:android": "npm run prepare && cordova build android",
-        "build:android:signed": "npm run prepare && cordova build android --release -- --packageType=apk --keystore=./android-upload-key.keystore --storePassword=$PL_ANDROID_STORE_PASSWORD --alias=$PL_ANDROID_KEYSTORE_ALIAS --password=$PL_ANDROID_STORE_PASSWORD",
-        "build:ios": "npm run prepare && cordova build ios"
+        "prepare-build": "npm run build && cordova prepare",
```
BrunoBernardino, line 81: I forgot `prepare` was a reserved lifecycle event/script, so it was always running even if it wasn't necessary, which caused issues on windows. An automatic run for the cordova CI confirmed this still works fine.

package.json
```diff
@@ -28,7 +28,7 @@
         "pwa:start": "lerna run start --scope @padloc/pwa",
         "server:start": "lerna run start --scope @padloc/server --stream",
         "server:start-dry": "lerna run start-dry --stream --scope @padloc/server",
-        "electron:build": "cd packages/electron && npm run build && cd ../..",
+        "electron:build": "lerna run build --scope @padloc/electron",
```
BrunoBernardino, line 31: This just makes it consistent and links everything as expected.

MaKleSoft: I *think* there was a reason why I did it this way, but I can't remember it now. Since builds seem to be unaffected I assume this change is ok so let's keep it. I just hope this isn't one of those Chesterton's Fence moments 😄

.github/workflows/build-electron.yml
```diff
@@ -22,12 +22,9 @@ on:
...
             - "fix/**"
-        # paths:
-        #     - "packages/app/**"
-        #     - "packages/core/**"
-        #     - "packages/electron/**"
-        #     - "packages/pwa/**"
-        #     - "packages/locale/**"
+        paths:
+            - "packages/electron/**"
+            - "assets/**"
```
MaKleSoft, line 27: Removing all packages except electron (PWA is loaded remotely so the other packages are not involved in this build) and adding the `assets` dir since that contains the app icon and name and other metadata used during the build process.

MaKleSoft (approved):
Looks good! I took the liberty of adding a signed & notarized version for macos. And changed the `paths` property to a more appropriate value. Let's merge!

.github/workflows/build-electron.yml
BrunoBernardino: Good point! I forgot that this only really loads the PWA, so it doesn't need to be rebuilt unless something in itself actually changes.

BrunoBernardino:
Perfect, thanks!

Issue #281, opened 2020-07-19
Fully Functional Tauri Client


nothingismagick:
We would like to help make sure that Tauri builds properly, especially because we want to start using this sweet piece of awesomeness ourselves. It would include a proper GH Actions integration - [see this repo](https://github.com/tauri-apps/tauri-action). 

If you are so inclined, please feel invited get in touch on our discord, because we are considering making this project a new "special focus" project where the core team makes sure that your problems are not only resolved, but that your needs are met - even if that means API revisions or entirely new interfaces. Obviously we will only do this if you are interested.

At any rate, kudos and let's see how we can collaborate.

nothingismagick:
Check this out: 
https://tauri.studio/docs/usage/recipes/plugin/#writing-a-plugin (specifically the `init_script` part)

MaKleSoft:
@nothingismagick Thank you so much for reaching out! I was actually planning to get in touch with you to suggest some kind of collaboration, too 😄 Tauri looks like perfect fit for Padloc and I believe Padloc could be great proof of concept for Tauri in turn, especially considering your security-centric approach. Replacing our Electron and Cordova wrappers with Tauri could be a huge win for both security and performance and would allow us to streamline our build pipeline significantly. I'll be reaching out on Discord soon!

MaKleSoft:
I haven't looked at the Tauri integration in a while, but last time I checked there where still the following blockers:

- [ ] **Clipboard support**: The [Tauri roadmap](https://tauri.studio/en/#roadmap) says that clipboard support is still to come, but I'm hoping it's just not up to date. Hopefully just adding an application menu with the appropriate items will do the trick. Related issues: https://github.com/tauri-apps/tauri/issues/1055 https://github.com/tauri-apps/tauri/issues/2397. See also: https://tauri.studio/en/docs/usage/guides/visual/menu. We're also using [`document.execCommand("cut")`](https://github.com/padloc/padloc/blob/master/packages/app/src/lib/platform.ts#L37) and [`document.execCommand("paste")`](https://github.com/padloc/padloc/blob/master/packages/app/src/lib/platform.ts#L48) to access the clipboard programatically, so we'll have to check if that works in Tauri (it does in electron).
- [ ] **Storage**: Padloc uses [localForage](https://github.com/localForage/localForage) for local storage. Last time I checked there were problems with using "vanilla" localForage, but luckily there seems to be a plugin for that now: https://github.com/tauri-apps/tauri-forage
[...]

PR #357, opened 2021-12-25
CI: Build Tauri

BrunoBernardino:
Build Tauri for Desktop on code push.

Related to #330

Also tweak the other CI actions to use ENV instead of setting runtime variables.


.github/workflows/build-tauri.yml
```diff
@@ -1,5 +1,20 @@
...
+
+    push:
+        branches:
+            - "v4"
+            - "master"
+            - "feature/**"
+            - "fix/**"
+        paths:
+            - "packages/electron/**"
+            - "assets/**"
```
BrunoBernardino, line 17: @MaKleSoft I actually am not sure if these are enough (or if there's anything else.)

.github/workflows/build-cordova.yml
```diff
@@ -56,8 +56,11 @@ jobs:
                   base64 -d android-upload-key.keystore.txt > android-upload-key.keystore
                   rm -f android-upload-key.keystore.txt
             - name: Build (Android)
-              run: |
-                  PL_SERVER_URL='${{ secrets.PL_SERVER_URL }}' PL_ANDROID_STORE_PASSWORD='${{ secrets.PL_ANDROID_STORE_PASSWORD }}' PL_ANDROID_KEYSTORE_ALIAS='${{ secrets.PL_ANDROID_KEYSTORE_ALIAS }}' PL_ANDROID_STORE_PASSWORD='${{ secrets.PL_ANDROID_STORE_PASSWORD }}' npm run cordova:build:android:signed
+              env:
```
BrunoBernardino, line 59: These changes worked fine: https://github.com/padloc/padloc/actions/runs/1622702353

.github/workflows/build-web-extension.yml
```diff
@@ -32,7 +32,9 @@ jobs:
                   npm i -g npm@8.2.0 web-ext@6.6.0
                   npm ci
             - name: Build
-              run: PL_SERVER_URL=${{ secrets.PL_SERVER_URL }} npm run web-extension:build
+              env:
```
BrunoBernardino, line 35: These changes worked fine (beta): https://github.com/padloc/padloc/actions/runs/1622702248

BrunoBernardino:
CI failed, will need to investigate further tomorrow: https://github.com/padloc/padloc/runs/4633584260?check_suite_focus=true

.github/workflows/build-tauri.yml
MaKleSoft: The Tauri version actually builds and bundles the app code instead of loading the pwa remotely, se we do need the other packages (app, core, locale). Also, change `packages/electron` to `packages/tauri` obviously.

MaKleSoft:
@BrunoBernardino FYI, the existing action is quite old, and I'm not sure if it's still compatible with the latest version of Tauri. Check out https://github.com/tauri-apps/tauri-action for the latest and greatest.

MaKleSoft:
You may also want to make sure that the `@padloc/tauri` package itself uses the latest version of Tauri (see https://tauri.studio/en/docs/usage/development/updating).

BrunoBernardino:
This is ready for review, with the beta artifacts built at https://github.com/padloc/padloc/actions/runs/1623848511 (I only tested AppImage, which doesn't get past the v3 loading screen, which I assume needs to be fixed in #281 instead of this PR)

MaKleSoft (changes requested):
While you're at it, can you switch the `PL_SERVER_URL` and `PL_PWA_URL` variables to regular environment variables? These are not meant to be secret and should in fact be public for the sake of transparency.

Looks good otherwise 👍

BrunoBernardino:
@MaKleSoft I think it's best to discuss that in a separate issue, because [GitHub doesn't have a "great solution" for it](https://github.community/t/non-secret-variable-for-environments/152691/2), so we'd need to decide on how we wanted to do it (via `.env` files or some other, more descriptive way). If you agree, I'll just merge this one and we can create a separate issue for this.

MaKleSoft:
> I think it's best to discuss that in a separate issue, because GitHub doesn't have a "great solution" for it, so we'd need to decide on how we wanted to do it (via .env files or some other, more descriptive way). If you agree, I'll just merge this one and we can create a separate issue for this.

Huh, I didn't know Github environments only supported secrets. Seems like a huge oversight... Ok, let's merge.

MaKleSoft (approved):

PR #358, opened 2021-12-26
Add setting for disabling/enabling badge on extension icon

MaKleSoft:
Title says it all. Setting can be found under Settings/Extension.
Screenshot 2021-12-26 at 11 43 18

### DIFF
--- a/packages/app/src/elements/settings-extension.ts
+++ b/packages/app/src/elements/settings-extension.ts
@@ -6,21 +6,16 @@ import { router } from "../globals";
 import { translate as $l } from "@padloc/locale/src/translate";
 import { customElement, query } from "lit/decorators.js";
 import { shared } from "../styles";
-import { Select } from "./select";
-import { Settings } from "@padloc/core/src/app";
 import { ToggleButton } from "./toggle-button";
 import "./popover";
 import "./icon";
 
-@customElement("pl-settings-display")
-export class SettingsDisplay extends StateMixin(LitElement) {
+@customElement("pl-settings-extension")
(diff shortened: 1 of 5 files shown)


packages/app/src/elements/settings-extension.ts
```diff
@@ -0,0 +1,69 @@
...
+                </header>
+
+                <pl-scroller class="stretch">
+                    <div class="double-margined box">
+                        <h2 class="padded bg-dark border-bottom semibold horizontal center-aligning layout">
+                            <div class="uppercase stretch">${$l("Badge")}</div>
+                            <pl-icon icon="info-round" class="subtle"></pl-icon>
+                            <pl-popover trigger="hover" class="small double-padded regular" style="max-width: 20em">
+                                ${$l(
+                                    "If this option is enabled, the extension icon will show a badge with the number of matching items (if any) for the currently active tab. NOTE: Changing this setting can take up to a minute to take effect."
```
BrunoBernardino, line 49: Is there no way to force it to become instant, or is too much work?

BrunoBernardino (approved):
The changes look simple, readable and clear. Great work! I didn't pull this down and I only have a non-blocking question.

packages/app/src/elements/settings-extension.ts
MaKleSoft: There probably is, but it's not worth the trouble at this point, especially since this is a relatively "advanced" setting that probably won't be used by too many users and even then will likely ever only be changed once.

BrunoBernardino: Makes sense!

PR #367, opened 2021-12-29
CI: Cordova - Build iOS

BrunoBernardino:
This implements building the iOS package + uploading the artifact.

It builds successfully though I was unable to get past the loading screen, even locally (not sure what's wrong, might need help).

The QRScanner problem was a journey. It started with trying to install the package via url, from https://github.com/padloc/cordova-plugin-qrscanner and I ran into https://github.com/lerna/lerna/issues/2674 then I tried to simply publish a new package name (`padloc-cordova-plugin-qrscanner`), but as I was replacing the plugin name in the code, it started blowing up in a few different places, and it just felt wrong. Finally, I just created a patch command in `npm`, which can later evolve into a bash script or whatever we need. It's not pretty, but that worked!

The signed command doesn't run yet for iOS because building for the app store will be a separate PR, for publishing, like the other targets.

This will close the development/CI/CD part of #330

Beta artifacts will be at https://github.com/padloc/padloc/actions/runs/1635370785


BrunoBernardino:
The Beta APK works on my device, but on my emulators I get a blank screen past login. The Beta iOS app doesn't even show me the loading screen on my simulator, just a blank one. Thoughts welcome!

MaKleSoft:
I have no idea how to make an unsigned `.app` file run on the simulator and I don't think it's even possible on an actual device, so I can't test the output from the Github action, but if I run it locally on my iPhone it works fine! I also get the blank screen on the simulator though.

MaKleSoft (approved):
Added a few things of my own:
- Update `config.xml` with properties from `assets/manifest.json` as part of the build preparation
- Updated the csp to allow loading favicons

The app doesn't seem to work on the Simulator for me, either, but I wouldn't let this stop us from merging this. It works on a real device and that's the only thing that counts, really.

BrunoBernardino:
Ok, makes sense!

PR #371, opened 2021-12-31
Add Tauri auto-updates

### LINKED ISSUE #365: [Tauri] Implement auto updates
See https://tauri.studio/en/docs/usage/guides/updater

BrunoBernardino:
Closes #365

Still unfinished, I need to learn where are the build signatures stored.

### DIFF
--- a/.github/workflows/build-cordova.yml
+++ b/.github/workflows/build-cordova.yml
@@ -5,6 +5,7 @@ on:
         inputs:
             environment:
                 type: environment
+                default: "Local"
             targets:
                 type: choice
                 description: "Targets to build"
--- a/.github/workflows/build-tauri.yml
+++ b/.github/workflows/build-tauri.yml
@@ -55,11 +55,15 @@ jobs:
               run: npm run tauri:build
               env:
                   PL_SERVER_URL: ${{ secrets.PL_SERVER_URL }}
+                  TAURI_PRIVATE_KEY: ${{ secrets.TAURI_PRIVATE_KEY }}
+                  TAURI_KEY_PASSWORD: ${{ secrets.TAURI_KEY_PASSWORD }}
             - name: Build (production)
               if: github.event.inputs.environment == 'Production'
               run: npm run tauri:build:production
               env:
                   PL_SERVER_URL: ${{ secrets.PL_SERVER_URL }}
+                  TAURI_PRIVATE_KEY: ${{ secrets.TAURI_PRIVATE_KEY }}
+                  TAURI_KEY_PASSWORD: ${{ secrets.TAURI_KEY_PASSWORD }}
             - name: Archive AppImage
               uses: actions/upload-artifact@v2
               if: matrix.platform == 'ubuntu-latest'
--- a/.github/workflows/build-tauri.yml
+++ b/.github/workflows/build-tauri.yml
@@ -5,6 +5,7 @@ on:
         inputs:
             environment:
                 type: environment
+                default: "Local"
 
     push:
         branches:
--- a/.github/workflows/release-tauri.yml
+++ b/.github/workflows/release-tauri.yml
@@ -20,6 +20,7 @@ jobs:
 
         runs-on: ubuntu-latest
         steps:
+            - uses: actions/checkout@v2
             - uses: rickstaa/action-create-tag@v1
               with:
                   tag: "tauri-v${{ env.RELEASE_VERSION }}"
--- a/.github/workflows/release-tauri.yml
+++ b/.github/workflows/release-tauri.yml
@@ -45,11 +45,10 @@ jobs:
 
     release:
         environment: ${{ github.event.inputs.environment || 'Beta' }}
+        needs: create_release_tag
         env:
             RELEASE_VERSION: ${{ github.event.inputs.version || '0.0.1' }}
-            RELEASE_ID: ${{ jobs.create_release_tag.steps.create_release.outputs.id }}
-
-        needs: create_release_tag
+            RELEASE_ID: ${{ needs.create_release_tag.outputs.create_release.id }}
 
         strategy:
             fail-fast: false
--- a/.github/workflows/release-tauri.yml
+++ b/.github/workflows/release-tauri.yml
@@ -77,10 +77,23 @@ jobs:
               with:
                   releaseId: ${{ needs.create_release_tag.steps.create_release.outputs.id }}
                   projectPath: packages/tauri
+
+    add_tauri_update:
+        name: "Add Tauri Update"
+        environment: ${{ github.event.inputs.environment || 'Beta' }}
+        env:
+            RELEASE_VERSION: ${{ github.event.inputs.version || '0.0.1' }}
+
+        needs: release
+
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v2
             - name: Build tauri-update.json
               run: cd packages/tauri && node generate-tauri-update-file.js
             - name: Add tauri-update.json
               uses: softprops/action-gh-release@v1
               with:
+                  tag_name: "tauri-v${{ env.RELEASE_VERSION }}"
                   files: packages/tauri/tauri-update.json
                   fail_on_unmatched_files: true
--- a/.github/workflows/release-tauri.yml
+++ b/.github/workflows/release-tauri.yml
@@ -1,0 +1,58 @@
+name: Release Tauri
+
+on:
+    workflow_dispatch:
+        inputs:
+            environment:
+                type: environment
+                default: "Beta"
+
+jobs:
+    release:
+        environment: ${{ github.event.inputs.environment || 'Beta' }}
+
+        strategy:
+            fail-fast: false
+            matrix:
+                platform: [macos-latest, ubuntu-latest, windows-latest]
+
+        runs-on: ${{ matrix.platform }}
+        steps:
+            - uses: actions/checkout@v2
+            - uses: actions/setup-node@v2
+              with:
+                  node-version-file: ".nvmrc"
+            - name: Install Rust stable
+              uses: actions-rs/toolchain@v1
+              with:
+                  toolchain: stable
+            - name: Install tauri bundler
+              run: cargo install tauri-bundler --force
+            - name: Install webkit2gtk (ubuntu only)
+              if: matrix.platform == 'ubuntu-latest'
+              run: |
+                  sudo apt-get update
+                  sudo apt-get install -y webkit2gtk-4.0
+            - name: Install dependencies
+              run: |
+                  npm i -g npm@8.2.0
+                  npm ci
+            - uses: tauri-apps/tauri-action@v0.2.0
+              env:
+                  PL_SERVER_URL: ${{ secrets.PL_SERVER_URL }}
+                  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+                  TAURI_PRIVATE_KEY: ${{ secrets.TAURI_PRIVATE_KEY }}
+                  TAURI_KEY_PASSWORD: ${{ secrets.TAURI_KEY_PASSWORD }}
+              with:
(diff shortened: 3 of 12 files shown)


MaKleSoft:
> I need to learn where are the build signatures stored.
According to https://tauri.studio/en/docs/usage/guides/updater#update-file-json-format:

```json
{
    "name":"v1.0.0",
    "notes":"Test version",
    "pub_date":"2020-06-22T19:25:57Z",
    "platforms": {
        "darwin": {
            "signature":"", <-- Signature goes here
            "url":"https://github.com/lemarier/tauri-test/releases/download/v1.0.0/app.app.tar.gz"
        },
        "linux": {
            "signature":"", <-- Signature goes here
            "url":"https://github.com/lemarier/tauri-test/releases/download/v1.0.0/app.AppImage.tar.gz"
        },
        "win64": {
            "signature":"", <-- Signature goes here
            "url":"https://github.com/lemarier/tauri-test/releases/download/v1.0.0/app.x64.msi.zip"
        }
    }
}
```

BrunoBernardino:
This is now ready for review. The tag is at https://github.com/padloc/padloc/releases/tag/tauri-v4.0.0 and the release at https://github.com/padloc/padloc/releases/tag/untagged-c25cf3a6f57f16f96e85 though we should manually delete those before we merge, to avoid confusion.

.github/workflows/build-cordova.yml
```diff
@@ -5,6 +5,7 @@ on:
         inputs:
             environment:
                 type: environment
+                default: "Local"
```
BrunoBernardino, line 8: This is just for consistency, once this shows for manual triggering.

.github/workflows/build-tauri.yml
```diff
@@ -54,11 +55,15 @@ jobs:
               run: npm run tauri:build
               env:
                   PL_SERVER_URL: ${{ secrets.PL_SERVER_URL }}
+                  TAURI_PRIVATE_KEY: ${{ secrets.TAURI_PRIVATE_KEY }}
+                  TAURI_KEY_PASSWORD: ${{ secrets.TAURI_KEY_PASSWORD }}
```
BrunoBernardino, line 59: This signs the builds, now.

.github/workflows/release-tauri.yml
```diff
@@ -0,0 +1,89 @@
+name: Release Tauri
```
BrunoBernardino, line 1: This works and will create a separate tag and release for tauri, for now. We can, instead, simply eventually rename this action, in a follow-up PR, to "Release", and have it create the release and upload everything it can (electron, cordova, web extensions, etc.), automatically, and then we just need to "finalize it" (publish as not pre-release in GitHub), then download and upload the files to the respective stores.

.github/workflows/release-tauri.yml
```diff
@@ -0,0 +1,89 @@
...
+    workflow_dispatch:
+        inputs:
+            environment:
+                type: environment
+                default: "Beta"
+            version:
+                type: string
+                description: "Version (semver) for the release."
+                default: "0.0.1"
+                required: true
```
BrunoBernardino, line 13: We can try to deduce this from the files, but it's not a bad idea to manually set it, since we'll be creating a tag from this. At least for the first release or so.

packages/tauri/src-tauri/tauri.conf.json
```diff
@@ -57,7 +57,10 @@
             ]
         },
         "updater": {
-            "active": false
+            "active": true,
+            "endpoints": ["https://github.com/padloc/padloc/releases/download/tauri-latest/tauri-update.json"],
```
BrunoBernardino, line 61: This URL won't really exist until we publish this tag. We'll also need to manually tag something as latest (which we can do in a follow-up "release" action, which can be the existing one, renamed, which would delete the current latest and tag the new code as such). We might want to change this to `latest` instead of `tauri-latest` if we don't want to consider those separate releases.

packages/tauri/src-tauri/Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
 name = "Padloc"
-version = "0.1.0"
+version = "4.0.0"
```
MaKleSoft, line 3: I don't think we need to bother updating this field, since the version from `tauri.conf.json` take precedence over this one.

packages/tauri/src-tauri/Cargo.toml
```diff
@@ -1,12 +1,12 @@
 [package]
 name = "Padloc"
```
MaKleSoft, line 2: Let's change this to lowercase, to get rid of that warning during the build process.

packages/tauri/src-tauri/tauri.conf.json
MaKleSoft: Does this mean we have to publish a separate tag named `tauri-latest` (or simply `latest`), in addition to the regular releases? What is the advantage over simply using `https://github.com/padloc/padloc/releases/latest/download/tauri-update.json`, which automatically resolves to the latest release?

MaKleSoft (commented):
Great stuff! Have you been able to test this yet?

.github/workflows/release-tauri.yml
BrunoBernardino: It's not common practice, and I think a safer approach would be to confirm the versions in the code match what's being asked instead, but we can definitely have it update the files and commit, if that's easier.

packages/tauri/src-tauri/tauri.conf.json
BrunoBernardino: If we keep it as is, yes. Since this is not really part of "latest" right now, I kept it separate, but if we want to move forward with this (and keep Tauri as a target like Electron, for the same version), then we don't need this tag separation at all.

packages/tauri/generate-tauri-update-file.js
```diff
@@ -0,0 +1,41 @@
...
+            signature: "[URL]",
+            url: "[URL]",
+        },
+    },
+};
+
+const packageFileContents = readFileSync(packageFilePath, "utf-8");
+const package = JSON.parse(packageFileContents);
+
+const { version } = package;
```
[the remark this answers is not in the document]
BrunoBernardino: :facepalm:

packages/tauri/generate-tauri-update-file.js
```diff
@@ -0,0 +1,41 @@
...
+const now = new Date().toISOString();
+
+tauriUpdate.name = `v${version}`;
+tauriUpdate.pub_date = now;
+tauriUpdate.platforms.darwin.url = `https://github.com/padloc/padloc/releases/download/v${version}/Padloc.app.tar.gz`;
+tauriUpdate.platforms.darwin.signature = `${tauriUpdate.platforms.darwin.url}.sig`;
+tauriUpdate.platforms.linux.url = `https://github.com/padloc/padloc/releases/download/v${version}/padloc_${version}_amd64.AppImage.tar.gz`;
+tauriUpdate.platforms.linux.signature = `${tauriUpdate.platforms.linux.url}.sig`;
+tauriUpdate.platforms.win64.url = `https://github.com/padloc/padloc/releases/download/v${version}/Padloc_${version}_x64.msi.zip`;
+tauriUpdate.platforms.win64.signature = `${tauriUpdate.platforms.win64.url}.sig`;
```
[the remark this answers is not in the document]
BrunoBernardino: The main advantage here is that I can reuse the objects (like `${tauriUpdate.platforms.linux.url}.sig`), otherwise we'll have to duplicate the repo URL. Both are fine, it's mostly so we have a "template" to start with that we populate, which I think makes things clearer. I'm fine to do it either way, here. So, your call.

packages/tauri/src-tauri/Cargo.toml
BrunoBernardino: I have found that to _not_ be the case for linux, for some reason, only for macos/windows. Similarly, the file name is `padloc` in Linux, but `Padloc` in macos/windows, which is strange, but not a deal-breaker, since the app is still `Padloc` everywhere.

packages/tauri/src-tauri/Cargo.toml
BrunoBernardino: I considered it, but it makes the menu/window name in macOS show lowercase, which is inconsistent with their guidelines. I prefer to see that warning instead, but can change it!

.github/workflows/release-tauri.yml
MaKleSoft: I think I'd like to try automatically updating the files in the repo, simply because that'll safe us more step and we can be absolutely sure the version is consistent everywhere. Can you think of any reason this may cause problems?

packages/tauri/src-tauri/tauri.conf.json
MaKleSoft: I definitely want to have everything in a single release, as discussed above. Previously, the separate Tauri release was really just a way to make it very clear that the Tauri app was still super experimental. Going forward it should be released alongside everything else and eventually replace the Electron version altogether.

packages/tauri/generate-tauri-update-file.js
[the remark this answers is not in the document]
MaKleSoft: If you're worried about repetition, you could simply store the "base url" in a variable. For me, the "canonical form" of the whole thing would look something like this:

```js
const baseUrl = `https://github.com/padloc/padloc/download/v${version}/`
const tauriUpdate = {
    name: version,
    pub_date: new Date(),
    platforms: {
        darwin: {
            url: `${baseUrl}/Padloc.app.tar.gz`,
            signature: `${baseUrl}/Padloc.app.tar.gz.sig`,
        },
        linux: {
            url: `${baseUrl}/padloc_${version}_amd64.AppImage.tar.gz`,
            signature: `${baseUrl}/padloc_${version}_amd64.AppImage.tar.gz.sig`,
        },
        win64: {
            url: `${baseUrl}/Padloc_${version}_x64.msi.zip`,
            signature: `${baseUrl}/Padloc_${version}_x64.msi.zip.sig`,
        },
    },
};
```

But that's just my preference and I feel like we're getting dangerously close to shedding bikes again (especially since this is basically just a "support script"), so I'm totally fine with keeping it as-is 😊

packages/tauri/src-tauri/Cargo.toml
MaKleSoft: I think that's only if you're running it in dev mode. Once bundled the name in the app menu should be uppercase. Can you verify that?

MaKleSoft (approved):
This will probably need some tweaking in the future but I don't see any critical issues right now, and this is definitely a huge step in the right direction. Feel free to merge at your discretion!

packages/tauri/src-tauri/Cargo.toml
BrunoBernardino: I'll change and verify this later when I pull up my macbook again, if that's fine.

packages/tauri/generate-tauri-update-file.js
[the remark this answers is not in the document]
BrunoBernardino: I'll do it your way (except for `new Date()`, because I want to make sure it's clear what the expectation is, and if the object would be passed around before "stringified", it could be an `int` instead of `string`, and I don't like to promote that JS inconsistency).

packages/tauri/src-tauri/tauri.conf.json
BrunoBernardino: Ok. Since we already have a v3 release going and I don't want to create confusion, we'll continue with the `tauri-` prefix here for now. We'll remove the prefix once this script becomes a wider release, if that makes sense.

.github/workflows/release-tauri.yml
BrunoBernardino: Yes, it can more easily cause conflicts, and, for example, I can't test this without "adding noise" to the branch (though I can then force-push without that commit), but it shouldn't be a big problem, and I can appreciate the consistency ask.

I would, however, prefer to do it in a follow-up PR, once this script becomes a release for everything, if that's fine.

Issue #360, opened 2021-12-26
Expose some env variables in GitHub Actions


BrunoBernardino:
> (...) can you switch the `PL_SERVER_URL` and `PL_PWA_URL` variables to regular environment variables? These are not meant to be secret and should in fact be public for the sake of transparency.

The two most common workarounds (because [GitHub doesn't have a great solution for this](https://github.community/t/non-secret-variable-for-environments/152691/2)) are to have a `.github/*.env` per environment and load that via a command/step, or to have some scripting logic to define those variables in the actions themselves.

I personally don't think either are "pretty", but the first one is transparent and easy to see/check/update.

What do you think @MaKleSoft ?

MaKleSoft:
I'll have to think about this for a bit. Let's table this discussion for now and revisit it later.

PR #373, opened 2022-01-04
Build Signed CRX Web Extension

### LINKED ISSUE #330: Build automation using github actions

BrunoBernardino:
This is a complement to the already closed #330, to make it easier to test the dev builds of the extension in Chrome.

It also disables Firefox signed builds for the local environment, and keeps the identifier, which means it needs to update the version in the manifest to avoid errors.


BrunoBernardino:
Beta artifacts at https://github.com/padloc/padloc/actions/runs/1653227254 and CRX worked perfectly in chrome!

update-manifest-build.js
```diff
@@ -0,0 +1,12 @@
+const { resolve } = require("path");
+const { writeFileSync } = require("fs");
+
+const manifestFilePath = resolve(__dirname, "assets", "manifest.json");
+
+const manifest = require(manifestFilePath);
+
+const buildVersion = parseInt(new Date().getTime().toString().slice(-3), 10).toString();
```
BrunoBernardino, line 8: This convoluted thing ensures that the build will have at most 3 digits (since it can't have a leading 0, we `parseInt` after extracting the last 3), coming from the milliseconds.

.github/workflows/build-web-extension.yml
```diff
@@ -54,3 +58,20 @@ jobs:
                       !packages/extension/dist/web-ext-artifacts
                       !packages/extension/dist/*.xpi
                   if-no-files-found: error
+            - name: Build ZIP for signing Chrome Extension
+              uses: cardinalby/webext-buildtools-pack-extension-dir-action@v1
+              with:
+                  extensionDir: packages/extension/dist
+                  zipFilePath: packages/extension/chrome.zip
+                  zipIgnore: web-ext-artifacts/**|*.xpi
```
BrunoBernardino, line 66: Unfortunately building an artifact doesn't allow us to reuse the zip (it's created in a container), so we need to do it "again".

.github/workflows/build-web-extension.yml
```diff
@@ -36,10 +36,14 @@ jobs:
...
               env:
                   PL_SERVER_URL: ${{ secrets.PL_SERVER_URL }}
-              run: npm run web-extension:build
+              run: |
+                  node update-manifest-build.js
+                  npm run web-extension:build
             - name: Sign for Firefox
-              run: cd packages/extension/dist && web-ext sign --channel=unlisted --api-key=${{ secrets.PL_WEB_EXTENSION_FIREFOX_API_KEY }} --api-secret=${{ secrets.PL_WEB_EXTENSION_FIREFOX_API_SECRET }}
+              if: github.event.inputs.environment == 'Beta' || github.event.inputs.environment == 'Production'
+              run: cd packages/extension/dist && web-ext sign --id="{9e5ca7e8-2073-49d1-a93f-ac2f2088ff14}" --channel=unlisted --api-key=${{ secrets.PL_WEB_EXTENSION_FIREFOX_API_KEY }} --api-secret=${{ secrets.PL_WEB_EXTENSION_FIREFOX_API_SECRET }}
```
BrunoBernardino, line 44: I considered adding the ID to a secret, but this is publicly available and easily visible. It's also potentially better for someone searching for "where's an id from" to end up in this repo, I think. It can potentially be moved to a "visible env var" once we decide on something for #360

update-manifest-build.js
MaKleSoft: You know what, the extension manifest is generated on the fly during the build process anyway, so might as well set the build number there: https://github.com/padloc/padloc/blob/v4/packages/extension/webpack.config.js#L90

MaKleSoft (changes requested):
Looks good! The only change I'd like to see is setting the build number inside `packages/extension/webpack.config.js` so you can get rid of the `update-manifest-build.js` script.

BrunoBernardino:
Changes applied, all passed and looking good @MaKleSoft !

MaKleSoft (approved):
👍

PR #376, opened 2022-01-07
CI: Full Release

BrunoBernardino:
This implements a full release action that will take care of automating as much as possible, and creating a GitHub release with all the relevant artifacts.

Still a lot to do before it's testable.


package.json
```diff
@@ -49,6 +49,8 @@
         "add": "lerna add $1 --scope=@padloc/$scope",
         "remove": "rm packages/$scope/package-lock.json && lerna exec \"npm uninstall $1\" --scope=@padloc/$scope",
         "prettier": "prettier --write .",
-        "prettier:check": "prettier --check ."
+        "prettier:check": "prettier --check .",
+        "update-version": "lerna version $1",
```
MaKleSoft, line 53: Why not add `&& node update-release-version.js $1 $2` here? Then we wouldn't have to call it explicitly every time (I assume the two command would always be run together).

BrunoBernardino: Good point, I'll change that.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,163 @@
...
+        inputs:
+            environment:
+                type: environment
+                default: "Beta"
+            version:
+                type: string
+                description: "Version (semver) for the release."
+                default: "0.0.1"
+                required: true
+            build:
```
MaKleSoft, line 14: I'm not sure if it makes sense to include a build number in the release action. The build number was only ever intended to be used internally, as a way to distinguish iterations between releases. When we create a release, the version number gets a bump (i.e. v4.0.0 -> v4.0.1) which resets the build number, so it's always going to be `0`. And there is not going to be any releases between v4.0.1 and v4.0.2. Conversely, for our continuous builds (which are not releases) the version number will stay the same, and the build number should increase with every build. Does that make sense?

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,163 @@
...
+                  tag_name: "v${{ env.RELEASE_VERSION }}"
+                  name: "Padloc v${{ env.RELEASE_VERSION }}"
+                  body: "Padloc v${{ env.RELEASE_VERSION }}"
+                  prerelease: true
+                  draft: true
+                  files: packages/tauri/tauri-update.json
+                  fail_on_unmatched_files: true
+
+    publish_npm:
+        name: "Publish packages to npm"
```
MaKleSoft, line 89: Let's skip this for now. I'm not yet sure how the `@padloc/...` packages fit into the node ecosystem, and whether it will ever make sense to publish them to npm at all (some of them are published already, but not really used anywhere afaik).

.github/workflows/publish-release.yml
BrunoBernardino: Yes it does! I think it doesn't harm to let this be a parameter since it's actively used for the Firefox extension publishing (and Chrome, once uploaded), and just always let it be the default of 0. If you think it's confusing, I can just remove it entirely, though.

MaKleSoft:
I think I should explain the purpose of the `assets/manifest.json` file and the role it plays in the versioning of the app.

Starting with v4, the code base is doesn't contain any hardcoded user-facing references to the name "Padloc" or any other branding like the logo, icon or color scheme. These are all defined in the `assets` directory and injected into the app during build time.

The idea is that this assets directory can now be switched out (or loaded from an external source via `PL_ASSETS_DIR`) in order to give the app an entirely different branding, including the name, logo, icon, styling (to a certain extend) and yes, the versioning as well. Why the version? Well, if someone wanted to release a white-label version of the app to their own customer base, they probably wouldn't want to start with version 4.x.

This means that there are now two different versions. The "true" version, defined in the `package.json` files and the "vendor version", defined in the `manifest.json`. The "true" version is used internally for things like versioning data structures (see for example https://github.com/padloc/padloc/blob/v4/packages/core/src/migrations.ts). while the "vendor version" is what is ultimately visible to the end user.

Having said all that, defining the vendor version in the `assets/manifest.json` was kind of an interim solution, and probably not the best way to do it. With all the cool new tools we have in our arsenal now, I think there's a better way. Here is how I envision things going forward:

- The "true" version tracks the progress of the project in general and is the same across all "vendors". It is tracked in the root `package.json` and the `package.json` of all the packages. To update it, we can simply use `lerna version`.
- The "vendor version" is used everywhere user-facing, which includes releases on Github, app store listings etc. Instead of being tracked in the code base this should be provided during build time and automatically inserted in the all the necessary places going forward. Namely:
    - All the visible occurrences of the version within the app itself. This happens via the [webpack environment plugin](https://webpack.js.org/plugins/environment-plugin/).
    - In `packages/tauri/src-tauri/tauri.conf.json`
    - In `packages/cordova/config.xml`
    - In the web extension manifest, which is [generated during build time](https://github.com/padloc/padloc/blob/v4/packages/extension/webpack.config.js#L89-L107).
    - In the electron package.json, also [generated during build time](https://github.com/padloc/padloc/blob/v4/packages/electron/webpack.config.js#L108-L116).
    - One or two other places I'm probably forgetting.
- This repo / code base is considered the "canonical" version of the project, which means the vendor version is always identical to the true version.

.github/workflows/publish-release.yml
MaKleSoft: I guess you're right that it doesn't hurt, especially if it's optional. So let's keep it in for now.

MaKleSoft:
So here is what I think we should do, specifically:

- Remove the `version` and `build` parameters from `assets/manifest.json`.
- Define an environment variable, `PL_VENDOR_VERSION` (and perhaps also `PL_VENDOR_BUILD` but let's talk about that later)
- During build time, inject the version defined in `PL_VENDOR_VERSION` into all the necessary places.
- In the release action, allow specifying an optional vendor version in addition to the "true" version. If no vendor version is specified, default to the "true" version.
- This means that in the release action, we only have to update the `package.json` files. The vendor version will only ever "manifest" in the build artifacts, plus the release name.

Does that make sense?

packages/cordova/update-config-xml.js
```diff
@@ -2,20 +2,28 @@ const { resolve, join } = require("path");
 const fs = require("fs");
 const { xml2js, js2xml } = require("xml-js");
 
+require("dotenv").config();
+
 const rootDir = resolve(__dirname, "../..");
 const assetsDir = resolve(rootDir, process.env.PL_ASSETS_DIR || "assets");
```
BrunoBernardino, line 8: `process.env.*` was always empty without the `dotenv` above.

packages/tauri/build-tauri-conf.js
```diff
@@ -0,0 +1,40 @@
+const { resolve } = require("path");
```
BrunoBernardino, line 1: This is a new file that takes care of making sure `tauri` builds have the vendor `version`, `name`, and a few other things (`author`, for example). It updates the `tauri.conf.json` and the `Cargo.toml` files.

packages/tauri/build-tauri-conf.js
```diff
@@ -0,0 +1,40 @@
...
+const manifestFilePath = resolve(assetsDir, "manifest.json");
+
+const tauriConfig = require(tauriConfigFilePath);
+const { version } = require(packageFilePath);
+const manifest = require(manifestFilePath);
+
+const vendorVersion = process.env.PL_VENDOR_VERSION || version;
+const vendorName = manifest.name;
+const vendorNameLowercase = vendorName.toLowerCase();
+const vendorBaseUrl = process.env.PL_VENDOR_BASE_URL || "https://github.com/padloc/padloc";
```
BrunoBernardino, line 22: This is a new variable that I haven't set in GitHub actions because it'll be the default in this repo.

packages/tauri/package.json
```diff
@@ -26,6 +28,7 @@
         "webpack": "5.52.0",
         "webpack-cli": "4.9.1",
         "webpack-dev-server": "4.7.1",
+        "@iarna/toml": "2.2.5",
```
BrunoBernardino, line 31: This was the smallest and easiest package I found to parse + generate `toml` files. It's also only used to build things, not in the built app.

BrunoBernardino:
@MaKleSoft what you said made perfect sense and I already pushed updates for all of that. What's missing right now is publishing the cordova artifacts in a release, and then testing the whole thing, which I imagine will have many issues and take a while to get right. It'll also involve some manual work to undo the automatic commit + tagging this action will do, while testing.

BrunoBernardino:
Lovely! This is almost ready for review.

Screenshot from 2022-01-11 19-12-13
Screenshot from 2022-01-11 19-12-31

@MaKleSoft I'm blocked right now, so you'll need to do 3 things before I can move on with this:

1. Delete all the 4.0.0.x releases from Firefox add-ons, just so I can start with 0 again.
2. Add `PL_IOS_CODE_SIGN_IDENTITY` and `PL_IOS_DEVELOPMENT_TEAM` as repository secrets (or share them with me via Padloc).
3. Go into the Apple Developer center and sign whatever new contracts are there.

packages/extension/webpack.config.js
```diff
@@ -89,11 +89,12 @@ module.exports = {
...
                             ...manifest,
-                            version:
-                                process.env.PL_BUILD_ENV === "Production"
-                                    ? build
-                                    : `${vendorVersion}.${devBuildVersion}`,
-                            version_name: vendorVersion,
+                            version: `${process.env.PL_VENDOR_VERSION || version}.${
+                                ["Production", "Beta"].includes(process.env.PL_BUILD_ENV)
+                                    ? process.env.RELEASE_BUILD
+                                    : devBuildVersion
```
MaKleSoft, line 95: It would really be a lot nicer if we could have the build number increment with each build. Is there a way to store the last build number somewhere and update it on every build without introducing a lot of noise to the commit history? Maybe with [git notes](https://git-scm.com/docs/git-notes)? Or a tag? Or within a text file "attached" to the last build as an artifact? Some place where it can easily be retrieved, incremented and updated from the github action.

BrunoBernardino: It's not easily possible and I think that's _way_ overkill. I'd be fine with just deleting the `devBuildVersion` logic since we no longer run these automatically.

A good way would be to have Firefox tell us which was the last version, but I can't find anyway to get that information for unlisted builds.

.github/workflows/publish-release.yml
```diff
@@ -226,6 +226,12 @@ jobs:
                   GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
                   TAURI_PRIVATE_KEY: ${{ secrets.TAURI_PRIVATE_KEY }}
                   TAURI_KEY_PASSWORD: ${{ secrets.TAURI_KEY_PASSWORD }}
+                  ENABLE_CODE_SIGNING: true
+                  APPLE_CERTIFICATE: ${{ secrets.PL_MACOS_SIGNING_CERT_BASE64 }}
+                  APPLE_CERTIFICATE_PASSWORD: ${{ secrets.PL_MACOS_SIGNING_CERT_PASSWORD }}
+                  APPLE_SIGNING_IDENTITY: ${{ secrets.PL_MACOS_SIGNING_IDENTITY }}
+                  APPLE_ID: ${{ secrets.PL_MACOS_NOTARIZE_APPLE_ID }}
+                  APPLE_PASSWORD: ${{ secrets.PL_MACOS_NOTARIZE_PASSWORD }}
```
BrunoBernardino, line 234: This is actually new stuff from today. I updated the commit instead of creating a new one by mistake.

packages/extension/webpack.config.js
```diff
@@ -84,16 +84,10 @@ module.exports = {
...
-
                     const jsonString = JSON.stringify(
                         {
                             ...manifest,
-                            version: `${process.env.PL_VENDOR_VERSION || version}.${
-                                ["Production", "Beta"].includes(process.env.PL_BUILD_ENV)
-                                    ? process.env.RELEASE_BUILD
-                                    : devBuildVersion
-                            }`,
+                            version: `${process.env.PL_VENDOR_VERSION || version}.${process.env.PL_BUILD_ENV || 0}`,
```
BrunoBernardino, line 90: Same as above.

packages/extension/webpack.config.js
MaKleSoft: Fair enough. We certainly shouldn't over-engineer it just because of Firefox' stupid rules.

BrunoBernardino:
Last run is at https://github.com/padloc/padloc/runs/4796006338 where iOS is complaining about missing a development team. I _think_ it's because I need the mobile provision file, since we're passing the team id, but it's also possible there's just some param mismatch.

BrunoBernardino:
Ok @MaKleSoft. I've skipped iOS for now and brought back everything else. I'm doing a new push that will create a new release + tag to do a final test, but otherwise this is ready for review.

BrunoBernardino:
Alright @MaKleSoft ! A final run with every target built (including iOS) is now at https://github.com/padloc/padloc/actions/runs/1704077589 with the target artifacts available at https://github.com/padloc/padloc/releases/tag/v4.0.0 !

BrunoBernardino:
@MaKleSoft it's now ready without automatically committing. https://github.com/padloc/padloc/actions/runs/1704809959 and https://github.com/padloc/padloc/releases/tag/v4.0.0

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,358 @@
...
+    workflow_dispatch:
+        inputs:
+            environment:
+                type: environment
+                default: "Beta"
+            version:
+                type: string
+                description: "Version (semver) for the release."
+                default: "0.0.1"
+                required: true
```
MaKleSoft, line 13: Do we still need this input, now that we're updating the version manually?

MaKleSoft (changes requested):
Looking good! Just a few minor things (see my comments):
- remove version input
- remove xcode setup
- simplify cordova build command

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,405 @@
...
+                  security import $DIST_CERT_PATH -P "$PL_IOS_DISTRIBUTION_CERT_PASSWORD" -A -t cert -f pkcs12 -k $KEYCHAIN_PATH
+                  security import $DEV_CERT_PATH -P "$PL_IOS_DISTRIBUTION_CERT_PASSWORD" -A -t cert -f pkcs12 -k $KEYCHAIN_PATH
+                  security list-keychain -d user -s $KEYCHAIN_PATH
+                  security set-key-partition-list -S apple-tool:,apple: -s -k "$PL_IOS_KEYCHAIN_PASSWORD" $KEYCHAIN_PATH
+
+                  # apply provisioning profile
+                  mkdir -p ~/Library/MobileDevice/Provisioning\ Profiles
+                  cp $DIST_PP_PATH ~/Library/MobileDevice/Provisioning\ Profiles
+                  cp $DEV_PP_PATH ~/Library/MobileDevice/Provisioning\ Profiles
+            - name: Setup Xcode
```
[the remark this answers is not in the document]
BrunoBernardino: This helps make sure we always run a consistent Xcode version. It's a kind of dependency version locking. I would recommend we keep it.

.github/workflows/publish-release.yml
```diff
@@ -0,0 +1,405 @@
...
+              if: matrix.platform == 'macos-latest'
+              with:
+                  xcode-version: "13.2.1"
+            - name: Build iOS
+              if: matrix.platform == 'macos-latest'
+              run: |
+                  cd packages/cordova
+                  npm run prepare-build
+                  npm run patch-ios
+                  ./node_modules/.bin/cordova build ios --release --device --automaticProvisioning --packageType=app-store --codeSignIdentity="${{ secrets.PL_IOS_CODE_SIGN_IDENTITY }}" --developmentTeam="${{ secrets.PL_IOS_DEVELOPMENT_TEAM }}" --buildFlag="-allowProvisioningUpdates" --buildFlag="DEVELOPMENT_TEAM=${{ secrets.PL_IOS_DEVELOPMENT_TEAM }}"
```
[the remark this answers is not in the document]
BrunoBernardino: Ok, I'll try that.

.github/workflows/publish-release.yml
BrunoBernardino: You're right, we shouldn't need this.

.github/workflows/publish-release.yml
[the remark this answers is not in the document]
MaKleSoft: Ah, ok. That makes sense 👍

BrunoBernardino:
@MaKleSoft Changes applied, release fully green at https://github.com/padloc/padloc/runs/1708896046 and https://github.com/padloc/padloc/releases/tag/v4.0.0

I also noticed the electron targets weren't being added to the release, and fixed that, so a full release should have 28 assets now.

MaKleSoft (approved):
Awesome! Let's merge this puppy.

PR #387, opened 2022-01-16
Better preserve context in extension when closing/opening popup

MaKleSoft:
- Store search string in url when leaving input
- Restore search when navigating to items view and `search` param is set.
- Only redirect to vault items matching current url if there are **new** matching vault items, i.e. if the page context changes.

### DIFF
--- a/packages/app/src/elements/app.ts
+++ b/packages/app/src/elements/app.ts
@@ -63,7 +63,7 @@ export class App extends ServiceWorker(StateMixin(AutoSync(ErrorHandling(AutoLoc
     ];
 
     @state()
-    private _page: string = "start";
+    protected _page: string = "start";
 
     @state()
     private _menuOpen: boolean = false;
--- a/packages/app/src/elements/items-list.ts
+++ b/packages/app/src/elements/items-list.ts
@@ -450,6 +450,14 @@ export class ItemsList extends StateMixin(LitElement) {
         this._listItems = this._getItems();
     }, 50);
 
+    private _updateFilterParam() {
+        const { search, ...params } = router.params;
+        if (this._filterInput.value) {
+            params.search = this._filterInput.value;
+        }
+        router.go(router.path, params, true);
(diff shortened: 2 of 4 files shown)


MaKleSoft:
@BrunoBernardino This addresses some of the annoyances in the web extension that we talked about. This was a little tricky to get right as there are a few conflicting behaviours, so it would be great if you could give it a spin before we'll merge this. A beta build can be found here: https://github.com/padloc/padloc/actions/runs/1704342780

BrunoBernardino:
@MaKleSoft the code changes look good, but I'm not seeing it behave as expected.

For example I'm in GitHub, open the extension, and there are a few items matching. I search for something else that's not in GitHub (say, a credit card), copy its number, leave the extension (click anywhere on github), and then when I open it again, it's back on the github matching items.

That is the common one that bugs me.

Should I test some other scenario?

MaKleSoft:
> For example I'm in GitHub, open the extension, and there are a few items matching. I search for something else that's not in GitHub (say, a credit card), copy its number, leave the extension (click anywhere on github), and then when I open it again, it's back on the github matching items.

Good catch! That was one of the scenarios that I wanted to cover but somehow forgot about it amidst the other changes. Should be fixed now. New beta build will be here: https://github.com/padloc/padloc/actions/runs/1707047921

In general, the intended behaviour is as follows

- When opening the popup, the previous navigation context should be restored, including any active search/filter within the list view
- That context will be overwritten with a list of items matching the current tab url, UNLESS any of the following is true
    - There are no such matching items
    - The list of matching items is identical to the last time the popup was opened (same tab)
    - The user was previously on any other view than the list view
    - The user was previously on the list view but had an active search

BrunoBernardino:
Yes! I just tested this new scenario and it seems to work well. I haven't covered many more yet, but I'll report any problems I find.

BrunoBernardino (approved):

PR #390, opened 2022-01-19
Fix tests

BrunoBernardino:
This fixes the existing tests and enables them for the "basic" CI run on every push.

Some tests/files had to be completely commented out because they're very outdated, and I'd rather fix those in separate PRs.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -19,11 +19,11 @@
         "@types/mocha": "8.2.2",
         "chai": "4.3.4",
         "mocha": "8.4.0",
-        "ts-node": "10.0.0",
+        "ts-node": "10.1.0",
         "typedoc": "0.22.4"
     },
     "scripts": {
-        "test": "cd test && mocha -r ts-node/register *.ts",
+        "test": "tsc --noEmit && cd test && mocha -r ts-node/register *.ts",
         "docs": "typedoc --mode modules --out docs"
     },
     "repository": {
--- a/packages/core/src/spec/app.ts
+++ b/packages/core/src/spec/app.ts
@@ -1,295 +1,295 @@
-import { App, AppState } from "../app";
-import { Server, ServerConfig } from "../server";
-import { StubMessenger } from "../messenger";
-import { MFAMessage, InviteCreatedMessage, MemberAddedMessage } from "../messages";
-import { DirectSender } from "../transport";
-import { MemoryStorage } from "../storage";
-import { MemoryAttachmentStorage } from "../attachment";
-import { ErrorCode } from "../error";
-import { OrgType } from "../org";
-import { Logger } from "../log";
-import { MFAPurpose } from "../mfa";
-import { Spec, assertResolve, assertReject } from "./spec";
+// import { App, AppState } from "../app";
+// import { Server, ServerConfig } from "../server";
+// import { StubMessenger } from "../messenger";
+// import { MFAMessage, InviteCreatedMessage, MemberAddedMessage } from "../messages";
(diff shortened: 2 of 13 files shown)


packages/core/package.json
```diff
@@ -19,11 +19,11 @@
         "@types/mocha": "8.2.2",
         "chai": "4.3.4",
         "mocha": "8.4.0",
-        "ts-node": "10.0.0",
+        "ts-node": "10.1.0",
         "typedoc": "0.22.4"
     },
     "scripts": {
-        "test": "cd test && mocha -r ts-node/register *.ts",
+        "test": "tsc --noEmit && cd test && mocha -r ts-node/register *.ts",
```
BrunoBernardino, line 26: `tsc` is faster to report type errors than `ts-node`. This is actually a recommendation from `ts-node`.

packages/core/tsconfig.json
```diff
@@ -2,11 +2,14 @@
...
         "outDir": "lib",
         "module": "esnext"
     },
     "include": ["src/**/*.ts"],
     "exclude": ["node_modules/**/*.ts"],
-    "sourceMap": true
+    "sourceMap": true,
+    "ts-node": {
+        "transpileOnly": true
+    }
```
BrunoBernardino, line 14: This disables the type checking in `ts-node`, because we run `tsc` now.

packages/core/tsconfig.json
```diff
@@ -2,11 +2,14 @@
     "extends": "../../tsconfig.json",
     "compilerOptions": {
         "baseUrl": ".",
-        "rootDir": "src",
+        "rootDir": ".",
```
BrunoBernardino, line 5: Without this change, the `vendor` files wouldn't properly load, thus throw the error we were seeing.

packages/server/package.json
```diff
@@ -49,7 +49,7 @@
         "repl": "ts-node src/init-repl-client.ts",
         "dev": "ts-node-dev src/init.ts",
         "dev-inspect": "node -r ts-node/register --inspect-brk --stack-trace-limit=1000 src/init.ts",
-        "test": "cd test && mocha -r ts-node/register *.ts --timeout 5000"
+        "test": "tsc --noEmit && cd test && mocha -r ts-node/register *.ts --timeout 5000"
```
BrunoBernardino, line 52: Same as above.

packages/server/src/init-repl-client.ts
```diff
@@ -1,9 +1,9 @@
-import { ReplClient } from "./repl";
+// import { ReplClient } from "./repl";
```
BrunoBernardino, line 1: This was calling things that eventually called files and code that doesn't exist anymore.

MaKleSoft (approved):
Looks good! 👍

packages/core/src/spec/app.ts
```diff
@@ -1,295 +1,295 @@
...
+// import { App, AppState } from "../app";
+// import { Server, ServerConfig } from "../server";
+// import { StubMessenger } from "../messenger";
+// import { MFAMessage, InviteCreatedMessage, MemberAddedMessage } from "../messages";
+// import { DirectSender } from "../transport";
+// import { MemoryStorage } from "../storage";
+// import { MemoryAttachmentStorage } from "../attachment";
+// import { ErrorCode } from "../error";
+// import { OrgType } from "../org";
+// import { Logger } from "../logging";
```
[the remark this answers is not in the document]
MaKleSoft: Yep, that was definitely the right call 👍

PR #393, opened 2022-01-21
First e2e test (via cypress)

BrunoBernardino:
This adds the first e2e test via [cypress](https://cypress.io). It's still incomplete, but it was a bit of a pain to figure out the shadow dom stuff.

The next step is to force a specific verification code when running via tests so the test can progress.


cypress/support/commands.ts
```diff
@@ -0,0 +1,26 @@
+Cypress.Commands.add("signup", () => {
+    cy.visit("/");
+
+    const { email, password } = Cypress.env();
+
+    cy.get("pl-app").find("pl-start").find("pl-login-signup").find("pl-input#emailInput").find("input").type(email);
```
BrunoBernardino, line 6: Once I have the whole login passing, I'll spend some time investigating a less verbose way of doing this. My initial tries (just `.get('pl-app pl-start pl-login-signup...')`) didn't work, unfortunately. I also had spent quite some time properly adding `.shadow()` in the right places before realizing `includeShadowDom` was a thing, now.

package.json
```diff
@@ -16,10 +16,13 @@
     },
     "main": "main.js",
     "devDependencies": {
+        "concurrently": "7.0.0",
```
BrunoBernardino, line 19: I hope these 3 dependencies are acceptable because they make things more reliable (starting server + app + cypress without `concurrently` is hard, and starting `cypress run` before making sure the app and server are ready -- `wait-on` -- can lead to timeouts).

BrunoBernardino:
@MaKleSoft  I finished setting up the signup and login tests, but there's still a problem of not being able to always run from a clean slate because the `packages/server/data` folder is always being used (I set `PL_DATA_DIR` to `./test-data` but that doesn't get created/used), and the browser appears to keep some state, somehow. Is there something more than `localStorage` and/or `sessionStorage` being used? Cypress clears those, but it’s remembering my session/device between runs after deleting `packages/server/data`, so I have to manually open, reset my session, and only then can I run the tests successfully.

MaKleSoft:
> (I set PL_DATA_DIR to ./test-data but that doesn't get created/used)

That env variable is no longer supported. The new config system uses a more complex system of env variables that allow specifying both the storage backend and associated configuration. The config variables are names-paced based on the structure of the config object defined here: https://github.com/padloc/padloc/blob/v4/packages/server/src/config.ts#L124

For leveldb, the variables looks like this:

```sh
PL_DATA_BACKEND=leveldb
PL_DATA_LEVELDB_DIR=./test-data
```

But for testing purposes, it's probably easier to just use the `MemoryStorage` backend, which was implemented for use cases just like this:

```sh
PL_DATA_BACKEND=memory
```

MaKleSoft:
> Is there something more than localStorage and/or sessionStorage being used
On the client side we use [localForage](https://github.com/localForage/localForage), which should be using IndexedDB as the default driver, falling back to localStorage on browsers that don't support it.

packages/core/src/auth/email.ts
```diff
@@ -15,6 +15,10 @@ export class EmailAuthServer implements AuthServer {
         authenticator.state = { email };
         if (authenticator.status !== AuthenticatorStatus.Active) {
             authenticator.state.activationCode = await this._generateCode();
+            // Override token if running e2e tests
+            if (process.env.PL_E2E_TESTS && process.env.PL_E2E_TESTS === "true" && email === "user@example.com") {
+                authenticator.state.activationCode = "000000";
+            }
```
MaKleSoft, line 21: I know this is probably just a stop-gap solution, but the spirit of a true e2e test, it would be better to find a way to simulate the code being delivered to the client instead bypassing the mechanism. I would also like to avoid having testing-related code inside the components to be tested.

In the old integration tests, I was using the [`StubMessenger`](https://github.com/padloc/padloc/blob/v4/packages/core/src/messenger.ts#L85) email backend to store the code in memory and then retrieve it later: https://github.com/padloc/padloc/blob/master/packages/core/src/spec/app.ts#L53
The "stub" backend isn't currently supported by the new config scheme but it could easily be added here: https://github.com/padloc/padloc/blob/v4/packages/server/src/init.ts#L116

Of course this only works if the server memory is accessible by the test code, which I don't know whether that's the case here. Alternatively, we could use the [ConsoleMessenger](https://github.com/padloc/padloc/blob/v4/packages/core/src/messenger.ts#L105) backend and parse the code from stdout (again don't know if that's feasible here).

You can use the `PL_EMAIL_BACKEND` variable to set the email driver.

BrunoBernardino: I understand what you mean and generally agree, and I'd add it's better to have as little differences in the code as possible.

This still runs everything as "in reality", except we simply force the code to be known.

I very much prefer something like this over mocking a whole piece of technology or using a different one. Generally, the less differences between a test and "reality", the better.

Does that make sense?

MaKleSoft: I hear what you're saying, but I don't agree with your conclusion. I agree that mocking is often overused, but the fact is we need **some** kind of mocking no matter what (unless we want to go as far as actually fetching and parsing emails from a real mailbox). In this case you're mocking the generation of activation code, whereas I would rather mock the delivery mechanism.

"You have to die one kind of death", as we say in Germany, and in this case I do think that mocking the email backend is the less painful one. Despite the shamefully bad test coverage, Padloc was actually designed to make this kind of mocking fairly trivial (in this case it is really just a configuration change), and I think we should use this to our advantage 😊

BrunoBernardino: Thanks @MaKleSoft ! I understand your point, and while I don't agree, I'd rather move on with the useful part here (the testing) instead of spending more time in the implementation _for now_, unless you disagree.

From my understanding, mocking the delivery mechanism + reading from it will not be trivial _at all_ (and I'll probably need to implement some new endpoint to do so, because the browser can't access the server's memory otherwise), and it'll likely take a few days to do, whereas right now _that part is working_, just in a different way than your preferred, and it isn't exposing any security vulnerability since it's checked from an environment variable + hardcoded test email.

I'm happy to implement your preferred mocking method later, after the test itself is working and verified, if you agree?

MaKleSoft: Oh I totally agree that we can leave this as-is for now, especially if my suggested method turns out to come with a significant overhead. Like I said, I don't even know if my approach is feasible at all in this case, since it would require either direct access to the server or some way to read stdout. Let's revisit this discussion once the PR is ready for review 👍

BrunoBernardino:
Alright, the signup test is working all the time, now (using memory as backend). There was also a race condition for the rendering of the "weak password" alert dialog which caused the browser to pass most of the times, but not headless.

I added a few more commands (clear indexedDB, unlock and lock), but the login is still throwing somewhere, and I haven't investigated that fully just yet. Those tests are currently commented out.

BrunoBernardino:
I think I'd prefer to merge this PR as is, with the login/lock/unlock tests incomplete (not running) because it's already quite big and already adds one first test (signup).

MaKleSoft:
Fails for me locally. Any idea why this could be?
Screenshot 2022-01-27 at 09 51 29

MaKleSoft:
Also failing in the visual test runner, but for a different reason

Screenshot 2022-01-27 at 10 20 54

MaKleSoft:
> Also failing in the visual test runner, but for a different reason
> 
> Screenshot 2022-01-27 at 10 20 54

Nevermind... Forgot to set the PL_E2E_TESTS variable 🤦‍♂️

MaKleSoft:
> Fails for me locally. Any idea why this could be? Screenshot 2022-01-27 at 09 51 29

Again, just a problem with my local environment configuration. Passing now!

MaKleSoft (approved):
Very nice! Watching the test runner walk automatically walk through the signup flow is **so** cool 😊

I still want to find a better solution for testing the email verification, but that can wait for a later time. I'm thinking something like [maildev](https://github.com/maildev/maildev) shouldn't be too hard to set up and would allow us to test all the components as-is without mocking anything or making any changes to the actual functionality.

MaKleSoft:
Btw, seeing the visual test runner gave me an idea: Would it be possible to use cypress to automatically **record** those test runs and upload them somewhere? That could be a really cool and efficient way to generate videos for the user manual.

.gitignore
```diff
@@ -24,6 +24,7 @@ packages/server/attachments
 /logs
 /pwa
 /data
+/test-data
```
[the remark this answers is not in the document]
BrunoBernardino: Yep, we don't!

package.json
```diff
@@ -45,6 +48,8 @@
         "tauri:build:production": "lerna run build:production",
         "repl": "cd packages/server && npm run repl && cd ../..",
         "test": "lerna run test",
+        "test:e2e": "rm -fr ./packages/server/test-data && concurrently --prefix=name --prefix-length=30 --kill-others --success=first -n app,cypress \"PL_DATA_BACKEND=memory PL_E2E_TESTS=true PL_DISABLE_SW=true npm start\" \"./node_modules/.bin/wait-on tcp:localhost:8080 && CYPRESS_CRASH_REPORTS=0 cypress run\"",
```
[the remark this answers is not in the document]
BrunoBernardino: You're right.

BrunoBernardino:
@MaKleSoft Cypress can do that (I like to disable all of it, though), but I think it'll run too fast to be helpful, and the videos will include the "test" on the left, which might be confusing.

PR #394, opened 2022-01-31
Refine Signup Flow

MaKleSoft:
This adds a couple of improvements to the signup flow, as well as some other nifty stuff:

- Added invite "preview" to login/signup/unlock screens to avoid confusion when arriving from an invite link
- Added additional signup step for accepting TOS and entering a name
- Added final success message. Yes, there is confetti. I regret nothing. 🥳

Screenshot 2022-01-31 at 08 14 35

Screenshot 2022-01-31 at 08 05 10


MaKleSoft:
@BrunoBernardino Do you have any idea why the e2e test may be failing here but passing for me locally? I've tried increasing the timeout but that didn't work. Maybe you can run the test locally on your machine to see if it passes there?

BrunoBernardino:
@MaKleSoft I can try to look into this later today, but a timeout longer than 200ms usually means trouble. I'm re-running the test just in case.

BrunoBernardino:
@MaKleSoft all passed now. I couldn't get the problem locally, but looking at where we had animations for elements that exist and simply change, I added a small delay, to make sure the test/runner doesn't try to click before it's finished. Since the element is available already, it's not possible to "wait until it's ready".

packages/core/src/server.ts
```diff
@@ -1112,13 +1112,26 @@ export class Controller extends API {
...
-                    let params = new URLSearchParams();
-                    params.set("email", invite.email);
-
                     const auth = await this._getAuth(invite.email);
                     auth.invites.push({ id: invite.id, orgId: org.id, orgName: org.name });
 
+                    let path = "";
+                    const params = new URLSearchParams();
+                    params.set("email", invite.email);
+                    // params.set("accountStatus", auth.accountStatus);
```
BrunoBernardino, line 1121: Is this meant to be commented out?

BrunoBernardino (approved):
Code changes look good, this worked for me locally, and I fixed the signup cypress test. I only have one question.

BrunoBernardino:
I'm merging because I want to finish the login tests, and it's best to start from this point.

MaKleSoft:
> I'm merging because I want to finish the login tests, and it's best to start from this point.

Yep, good call. And thanks for fixing the test!

PR #395, opened 2022-02-01
End-to-end login test

BrunoBernardino:
This finishes the e2e login test, and adds lock/unlock testing.

It also fixes an error the locking test caught (for accounts with no items), and caching of node_modules when testing, for faster results.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -10,6 +10,23 @@ jobs:
             - uses: actions/setup-node@v2
               with:
                   node-version-file: ".nvmrc"
+            - uses: actions/cache@v2
+              with:
+                  path: |
+                      ~/.npm
+                      ~/.nvm
+                      node_modules
+                      packages/app/node_modules
+                      packages/cordova/node_modules
+                      packages/core/node_modules
+                      packages/electron/node_modules
+                      packages/extension/node_modules
+                      packages/locale/node_modules
+                      packages/manage/node_modules
+                      packages/pwa/node_modules
+                      packages/server/node_modules
+                      packages/tauri/node_modules
+                  key: ${{ runner.os }}-${{ hashFiles('**/package-lock.json') }}
             - name: Install dependencies
               run: |
                   npm i -g npm@8.2.0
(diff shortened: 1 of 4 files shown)


packages/core/src/app.ts
```diff
@@ -477,7 +477,7 @@ export class App {
      * Locks the app and wipes all sensitive information from memory.
      */
     async lock() {
-        [this.account!, ...this.state.orgs, ...this.state.vaults].forEach((each) => each.lock());
+        [this.account!, ...this.state.orgs, ...this.state.vaults].forEach((each) => each && each.lock());
```
BrunoBernardino, line 480: This was the error it found. Basically some `each` was undefined or null, can't remember now, when locking an account without any items.

.github/workflows/run-tests.yml
```diff
@@ -10,6 +10,23 @@ jobs:
             - uses: actions/setup-node@v2
               with:
                   node-version-file: ".nvmrc"
+            - uses: actions/cache@v2
```
BrunoBernardino, line 13: This should shave up to a minute while running tests when the dependencies don't change.

packages/core/src/app.ts
MaKleSoft: Good catch! Btw, you can write this just a tiny bit more concisely using the optional chaining syntax:
```ts
each?.lock()
```

cypress/integration/01 - signup-login.ts
```diff
@@ -3,15 +3,15 @@ describe("Signup/Login", () => {
...
+        cy.login();
 
-    //     cy.lock();
+        cy.lock();
 
-    //     cy.unlock();
-    // });
+        cy.unlock();
+    });
 });
```
MaKleSoft, line 17: Just for my understanding: These tests each depend on a specific state left by the previous "command" and need to run in sequence, correct? Like the login test requires the signup test to complete first so there is an existing account, right?

MaKleSoft (approved):
Looks good! 👍

packages/core/src/app.ts
BrunoBernardino: Good suggestion!

PR #396, opened 2022-02-02
Create Item end-to-end test

BrunoBernardino:
This adds a test for creating an item.

It also caught another problem with a timeout that wasn't safety-checked, and I refactored a couple of items.

### DIFF
--- a/cypress/integration/02 - items.ts
+++ b/cypress/integration/02 - items.ts
@@ -1,0 +1,68 @@
+const testItem = {
+    name: "Google",
+    username: "example@google.com",
+    password: "somethingsecret",
+    url: "https://google.com",
+};
+
+describe("Items", () => {
+    it("can create an item without errors", () => {
+        cy.login();
+
+        // Click plus sign
+        cy.get("pl-app").find("pl-items").find("pl-items-list").find("pl-button:eq(2)").click({ force: true });
+
+        // Give the app some time to render the animations
+        cy.wait(100);
+
+        // Click create
+        cy.get("pl-app").find("pl-create-item-dialog").find("footer pl-button.primary").click({ force: true });
+
+        cy.url().should("include", "/items/");
+        cy.url().should("include", "/new");
+
+        // Give the app some (more) time to render the animations
+        cy.wait(100);
+
+        // Fill in form
+        cy.get("pl-app")
+            .find("pl-items")
+            .find("pl-item-view")
+            .find("pl-input#nameInput")
+            .find("input")
+            .type(testItem.name, { force: true });
+        cy.get("pl-app")
+            .find("pl-items")
+            .find("pl-item-view")
+            .find("pl-scroller")
+            .find("pl-list")
+            .find("pl-field:eq(0)")
+            .find("pl-input.value-input")
+            .find("input.input-element")
+            .type(testItem.username, { force: true });
+        cy.get("pl-app")
+            .find("pl-items")
+            .find("pl-item-view")
+            .find("pl-scroller")
+            .find("pl-list")
+            .find("pl-field:eq(1)")
+            .find("pl-input.value-input")
+            .find("input.input-element")
+            .type(testItem.password, { force: true });
+        cy.get("pl-app")
+            .find("pl-items")
+            .find("pl-item-view")
+            .find("pl-scroller")
+            .find("pl-list")
+            .find("pl-field:eq(2)")
+            .find("pl-input.value-input")
+            .find("input.input-element")
+            .type(testItem.url, { force: true });
+
+        // Click save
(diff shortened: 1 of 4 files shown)


cypress/integration/02 - items.ts
```diff
@@ -0,0 +1,68 @@
...
+        cy.wait(100);
+
+        // Click create
+        cy.get("pl-app").find("pl-create-item-dialog").find("footer pl-button.primary").click({ force: true });
+
+        cy.url().should("include", "/items/");
+        cy.url().should("include", "/new");
+
+        // Give the app some (more) time to render the animations
+        cy.wait(100);
```
BrunoBernardino, line 25: This _might_ need to be increased a bit more (this animation is slower), if, once we add the "find item", we can't consistently get the item properly typed.

packages/app/src/elements/field.ts
```diff
@@ -152,7 +152,7 @@ export class FieldElement extends LitElement {
 
     private _collapseSuggestions() {
         window.setTimeout(() => {
-            const drawer = this._valueInput.querySelector("pl-drawer") as Drawer;
+            const drawer = this._valueInput?.querySelector("pl-drawer") as Drawer;
```
BrunoBernardino, line 155: This was failing because the test ran too fast, and the timeout ran without a `this._valueInput` available.

packages/core/src/app.ts
```diff
@@ -477,7 +477,7 @@ export class App {
      * Locks the app and wipes all sensitive information from memory.
      */
     async lock() {
-        [this.account!, ...this.state.orgs, ...this.state.vaults].forEach((each) => each && each.lock());
+        [this.account!, ...this.state.orgs, ...this.state.vaults].forEach((each) => each?.lock());
```
BrunoBernardino, line 480: Refactored.

packages/app/src/elements/app.ts
```diff
@@ -547,7 +547,7 @@ export class App extends ServiceWorker(StateMixin(AutoSync(ErrorHandling(AutoLoc
         target.classList.add("dragging");
         target.addEventListener("dragend", () => target.classList.remove("dragging"), { once: true });
         const totp: TOTPElement | null = (target.querySelector("pl-totp") ||
-            (target.shadowRoot && target.shadowRoot.querySelector("pl-totp"))) as TOTPElement | null;
+            target.shadowRoot?.querySelector("pl-totp")) as TOTPElement | null;
```
BrunoBernardino, line 550: Refactored just because I saw it.

MaKleSoft: Good catch! I learned about the optional chaining syntax not so long ago so there are still quite a few places with the "old" pattern.

cypress/integration/02 - items.ts
```diff
@@ -0,0 +1,68 @@
...
+            .find("pl-items")
+            .find("pl-item-view")
+            .find("pl-input#nameInput")
+            .find("input")
+            .type(testItem.name, { force: true });
+        cy.get("pl-app")
+            .find("pl-items")
+            .find("pl-item-view")
+            .find("pl-scroller")
+            .find("pl-list")
```
MaKleSoft, line 38: I assume we need this long explicit chain because of shadow dom? If so, you can probably omit `pl-scroller` and `pl-list` here because their content is actually rendered into the "light dom" so should be able to query past their boundaries.

MaKleSoft (approved):
Looks good for a basic test. My only question is whether it makes sense to add a couple more assertions at the end? Like did it save the right name, are all fields there etc. Of course we can always add more granular checks later but I'm interested in how detailed you think checks should be in general for e2e tests.

cypress/integration/02 - items.ts
```diff
@@ -0,0 +1,68 @@
...
+            .find("pl-field:eq(2)")
+            .find("pl-input.value-input")
+            .find("input.input-element")
+            .type(testItem.url, { force: true });
+
+        // Click save
+        cy.get("pl-app").find("pl-items").find("pl-item-view").find("pl-button.primary").click({ force: true });
+
+        cy.url().should("include", "/items/");
+        cy.url().should("not.include", "/new");
```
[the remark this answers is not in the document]
BrunoBernardino: Right here I was just testing we created a basic item. We can certainly add those checks, but my plan is to use other tests to validate such assumptions. For example, once we add the test for searching/finding an item, we'll be able to verify we have the right item, which will test those assumptions "for free" (instead of testing them twice), if that makes sense.

cypress/integration/02 - items.ts
BrunoBernardino: Yes, it needs to be here because of the shadow dom. I can try removing these and seeing if that works, if you want, but I honestly think this makes reading (and potentially updating) the test easier.

PR #399, opened 2022-02-09
Add v3 compatibility end-to-end test

BrunoBernardino:
This is still a WIP. It adds the basic skeleton for the tests and is going through the signup, but currently unable to finish.

The tweaks I've made for the signup to work up until where it is might also not be correct.

The current error I get is `No srp session with the given id found!`, because the `srpId` sent is empty, even though the `auth` object does have `srpSessions`.


cypress/integration/03 - v3-compatibility.ts
```diff
@@ -0,0 +1,158 @@
...
+
+describe("v3 compatibility", () => {
+    it("can signup without errors", () => {
+        // NOTE: Temporary, until Billing is implemented, because that throws an error
+        cy.on("uncaught:exception", (error) => {
+            // @ts-ignore this exists
+            if (error.code === "invalid_request") {
+                return false;
+            }
+        });
```
BrunoBernardino, line 21: This can be removed once you get billing implemented, I believe.

cypress/support/commands.ts
```diff
@@ -246,3 +246,190 @@ Cypress.Commands.add("unlock", () => {
...
+
+    // Done!
+    cy.get("pl-app")
+        .find("pl-start")
+        .find("pl-login-signup")
+        .find("pl-drawer:eq(7)")
+        .find("pl-button")
+        .click({ force: true });
+
+    cy.url().should("include", "/items");
```
BrunoBernardino, line 360: This part is still "untested" (and it won't work since there's no `pl-login-signup`) because I can't get past the `srp` error.

packages/core/src/api.ts
```diff
@@ -139,7 +139,14 @@ export class RetrieveMFATokenResponse extends Serializable {
 export class StartRegisterAuthenticatorParams extends Serializable {
     type: AuthType = AuthType.Email;
 
-    purposes: AuthPurpose[] = [AuthPurpose.Signup, AuthPurpose.Login, AuthPurpose.Recover];
+    purposes: AuthPurpose[] = [
```
BrunoBernardino, line 142: I'm not entirely sure if this is the right approach, but the client sent `0` as the purpose, which I mapped to the `MFAPurpose` indexes in the v3 server.

packages/core/src/auth.ts
```diff
@@ -16,6 +16,9 @@ export enum AuthPurpose {
     GetLegacyData = "get_legacy_data",
     AccessKeyStore = "access_key_store",
     TestAuthenticator = "test_authenticator",
+    v3_Signup = 0,
+    v3_Login = 1,
+    v3_Recover = 2,
```
BrunoBernardino, line 21: I defined them here. I didn't define the `GetLegacyData` as well, but I believe I should, right?

packages/core/src/server.ts
```diff
@@ -1872,7 +1890,10 @@ export class Controller extends API {
                 (typeof requestId === "undefined" || r.id === requestId) &&
                 r.token === token &&
                 r.status === AuthRequestStatus.Verified &&
-                r.purpose === purpose
+                (r.purpose === purpose ||
+                    (purpose === AuthPurpose.Signup && r.purpose === AuthPurpose.v3_Signup) ||
+                    (purpose === AuthPurpose.Login && r.purpose === AuthPurpose.v3_Login) ||
+                    (purpose === AuthPurpose.Recover && r.purpose === AuthPurpose.v3_Recover))
```
BrunoBernardino, line 1896: This was necessary because the `authRequest` can come with the original `purpose`, but the `_useAuthToken` will come with the "new" purpose, so I wanted to make sure those would still match.

BrunoBernardino:
@MaKleSoft can you please review my PR comments/questions above and point me in the right direction for the srp thing? If not, it's fine, I'll just keep reverse-engineering and patching/pushing through.

MaKleSoft:
I took the liberty to fix the v3 compatibility issues myself. The differences in the endpoints are subtle but significant and it would have been quite a challenge for you to reverse-engineer the whole thing. Tests are passing now (at least locally) 🥳

packages/core/src/api.ts
```diff
@@ -29,7 +29,13 @@ export class CreateAccountParams extends Serializable {
...
 
     /**
-     * The verification token obtained from [[API.completeAuthRequest]].
+     * The authentication token obtained from [[API.completeAuthRequest]].
+     */
+    authToken: string = "";
+
+    /**
+     * The authentication token obtained from [[API.completeAuthRequest]].
+     * @deprecated Use `authToken` property instead
```
MaKleSoft, line 38: Been wanting to rename this for a while.

packages/core/src/server.ts
```diff
@@ -1913,7 +1832,7 @@ export class Server {
     private _requestQueue = new Map<AccountID | OrgID, Promise<void>>();
 
     makeController(ctx: Context) {
-        return new Controller(this, ctx);
+        return new (V3Compat(Controller))(this, ctx);
```
MaKleSoft, line 1835: By using a mixin allows us to neatly separate the "compatibility layer" and the regular api. Once all v3 clients have disappeared from the wild, all we have to do is undo this change and delete the `v3-compat` module.

packages/core/src/v3-compat.ts
```diff
@@ -0,0 +1,318 @@
...
+                }
+
+                if (request.tries >= 3) {
+                    throw new Err(
+                        ErrorCode.AUTHENTICATION_TRIES_EXCEEDED,
+                        "You have exceed your allowed numer of tries!"
+                    );
+                }
+
+                const authenticators = await this._getAuthenticators(auth);
```
MaKleSoft, line 214: This is the step you were missing when trying to get the `retrieveMFAToken` endpoint to work. Some authenticators don't need to be registered by the users but are added "ad hoc". By default the email authenticator is one of them. The catch is that you have to use `_getAuthenticators` to have these "ad hoc" authenticators included.

packages/core/src/v3-compat.ts
```diff
@@ -0,0 +1,318 @@
...
+         * Initiate the login procedure for a given account by requesting the authentication params
+         * which are required for proceeding with [[createSession]].
+         * @deprecated
+         */
+        @Handler(InitAuthParams, InitAuthResponse)
+        async initAuth({ email, verify }: InitAuthParams): Promise<InitAuthResponse> {
+            const { accountId, srpId, keyParams, B } = await this.startCreateSession(
+                new StartCreateSessionParams({ email, authToken: verify })
+            );
+            srpIds.set(accountId, srpId);
```
MaKleSoft, line 288: This is kind of an ugly workaround, but it'll do. The problem is that the v4 api requires the client to hold on to the id of the srp session, which is required by the `completeCreateSession` endpoint. What we're doing here is have the compatibility layer remember the srp id for the client by mapping it to the account id. This has the downside that only one session can be created at a time using this endpoint, but in practice that shouldn't be a problem.

BrunoBernardino:
@MaKleSoft thank you so much for your work on the server part. I got these working again in CI. The main issue was that we were trying to signup the same user twice (didn't happen when we _just_ ran the v3 tests, only with the whole suite). I decided to create another separate user (because the existing one was created with the v4 client, not v3). I also tweaked some "wait"s that sometimes failed.

I'm happy to merge this in for the v3 signup test and go at the rest of them in separate PRs, unless you want them all in a single PR.

MaKleSoft (approved):
Good to merge 👍

PR #401, opened 2022-02-14
v3 client compatibility end-to-end tests

BrunoBernardino:
This adds the final tests to login/lock/unlock, create item, and find item, with the v4 server and v3 client.


MaKleSoft (approved):
👍

PR #404, opened 2022-02-18
Use maildev for catching and parsing verification emails in e2e tests

MaKleSoft:
I figured I'd give this a shot since it seemed like a straightforward way of properly testing email authentication without the need of hardcoding a fake verification code. Not sure if my approach fetching and parsing the code in cypress is the correct one though, since I don't fully understand how chaining and assertions work in cypress. @BrunoBernardino Could you take a look?


cypress/support/commands.ts
```diff
@@ -28,14 +28,21 @@ Cypress.Commands.add("signup", () => {
...
+        expect(latest).to.not.be.undefined;
+        const matchCode = latest.text.match(/(\d{6})/);
+        const emailToken = matchCode && matchCode[1];
+        expect(emailToken).to.not.be.null;
+        cy.get("pl-app")
+            .find("pl-prompt-dialog")
+            .find("pl-input")
+            .find("input[placeholder='Enter Verification Code']")
+            .type(emailToken, { force: true });
+    });
```
MaKleSoft, line 45: The approach is pretty straighforward: Grab the latest email from the [maildev rest endpoint](https://github.com/maildev/maildev/blob/master/docs/rest.md) and extract the verification code. Not sure if there is a better way to do this with cypress though.

cypress/support/commands.ts
```diff
@@ -281,15 +295,21 @@ Cypress.Commands.add("v3_signup", () => {
...
+        const latest = res.body.sort((a, b) => (a.time > b.time ? -1 : 1))[0];
+        expect(latest).to.not.be.undefined;
+        const matchCode = latest.text.match(/(\d{6})/);
+        const emailToken = matchCode && matchCode[1];
+        expect(emailToken).to.not.be.null;
+        cy.get("pl-app")
+            .find("pl-prompt-dialog")
+            .find("pl-input")
+            .find("input[placeholder='Enter Verification Code']")
+            .type(emailToken, { force: true });
```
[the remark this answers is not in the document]
MaKleSoft: Yep, blindly copy and pasted it from the v4 test 🤦‍♂️

MaKleSoft:
Ok, so the flakyness of the tests was driving me crazy, and adding `cy.wait` all over the place just didn't seem like a good solution. So I ended up going down the Cypress rabbit hole to see if there was any way to improve the reliability of the tests. Here is what I found:

- Apparently one of the important concepts of Cypress is [retry-ability](https://docs.cypress.io/guides/core-concepts/retry-ability#Use-aliases). The `cy.get` command will retry finding an element until it has either found it and all assertions on it have passed or a timeout is reached. This means that even if the app takes a few moments to render an element, the test will still succeed as long as things don't take too long.
- While the `cy.get` command has this retry-ability built in, `cy.find` does not. This is a problem because web components will often take a moment to render their shadow dom and other content even after the element itself has been added to the dom.
- We can work around this problem by using [`cy.within`](https://docs.cypress.io/api/commands/within), which allows us to use `cy.get` all the way down. This gives us the retry-ability back that we lose when using `cy.find`.
- So
    ```ts
    cy.get("pl-app").find("pl-start").find("pl-login-signup").find("pl-button").click()
    ```
    becomes
    ```ts
    cy.get("pl-app").within(() => cy.get("pl-start").within(() => cy.get("pl-login-signup").within(() => cy.get("pl- 
   button").click())))
    ```
    Obviously this is horrible, so I wrote a helper command `cy.doWithin` that allows us to to write this more cleanly:
    ```ts
    cy.doWithin(["pl-app", "pl-start", "pl-login-signup"], () => {
        cy.get("pl-button").click();
        // do other stuff within pl-login-signup
    });
    ```

MaKleSoft:
Now, after going through all that trouble, I noticed that test still randomly failed in some places... So I spent some time debugging and found that in most cases, the root cause was neither delayed rendering nor animations, but the fact that the app is sometimes focusing and/or clearing input elements **while the `cy.type` command is being executed**. Here is an example: https://github.com/padloc/padloc/blob/v4/packages/app/src/elements/prompt-dialog.ts#L158. So a lot of times we were simply losing the first few letters when filling in stuff, which caused problems for obvious reasons.

Unfortunately this means that we still need delays in some strategic places, but at least now we know why...

MaKleSoft:
In any case, I think the `doWithin` command (and it's sibling `typeWithin`) still make things a bit more robust and generally more readable, so I wouldn't say that part was still worthwhile 😊

cypress/integration/01 - signup-login.ts
```diff
@@ -1,17 +1,19 @@
 describe("Signup/Login", () => {
+    const email = `${Math.floor(Math.random() * 1e8)}@example.com`;
```
MaKleSoft, line 2: While debugging the tests I noticed that I couldn't easily replay the signup test once it had passed once, since the second time the account already existed. This made debugging really annoying because I had to rerun the `test:e2e:dev` command every time instead of just clicking the replay button. I thought about ways to reset the server state before every test run, but ultimately decided it was easier to simply use a randomised email.

cypress/integration/02 - items.ts
```diff
@@ -10,125 +10,78 @@ const itemSearch = {
...
+            cy.get("pl-button:eq(3)").click();
+
+            // Find non-existent Item
+            cy.typeWithin("pl-input#filterInput", itemSearch.nonexistent, { force: true });
+
+            // Confirm we find none
+            cy.get("main pl-virtual-list pl-scroller div.content").children("div").should("have.length", 0);
+
+            cy.get("main > div.centering").should("contain.text", "did not match any items");
+        });
```
MaKleSoft, line 85: I think this is a great example of how the `doWithin` command makes things more readable.

cypress/integration/03 - v3-compatibility.ts
```diff
@@ -10,70 +10,62 @@ const v3_itemSearch = {
     nonexistent: "apple",
 };
 
+const v3_email = `${Math.floor(Math.random() * 1e8)}@example.com`;
```
MaKleSoft, line 13: Again, randomised email for replay-ability.

cypress/support/commands.ts
```diff
@@ -11,353 +11,277 @@ Cypress.Commands.add("clearIndexedDb", () => {
...
+    };
+
+    const resolveValue = () => {
+        getCode().then((value) => {
+            // @ts-ignore
+            return cy.verifyUpcomingAssertions(value, options, {
+                onRetry: resolveValue,
+            });
+        });
+    };
```
MaKleSoft, line 37: Had to use some Cypress-foo to add retry-ability here, since I didn't want to wait an arbitrary amount of time for the email to arrive. Instead, we're just retrying the maildev endpoint until we get a match (or the timeout elapses). Unfortunately adding retry-ability to custom commands is not well documented so I had to piece this together from built-in commands and answers on Github. To be honest I don't fully understand what's happening here, but it works 😊.

cypress/support/commands.ts
```diff
@@ -11,353 +11,277 @@ Cypress.Commands.add("clearIndexedDb", () => {
...
+                return null;
+            }
+            const matchCode = latest.text.match(/(\d{6})/);
+            return matchCode && matchCode[1];
+        });
+    };
+
+    const resolveValue = () => {
+        getCode().then((value) => {
+            // @ts-ignore
```
MaKleSoft, line 32: Apparently the `cy.verifyUpcomingAssertions` command is missing from the types, hence the `@ts-ignore`

package.json
```diff
@@ -20,6 +20,7 @@
         "cypress": "9.3.1",
         "http-server": "14.1.0",
         "lerna": "4.0.0",
+        "maildev": "github:padloc/maildev",
```
MaKleSoft, line 23: Switched to a custom for of maildev since the original had some dependencies with high and critical vulnerabilities. See https://github.com/padloc/maildev/commit/a3ac6bf682dfb7cbb08b8b86e7b8afc9e5e41c66

cypress/integration/01 - signup-login.ts
BrunoBernardino: Nice one. I didn't do this initially because of the "hardocded email" need.

cypress/support/commands.ts
BrunoBernardino: Nice one!

BrunoBernardino (approved):
I wish you would've talked to me about this before (the `get` vs `find` and retry-ability, with the shadow dom), as it could've saved you some time, but this way you also probably got a deeper knowledge of cypress and how it works.

I like the `doWithin` especially because it doesn't greatly reduce readability and appears to improve test maintainability (which is usually what's lost with highly customized and magical test commands -- others can't see how `doWithin` works from Cypress' documentation).

You've probably guessed by now that the fact the shadow dom is "ready" to cypress before it's ready to receive inputs is the main reason for many of the reliability problems. The other major piece is the fact that many elements need to be clicked before they're available (the inputs are disabled before the parent element is clicked, with the label), which then introduces the possibility of typing being done before the input is ready, like you mentioned.

Fixing these is possible, just not trivial or quick, and the main goal was to get the tests working and doing their job, even if sub-optimally. I'm happy to see you making improvements so soon!

In sum, great job!

MaKleSoft:
> I wish you would've talked to me about this before (the get vs find and retry-ability, with the shadow dom), as it could've saved you some time, but this way you also probably got a deeper knowledge of cypress and how it works.

Yeah well, I wasn't planning on diving in this deep but it seemed like an interesting problem so I kinda just got sucked in as I often do 😅. I did learn a lot about Cypress in the process so I wouldn't worry about wasted time 😁

PR #405, opened 2022-02-23
Enable Auto-lock in web extension

### LINKED ISSUE #300: Auto-Lock does not work in Chrome Extension

BrunoBernardino:
This allows the web extension to auto-lock (tested in Chrome and Firefox), by running an alarm in the background, since the `setTimeout` is ignored past 10 seconds.

I noticed that if the popup is open, it does not lock until it’s closed (even after firing the alarm). I couldn’t find a way to change the route for the app from the background, and I wasn’t sure if it would worth it adding a new listener for this, as it’s not common for the extension to be open for a long period of time anyway.

Closes #300

### DIFF
--- a/packages/extension/src/background.ts
+++ b/packages/extension/src/background.ts
@@ -12,9 +12,9 @@ setPlatform(new ExtensionPlatform());
 class ExtensionBackground {
     app = new App(new AjaxSender(process.env.PL_SERVER_URL!));
 
-    _autoLockAlarmName = "pl_autoLock";
+    private _autoLockAlarmName = "pl_autoLock";
 
-    get _lockDelayInMinutes() {
+    private get _lockDelayInMinutes() {
         return this.app.settings.autoLockDelay;
     }
 
@@ -220,7 +220,7 @@ class ExtensionBackground {
         }
     }
 
-    async _cancelAutoLock() {
+    private async _cancelAutoLock() {
         await browser.alarms.clear(this._autoLockAlarmName);
     }
 
--- a/packages/extension/src/background.ts
+++ b/packages/extension/src/background.ts
@@ -12,6 +12,12 @@ setPlatform(new ExtensionPlatform());
 class ExtensionBackground {
     app = new App(new AjaxSender(process.env.PL_SERVER_URL!));
 
+    _autoLockAlarmName = "pl_autoLock";
+
+    get _lockDelayInMinutes() {
+        return this.app.settings.autoLockDelay;
+    }
+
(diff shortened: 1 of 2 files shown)


packages/extension/src/background.ts
```diff
@@ -12,6 +12,12 @@ setPlatform(new ExtensionPlatform());
 class ExtensionBackground {
     app = new App(new AjaxSender(process.env.PL_SERVER_URL!));
 
+    _autoLockAlarmName = "pl_autoLock";
+
+    get _lockDelayInMinutes() {
```
BrunoBernardino, line 17: A lot of code here is _similar_ (but not the same) as the `AutoLock` class, so I initially tried to make that work. In the end I was still duplicating most code, and since these alarms are very specific to the `background` piece of the `extension`, I ended up leaving them here.

packages/extension/src/background.ts
```diff
@@ -33,11 +39,13 @@ class ExtensionBackground {
                 case "loggedOut":
                 case "locked":
                     await this.app.load();
+                    this._cancelAutoLock();
```
BrunoBernardino, line 42: This should be unnecessary, but still, didn't seem to hurt.

BrunoBernardino:
Beta artifacts are being built at https://github.com/padloc/padloc/actions/runs/1888552953

packages/extension/src/background.ts
```diff
@@ -12,6 +12,12 @@ setPlatform(new ExtensionPlatform());
 class ExtensionBackground {
     app = new App(new AjaxSender(process.env.PL_SERVER_URL!));
 
+    _autoLockAlarmName = "pl_autoLock";
```
MaKleSoft, line 15: Minor nitpick: This should be `private`.

packages/extension/src/background.ts
MaKleSoft, line 17: Also missing `private` .

packages/extension/src/background.ts
```diff
@@ -205,6 +219,30 @@ class ExtensionBackground {
             browser.browserAction.setTitle({ title: process.env.PL_APP_NAME || "" });
         }
     }
+
+    async _cancelAutoLock() {
```
MaKleSoft, line 223: Again, `private` 😉

MaKleSoft (changes requested):
Some properties/methods/getters should be private. Looks good otherwise!

MaKleSoft (approved):

Issue #386, opened 2022-01-15
[Tauri] Can't upload attachments


MaKleSoft:

MaKleSoft:
Looks like a problem in tauri-apps/wry: https://github.com/tauri-apps/wry/issues/305

BrunoBernardino:
Looks like the Tauri issue has been fixed. I'll take a look at this again.

PR #407, opened 2022-02-25
Allow drag & drop for attachments

### LINKED ISSUE #379: Add attachments via Drag & Drop
https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/File_drag_and_drop

BrunoBernardino:
Enables adding an attachment to an item via drag & drop. An item needs to be selected/visible.

Works for electron, but not for Tauri (#386).

Fixes #379


packages/app/src/elements/item-view.ts
```diff
@@ -358,7 +371,13 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
                 </header>
 
                 <pl-scroller class="stretch">
-                    <div class="vertical layout fill-vertically content">
+                    <div
```
BrunoBernardino, line 374: I decided to place this here and not support adding an attachment without a selected item via drag & drop, because it also requires adding a name, etc. It's possible, but the UX felt inferior, as if there's an item already selected.

packages/app/src/elements/item-view.ts
```diff
@@ -556,9 +575,7 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
         }
     }
 
-    private async _attachFile() {
-        const file = this._fileInput.files![0];
-        this._fileInput.value = "";
+    private async _addFileAttachment(file: File) {
```
BrunoBernardino, line 578: It looks weird here, but I made this piece reusable, to use when clicking/selecting and when drag+dropping.

packages/app/src/elements/item-view.ts
```diff
@@ -619,6 +642,44 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
         }
     }
 
+    private async handleDrop(event: DragEvent) {
+        event.preventDefault();
+
+        this._isDraggingFileToAttach = true;
```
BrunoBernardino, line 648: I considered adding a "uploading" view here, but:

1) We didn't have it for uploading via click/input.
2) It's barely visible since the dialog for the upload/file is on top.
3) There's already a notification when the upload finishes.

packages/app/src/elements/item-view.ts
```diff
@@ -619,6 +642,44 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
...
+    }
+
+    private async handleDragOver(event: DragEvent) {
+        event.preventDefault();
+        this._isDraggingFileToAttach = true;
+    }
+
+    private async handleDragLeave(event: DragEvent) {
+        event.preventDefault();
+        // Only cancel when it's leaving to a different element, since this fires while dragging inside the same element
```
BrunoBernardino, line 676: This was a tough thing to tweak. I'm not sure if it's because of lit-element or something else, but this fired while dragging, so it kept flickering between the "drop" screen and the item info, without this code.

package.json
```diff
@@ -33,6 +33,7 @@
         "pwa:start": "lerna run start --scope @padloc/pwa",
         "server:start": "lerna run start --scope @padloc/server --stream",
         "server:start-dry": "lerna run start-dry --stream --scope @padloc/server",
+        "electron:start": "lerna run start --scope @padloc/electron",
```
MaKleSoft, line 36: FYI, I changed this script's name to `dev` in my current PR.

MaKleSoft:
Love it! Regarding the UX, I agree that requiring an item to be created and selected first is the most straightforward approach. Dragging a file anywhere in the app to create a new Vault Item with that attachment would be cool, too, but that seems much more complicated to get right UX-wise so for now this is the better way.

I would even go a step further and add a dedicated drop zone, like this:

Screenshot 2022-02-25 at 17 31 56

This would make this feature (and attachments in general) more discoverable and the UX much more obvious. We could still have the actual drop zone to be the full item view and just highlight that area if a file drag is detected. What do you think?

BrunoBernardino:
That is a great suggestion. I'll need to make sure that only shows on PWA and not on mobile/extensions though, and nice catch on the translation, which I had forgotten.

I'll only tackle this tomorrow, so feel free to merge your PR and I'll resolve the conflicts here later.

MaKleSoft (changes requested):
I'm impressed how fast you got this to work! Like I said in my other comment, I think it would be even better having an always visible drop zone that can also be clicked rather than an overlay that's only visible if the user is already dragging a file. Also see by other comments.

MaKleSoft:
> I'll need to make sure that only shows on PWA and not on mobile/extensions though
@BrunoBernardino Why? Pretty sure it'll work there, too! Dragging might be harder on mobile, but with split view and drag&drop working between apps now, at least on iOS I don't see a reason to withhold this goodness from our mobile users 😁

packages/app/src/elements/item-view.ts
```diff
@@ -234,6 +237,11 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
                     padding-bottom: calc(var(--inset-bottom) + 0.5em);
                 }
             }
+
+            .dropzone {
+                border: dashed 2px var(--color-highlight);
+                border-radius: var(--scrollbar-width, 0.8em);
```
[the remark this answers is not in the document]
BrunoBernardino: I just copied that from the item-view, so the border would be visible, with the same rounding. :sweat_smile:

package.json
BrunoBernardino: I don't see it? It's still `start` after your merged PR, which makes sense since it's not really "watching", just building it once and showing locally.

package.json
```diff
@@ -47,11 +48,11 @@
         "dev": "lerna run --parallel --scope '@padloc/{server,pwa}' --parallel dev",
         "tauri:dev": "lerna run --parallel --scope '@padloc/{server,tauri}' --parallel dev",
         "tauri:update": "lerna run update",
-        "tauri:build": "lerna run build --scope @padloc/tauri",
-        "tauri:build:debug": "lerna run build:debug --scope @padloc/tauri",
+        "tauri:build": "lerna run build:production",
+        "tauri:build:debug": "lerna run build --scope @padloc/tauri",
```
BrunoBernardino, line 52: We missed changing this here.

package.json
```diff
@@ -47,11 +48,11 @@
...
         "repl": "cd packages/server && npm run repl && cd ../..",
         "test": "lerna run test",
-        "test:e2e": "concurrently --prefix=name --prefix-length=30 --kill-others --success=first -n app,v3-app,cypress \"PL_DATA_BACKEND=memory PL_DISABLE_SW=true PL_EMAIL_BACKEND=smtp PL_EMAIL_SMTP_HOST=localhost PL_EMAIL_SMTP_PORT=1025 PL_EMAIL_SMTP_IGNORE_TLS=true npm start\" \"npm run start:v3\" \"npx maildev\" \"./node_modules/.bin/wait-on tcp:localhost:8080 && CYPRESS_CRASH_REPORTS=0 cypress run\"",
+        "test:e2e": "concurrently --prefix=name --prefix-length=30 --kill-others --success=first -n app,v3-app,maildev,cypress \"PL_DATA_BACKEND=memory PL_DISABLE_SW=true PL_EMAIL_BACKEND=smtp PL_EMAIL_SMTP_HOST=localhost PL_EMAIL_SMTP_PORT=1025 PL_EMAIL_SMTP_IGNORE_TLS=true npm start\" \"npm run start:v3\" \"npx maildev\" \"./node_modules/.bin/wait-on tcp:localhost:8080 && CYPRESS_CRASH_REPORTS=0 cypress run\"",
```
BrunoBernardino, line 55: I hadn't realized we'd missed this name as well

BrunoBernardino:
@MaKleSoft ready for a final review.

package.json
MaKleSoft: Good catch!

MaKleSoft:
I just made two minor adjustments:

- Make sure attachments section is displayed even if no attachments have been added yet (otherwise the "drop files here" button wouldn't be visible which would kind of defeat the purpose).
- Remove the overlay and just use some subtle styling on the drop area / button to indicate "drop readiness" instead

Hope you don't mind!

Screenshot 2022-02-26 at 16 16 37 (2)

BrunoBernardino:
Sounds good! I generally prefer the overlay as it's a more obvious indicator of what's happening and where can you drop the item. I also didn't keep the section displayed because I figured it could be too noisy since most items won't have attachments anyway.

Since those are arguably personal preferences and can be easily changed, I don't mind your choice at all!

PR #408, opened 2022-02-25
Add support for .env file to server package; update readme files

MaKleSoft:
- @padloc/server now supports loading configuration from a `.env` file
- Updated/added some npm scripts to improve ergonomics
- Updated existing readme files and added new ones for some packages

There is still some work to be done on documenting the different configuration options for the server, but I'd rather do this a little later since I'd like to get this merged so that we can merge the `v4` branch soon.

### DIFF
--- a/.github/workflows/build-tauri.yml
+++ b/.github/workflows/build-tauri.yml
@@ -52,14 +52,14 @@ jobs:
                   npm ci
             - name: Build (debug)
               if: github.event.inputs.environment != 'Production'
-              run: npm run tauri:build
+              run: npm run tauri:build:debug
               env:
                   PL_SERVER_URL: ${{ secrets.PL_SERVER_URL }}
                   TAURI_PRIVATE_KEY: ${{ secrets.TAURI_PRIVATE_KEY }}
                   TAURI_KEY_PASSWORD: ${{ secrets.TAURI_KEY_PASSWORD }}
             - name: Build (production)
               if: github.event.inputs.environment == 'Production'
-              run: npm run tauri:build:production
+              run: npm run tauri:build
               env:
                   PL_SERVER_URL: ${{ secrets.PL_SERVER_URL }}
                   TAURI_PRIVATE_KEY: ${{ secrets.TAURI_PRIVATE_KEY }}
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -3,8 +3,6 @@
     "files.associations": {
         "*.svg": "html"
     },
-    "editor.formatOnSave": true,
-    "csscomb.formatOnSave": true,
     "prettier.prettierPath": "./node_modules/prettier",
     "files.insertFinalNewline": true
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -4,5 +4,7 @@
         "*.svg": "html"
     },
     "editor.formatOnSave": true,
-    "csscomb.formatOnSave": true
+    "csscomb.formatOnSave": true,
+    "prettier.prettierPath": "./node_modules/prettier",
+    "files.insertFinalNewline": true
 }
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -2,5 +2,7 @@
     "lit-plugin.globalAttributes": ["disabled"],
     "files.associations": {
         "*.svg": "html"
-    }
+    },
+    "editor.formatOnSave": true,
+    "csscomb.formatOnSave": true
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
-and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+and this project adheres to
+[Semantic Versioning](http://semver.org/spec/v2.0.0.html).
 
 ## 3.1.2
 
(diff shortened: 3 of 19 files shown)


.github/workflows/build-tauri.yml
```diff
@@ -52,14 +52,14 @@ jobs:
                   npm ci
             - name: Build (debug)
               if: github.event.inputs.environment != 'Production'
-              run: npm run tauri:build
+              run: npm run tauri:build:debug
```
MaKleSoft, line 55: I felt that the regular "production" build should be the default.

.github/workflows/build-tauri.yml
```diff
@@ -52,14 +52,14 @@ jobs:
...
               env:
                   PL_SERVER_URL: ${{ secrets.PL_SERVER_URL }}
                   TAURI_PRIVATE_KEY: ${{ secrets.TAURI_PRIVATE_KEY }}
                   TAURI_KEY_PASSWORD: ${{ secrets.TAURI_KEY_PASSWORD }}
             - name: Build (production)
               if: github.event.inputs.environment == 'Production'
-              run: npm run tauri:build:production
+              run: npm run tauri:build
```
MaKleSoft, line 62: See previous comment

.vscode/settings.json
```diff
@@ -2,5 +2,9 @@
     "lit-plugin.globalAttributes": ["disabled"],
     "files.associations": {
         "*.svg": "html"
-    }
+    },
+    "editor.formatOnSave": true,
+    "csscomb.formatOnSave": true,
+    "prettier.prettierPath": "./node_modules/prettier",
```
MaKleSoft, line 8: I noticed that the auto formatting in VSCode (using the Prettier plugin) sometimes produced different results than `npm run prettier`, which was because VSCode was using a different Prettier version installed elsewhere. This is to make sure we're using a consistent version for formatting.

CHANGELOG.md
```diff
@@ -3,7 +3,8 @@
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](http://keepachangelog.com/en/1.0.0/)
-and this project adheres to [Semantic Versioning](http://semver.org/spec/v2.0.0.html).
+and this project adheres to
+[Semantic Versioning](http://semver.org/spec/v2.0.0.html).
```
MaKleSoft, line 7: Automatically re-wrapped due to change in the `.prettierrc.json`.

packages/server/src/config.ts
```diff
@@ -34,9 +36,6 @@ export class EmailConfig extends Config {
 
     @ConfigParam(SMTPConfig)
     smtp?: SMTPConfig;
-
-    @ConfigParam("string")
-    footer?: string;
```
MaKleSoft, line 39: This wasn't used anymore.

packages/server/src/config.ts
```diff
@@ -153,5 +152,9 @@ export class PadlocConfig extends Config {
 }
 
 export function getConfig() {
+    const envFile = process.argv.find((arg) => arg.startsWith("--env="))?.slice(6);
+    const path = envFile && resolve(process.cwd(), envFile);
+    const override = process.argv.includes("--env-override");
+    dotenv.config({ override, path });
```
MaKleSoft, line 158: Load `.env` file if supplied.

MaKleSoft:
This PR ist mostly for getting the v4 branch "in shape" for merging into master. The documentation is still far from complete, but at least the most essential info is up-to-date now.

.github/workflows/build-tauri.yml
MaKleSoft: Can you explain what you mean by "local" in this context? By default, the build still points to localhost. My line of thinking here was "what would somebody expect the default behaviour to be if they just wanted to build their own version?" In that context, building for debugging purposes should be the exception I think.

.vscode/settings.json
BrunoBernardino: Nice one. I don't have it installed globally, so maybe that's why I didn't run into this.

.github/workflows/build-tauri.yml
BrunoBernardino: Sure. That's what I meant by "it's not like this is deploying directly to production". In any case, usually, developing locally, you want to get as much info for debugging errors as possible, but you're right that if the most common scenario is to run the build for a "personal production", this way makes more sense.

.vscode/settings.json
```diff
@@ -2,5 +2,9 @@
     "lit-plugin.globalAttributes": ["disabled"],
     "files.associations": {
         "*.svg": "html"
-    }
+    },
+    "editor.formatOnSave": true,
+    "csscomb.formatOnSave": true,
```
BrunoBernardino, line 7: I really don't think these kinds of settings should be versioned (aka forced), because we already have linting (prettier and tsc) happening in CI. It's more of a personal preference kind of thing, though, as any malfunction of the "auto-formatters" is _generally_ easy to revert.

BrunoBernardino (approved):
I didn't read through everything carefully, but glanced most of the README changes. It looks good and clear!

.vscode/settings.json
MaKleSoft: Yes, you're right. I'll remove it.

.github/workflows/build-tauri.yml
MaKleSoft: Yep, "personal production" is a good way of putting it 😊. Basically I want to make the whole build/deployment process as ergonomic as possible for this kind of use case since personal users will be less inclined to understand the nuances of the build process.

Issue #348, opened 2021-12-23
Advanced CSV Import


MaKleSoft:
Allow importing arbitrary CSV files by providing a UI for mapping columns, field types etc.

BrunoBernardino:
Fixed by #411

Issue #223, opened 2019-09-09
Make a better effort at guessing field types when importing


MaKleSoft:
Any data not exported from Padloc 3 directly will probably not have any field metadata associated with it. We can make an educated guess based on field name and value though, and a sporadic version of that is implemented here: https://github.com/padloc/padloc/blob/master/packages/core/src/item.ts#L221-L237

Correctly guessing field types like phone numbers, email addresses and urls during import would greatly improve use experience so we should probably put a little more effort into these heuristics.  These are the basic strategies I can think of.

- Match against common field names like ("username", "password", "url", etc.)
- Match against common field names in other languages
- Pattern match things that have a clear structure, like emails, urls, phone numbers, credit cards etc.

BrunoBernardino:
Fixed by #411

PR #411, opened 2022-03-02
Advanced CSV Import UI

BrunoBernardino:
This is only starting some logic, very bare bones and with some notes in the code.

Related to #348

### DIFF
--- a/packages/app/src/elements/import-dialog.ts
+++ b/packages/app/src/elements/import-dialog.ts
@@ -8,10 +8,9 @@ import { Select } from "./select";
 import { Input } from "./input";
 import { Dialog } from "./dialog";
 import "./button";
+import { ToggleButton } from "./toggle-button";
 import { customElement, query, state, queryAll } from "lit/decorators.js";
 import { html, css } from "lit";
-import { saveFile } from "@padloc/core/src/platform";
-import { stringToBytes } from "@padloc/core/src/encoding";
 
 const fieldTypeOptions = Object.keys(FIELD_DEFS).map((fieldType) => ({
     label: FIELD_DEFS[fieldType].name as string,
@@ -29,8 +28,8 @@ export class ImportDialog extends Dialog<File, void> {
     @state()
     private _itemColumns: imp.ImportCSVColumn[] = [];
 
-    @query("#csvHasDataOnFirstRowCheckbox")
-    private _csvHasDataOnFirstRowCheckbox: HTMLInputElement;
+    @query("#csvHasColumnsOnFirstRowButton")
+    private _csvHasColumnsOnFirstRowButton: ToggleButton;
 
     @query("#formatSelect")
     private _formatSelect: Select<string>;
@@ -59,31 +58,7 @@ export class ImportDialog extends Dialog<File, void> {
         `,
     ];
 
-    // TODO: Fix dialog reset after closed (when re-opened)
-
     renderContent() {
-        // TODO: Fix wrong first selected value for selects
-        if (this._formatSelect?.value === imp.CSV.value) {
-            if (
-                this._nameColumnSelect &&
-                (this._nameColumnSelect.selectedIndex === -1 || this._nameColumnSelect.selectedIndex === undefined)
-            ) {
-                const selectedNameIndex = this._itemColumns.findIndex((itemColumn) => itemColumn.type === "name");
-                this._nameColumnSelect.value = selectedNameIndex;
-                this._nameColumnSelect.selectedIndex = selectedNameIndex;
-            }
-
-            if (
-                this._tagsColumnSelect &&
-                (this._tagsColumnSelect.selectedIndex === -1 || this._tagsColumnSelect.selectedIndex === undefined)
-            ) {
-                // +1 because we have the "none" option first
-                const selectedTagsIndex = this._itemColumns.findIndex((itemColumn) => itemColumn.type === "tags") + 1;
-                this._tagsColumnSelect.value = selectedTagsIndex - 1;
-                this._tagsColumnSelect.selectedIndex = selectedTagsIndex;
-            }
-        }
-
         return html`
             <div class="padded vertical spacing layout fit-vertically">
                 <h1 class="big text-centering margined">${$l("Import Data")}</h1>
@@ -97,8 +72,7 @@ export class ImportDialog extends Dialog<File, void> {
                 ></pl-select>
 
                 <div class="small padded" ?hidden=${this._formatSelect && this._formatSelect.value !== imp.CSV.value}>
-                    ${$l("Choose the correct column names and types for each column below. If you are having trouble,")}
-                    <a href="#" @click=${this._downloadCSVSampleFile}> ${$l("Download the Sample File")} </a>.
+                    ${$l("Choose the correct column names and types for each column below.")}
                 </div>
 
                 <pl-scroller
@@ -106,12 +80,13 @@ export class ImportDialog extends Dialog<File, void> {
                     ?hidden=${this._formatSelect && this._formatSelect.value !== imp.CSV.value}
                 >
                     <ul class="vertical spacing layout">
-                        <div class="padded">
-                            <label>
-                                <input type="checkbox" id="csvHasDataOnFirstRowCheckbox" />
-                                ${$l("Check if the CSV file has data on the first row (instead of the column names)")}
-                            </label>
-                        </div>
+                        <pl-toggle-button
+                            class="transparent"
+                            id="csvHasColumnsOnFirstRowButton"
+                            .label=${$l("First row contains field names")}
+                            reverse
+                        >
+                        </pl-toggle-button>
 
                         <pl-select
                             id=${"nameColumnSelect"}
@@ -240,23 +215,41 @@ export class ImportDialog extends Dialog<File, void> {
         await this.updateComplete;
         const result = super.show();
 
+        // Reset fields
+        this._items = [];
+        this._itemColumns = [];
+        this._csvHasColumnsOnFirstRowButton.active = true;
         this._file = file;
-        this._formatSelect.value = ((await imp.guessFormat(file)) || imp.CSV).value;
+
+        const importFormat = (await imp.guessFormat(file)) || imp.CSV;
+        this._formatSelect.value = importFormat.value;
+
+        // Select first options properly
+        if (importFormat.value === imp.CSV.value) {
+            if (
+                this._nameColumnSelect &&
+                (this._nameColumnSelect.selectedIndex === -1 || this._nameColumnSelect.selectedIndex === undefined)
+            ) {
+                const selectedNameIndex = this._itemColumns.findIndex((itemColumn) => itemColumn.type === "name");
+                this._nameColumnSelect.value = selectedNameIndex;
+                this._nameColumnSelect.selectedIndex = selectedNameIndex;
+            }
+
+            if (
+                this._tagsColumnSelect &&
+                (this._tagsColumnSelect.selectedIndex === -1 || this._tagsColumnSelect.selectedIndex === undefined)
+            ) {
+                // +1 because we have the "none" option first
+                const selectedTagsIndex = this._itemColumns.findIndex((itemColumn) => itemColumn.type === "tags") + 1;
+                this._tagsColumnSelect.value = selectedTagsIndex - 1;
+                this._tagsColumnSelect.selectedIndex = selectedTagsIndex;
+            }
+        }
+
         await this._parseData();
         this._vaultSelect.value = app.mainVault!;
 
         return result;
-    }
-
-    private async _downloadCSVSampleFile(e: Event) {
-        e.preventDefault();
-        saveFile(
-            `${process.env.PL_APP_NAME}_csv_import_sample.csv`,
-            "text/csv",
-            stringToBytes(`name,tags,url,username,password,notes
-Facebook,social,https://facebook.com/,john.doe@gmail.com,3kjaf93,"Some note..."
-Github,"work,coding",https://github.com,john.doe@gmail.com,129lskdf93`)
-        );
     }
 
     private async _parseData(): Promise<void> {
@@ -288,7 +281,7 @@ Github,"work,coding",https://github.com,john.doe@gmail.com,129lskdf93`)
                 this._items = await imp.asLastPass(file);
                 break;
             case imp.CSV.value:
-                const result = await imp.asCSV(file, this._itemColumns, this._csvHasDataOnFirstRowCheckbox.checked);
+                const result = await imp.asCSV(file, this._itemColumns, this._csvHasColumnsOnFirstRowButton.active);
                 this._items = result.items;
                 this._itemColumns = result.itemColumns;
                 break;
--- a/packages/app/src/elements/import-dialog.ts
+++ b/packages/app/src/elements/import-dialog.ts
@@ -17,6 +17,11 @@ const fieldTypeOptions = Object.keys(FIELD_DEFS).map((fieldType) => ({
     value: fieldType,
 }));
 
+fieldTypeOptions.push({
+    label: "Ignore/skip this column",
+    value: "skip",
+});
+
 @customElement("pl-import-dialog")
 export class ImportDialog extends Dialog<File, void> {
     @state()
@@ -90,58 +95,17 @@ export class ImportDialog extends Dialog<File, void> {
                         <pl-select
                             id=${"nameColumnSelect"}
                             .label=${$l("Name Column")}
-                            .options=${this._itemColumns.map((itemColumn, itemColumnIndex) => ({
-                                label: csvHasColumnsOnFirstRow
-                                    ? `${itemColumn.displayName} (${$l(
-                                          "Column {0}",
-                                          (itemColumnIndex + 1).toString()
-                                      )})`
-                                    : $l("Column {0}", (itemColumnIndex + 1).toString()),
-                                value: itemColumnIndex,
-                            }))}
+                            .options=${this._nameColumnSelectOptions()}
                             .selectedIndex=${this._nameColumnSelect?.selectedIndex}
-                            @change=${() => {
-                                const currentNameColumnIndex = this._itemColumns.findIndex(
-                                    (itemColumn) => itemColumn.type === "name"
-                                );
-                                const nameColumnIndex = this._nameColumnSelect?.value || 0;
-                                this._itemColumns[nameColumnIndex].type = "name";
-
-                                if (currentNameColumnIndex !== -1) {
-                                    this._itemColumns[currentNameColumnIndex].type = FieldType.Text;
-                                }
-
-                                this._parseData();
-                            }}
+                            @change=${this._handleNameColumnChange}
                         ></pl-select>
 
                         <pl-select
                             id=${"tagsColumnSelect"}
                             .label=${$l("Tags Column")}
-                            .options=${[
-                                { label: $l("None"), value: -1 },
-                                ...this._itemColumns.map((itemColumn, itemColumnIndex) => ({
-                                    label: `${itemColumn.displayName} (${$l(
-                                        "Column {0}",
-                                        (itemColumnIndex + 1).toString()
-                                    )})`,
-                                    value: itemColumnIndex,
-                                })),
-                            ]}
+                            .options=${this._tagsColumnSelectOptions()}
                             .selectedIndex=${this._tagsColumnSelect?.selectedIndex}
-                            @change=${() => {
-                                const currentTagsColumnIndex = this._itemColumns.findIndex(
-                                    (itemColumn) => itemColumn.type === "tags"
-                                );
-                                const tagsColumnIndex = this._tagsColumnSelect?.value || 0;
-                                this._itemColumns[tagsColumnIndex].type = "tags";
-
-                                if (currentTagsColumnIndex !== -1) {
-                                    this._itemColumns[currentTagsColumnIndex].type = FieldType.Text;
-                                }
-
-                                this._parseData();
-                            }}
+                            @change=${this._handleTagsColumnChange}
                         ></pl-select>
 
                         <div class="spacer"></div>
@@ -175,14 +139,8 @@ export class ImportDialog extends Dialog<File, void> {
                                         .label=${$l("Field Name")}
                                         class="field-name-input"
                                         .value=${itemColumn.displayName}
-                                        @change=${() => {
-                                            const newFieldNameInput = this._fieldNameInputs[itemColumnIndex];
-                                            const newFieldName = newFieldNameInput?.value;
-                                            if (newFieldName) {
-                                                this._itemColumns[itemColumnIndex].displayName = newFieldName;
-                                                this._parseData();
-                                            }
-                                        }}
+                                        @change=${() => this._handleFieldNameChange(itemColumnIndex)}
+                                        ?hidden=${itemColumn.type === "skip"}
                                     ></pl-input>
 
                                     <pl-select
@@ -192,13 +150,7 @@ export class ImportDialog extends Dialog<File, void> {
                                         .label=${$l("Field Type")}
                                         .options=${fieldTypeOptions}
                                         .value=${itemColumn.type}
-                                        @change=${() => {
-                                            const newValue = this._fieldTypeSelects[itemColumnIndex]?.value;
-                                            if (newValue) {
-                                                this._itemColumns[itemColumnIndex].type = newValue;
-                                                this._parseData();
-                                            }
-                                        }}
+                                        @change=${() => this._handleFieldTypeChange(itemColumnIndex)}
                                     ></pl-select>
                                 </li>
                             `
@@ -243,6 +195,74 @@ export class ImportDialog extends Dialog<File, void> {
         this._vaultSelect.value = app.mainVault!;
 
         return result;
+    }
+
+    private _nameColumnSelectOptions() {
+        const csvHasColumnsOnFirstRow = this._csvHasColumnsOnFirstRowButton?.active;
+
+        return this._itemColumns.map((itemColumn, itemColumnIndex) => ({
+            label: csvHasColumnsOnFirstRow
+                ? `${itemColumn.displayName} (${$l("Column {0}", (itemColumnIndex + 1).toString())})`
+                : $l("Column {0}", (itemColumnIndex + 1).toString()),
+            value: itemColumnIndex,
+        }));
+    }
+
+    private _tagsColumnSelectOptions() {
+        const csvHasColumnsOnFirstRow = this._csvHasColumnsOnFirstRowButton?.active;
+
+        return [
+            { label: $l("None"), value: -1 },
+            ...this._itemColumns.map((itemColumn, itemColumnIndex) => ({
+                label: csvHasColumnsOnFirstRow
+                    ? `${itemColumn.displayName} (${$l("Column {0}", (itemColumnIndex + 1).toString())})`
+                    : $l("Column {0}", (itemColumnIndex + 1).toString()),
+                value: itemColumnIndex,
+            })),
+        ];
+    }
+
+    private _handleNameColumnChange() {
+        const currentNameColumnIndex = this._itemColumns.findIndex((itemColumn) => itemColumn.type === "name");
+        const nameColumnIndex = this._nameColumnSelect?.value || 0;
+
+        this._itemColumns[nameColumnIndex].type = "name";
+
+        if (currentNameColumnIndex !== -1) {
+            this._itemColumns[currentNameColumnIndex].type = FieldType.Text;
+        }
+
+        this._parseData();
+    }
+
+    private _handleTagsColumnChange() {
+        const currentTagsColumnIndex = this._itemColumns.findIndex((itemColumn) => itemColumn.type === "tags");
+        const tagsColumnIndex = this._tagsColumnSelect?.value || 0;
+
+        this._itemColumns[tagsColumnIndex].type = "tags";
+
+        if (currentTagsColumnIndex !== -1) {
+            this._itemColumns[currentTagsColumnIndex].type = FieldType.Text;
+        }
+
+        this._parseData();
+    }
+
+    private _handleFieldNameChange(itemColumnIndex: number) {
+        const newFieldNameInput = this._fieldNameInputs[itemColumnIndex];
+        const newFieldName = newFieldNameInput?.value;
+        if (newFieldName) {
+            this._itemColumns[itemColumnIndex].displayName = newFieldName;
+            this._parseData();
+        }
+    }
+
+    private _handleFieldTypeChange(itemColumnIndex: number) {
+        const newValue = this._fieldTypeSelects[itemColumnIndex]?.value;
+        if (newValue) {
+            this._itemColumns[itemColumnIndex].type = newValue;
+            this._parseData();
+        }
     }
 
     private async _parseData(resetCSVColumns = false): Promise<void> {
@@ -326,7 +346,6 @@ export class ImportDialog extends Dialog<File, void> {
         }
 
         app.addItems(this._items, vault);
-        // this.dispatch("data-imported", { items: items });
         this.done();
         alert($l("Successfully imported {0} items.", this._items.length.toString()), { type: "success" });
     }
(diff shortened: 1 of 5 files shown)


packages/app/src/elements/import-dialog.ts
```diff
@@ -12,6 +12,13 @@ import { html } from "lit";
...
 import { stringToBytes } from "@padloc/core/src/encoding";
 
+const fieldTypeOptions = Object.keys(FIELD_DEFS).map((fieldType) => ({
+    name: FIELD_DEFS[fieldType].name as string,
+    value: fieldType,
+}));
+// TODO: Remove this
+console.log(fieldTypeOptions);
+
 @customElement("pl-import-dialog")
```
BrunoBernardino, line 22: @MaKleSoft is there a better UI element to use for this? If I'm limited to this dialog's "width", I'll list the columns vertically (which will also work for however many columns there are, instead of a horizontal scroll), though it's less familiar (visually). Also, I couldn't find any other UX pattern with such a complex/large dialog, and am not sure we should introduce one.

packages/app/src/elements/import-dialog.ts
```diff
@@ -40,14 +50,22 @@ export class ImportDialog extends Dialog<File, void> {
                 ></pl-select>
 
                 <div class="small padded" ?hidden=${this._formatSelect && this._formatSelect.value !== imp.CSV.value}>
-                    ${$l(
-                        "IMPORTANT: Before importing, please make sure that your CSV data " +
-                            "is structured according to {0}'s specific requirements!",
-                        process.env.PL_APP_NAME!
-                    )}
+                    ${$l("If you don't want to map columns to field types in the next step")}
```
BrunoBernardino, line 53: I'll probably need to rephrase this once I have a better UI.

packages/app/src/lib/import.ts
```diff
@@ -112,6 +118,29 @@ export async function asCSV(file: File, nameColIndex?: number, tagsColIndex?: nu
...
+    const parsed = papa.parse(data);
+    if (parsed.errors.length) {
+        throw new Err(ErrorCode.INVALID_CSV, "Failed to parse .csv file.");
+    }
+    const rows = parsed.data as string[][];
+
+    const columnNames = rows[0].map((column) => column.toLowerCase());
+
+    return columnNames.map((columnName) => {
+        const type = (Object.keys(FIELD_DEFS).filter((fieldType) => columnName.includes(fieldType))[0] ||
```
BrunoBernardino, line 133: I initially tried to use the `FIELD_DEFS.*.pattern` for matching, but many of them are `.*` so I didn't get a good "weight" for them anyway. This seems good enough, and we can later add "similarNames" or something like that to `FIELD_DEFS.*`, and use here as matching options. Could also help with #223 (which this kind of solves already)

packages/app/src/elements/import-dialog.ts
```diff
@@ -119,6 +137,7 @@ Github,"work,coding",https://github.com,john.doe@gmail.com,129lskdf93`)
                 this._items = await imp.asLastPass(file);
                 break;
             case imp.CSV.value:
+                this._itemColumns = await imp.asCSVColumns(file);
```
BrunoBernardino, line 140: Alternatively, I can make `.asCSV` return both `items` and `itemColumns`, so the file's only read once. I'll probably do that later.

BrunoBernardino:
@MaKleSoft very initial work here, the dialog question is my main "blocker", but I've got plenty to go on before that (the actual logic to showing the column dropdowns and making that reactive, then using that mapping when parsing as well). Thoughts appreciated.

packages/app/src/elements/import-dialog.ts
BrunoBernardino: @MaKleSoft in complexity I meant in terms of layering (one dialog -> import -> column mapping). Those examples are still helpful, though!

packages/app/src/lib/import.ts
BrunoBernardino: Thanks, that's helpful!

BrunoBernardino:
Alright, this is now functionally working (choosing field types and having those be applied for the import), and I honestly enjoyed the horizontal list of columns, would like your opinion on whether you think that's good, or if we need something else.

Note there are still quite a few things missing (I want to allow choosing the "name" and "tags" column as well, which is a different kind of "field type" which will require a validation that there's only one column with it, and there's a bunch of other `TODO`s in the code).

packages/app/src/elements/import-dialog.ts
```diff
@@ -20,12 +26,24 @@ export class ImportDialog extends Dialog<File, void> {
...
     private _vaultSelect: Select<Vault>;
 
+    // static styles = [
+    //     ...Dialog.styles,
+    //     css`
+    //         :host {
+    //             --pl-dialog-max-width: 40em;
+    //         }
+    //     `,
+    // ];
```
BrunoBernardino, line 45: @MaKleSoft So I tried this, but the dialog got transparent and full-width. What's wrong (note I also might not need this, but would like to understand, regardless)?

packages/app/src/elements/import-dialog.ts
```diff
@@ -40,14 +58,41 @@ export class ImportDialog extends Dialog<File, void> {
...
-                    )}
+                    ${$l("Choose the field type for each column. If you are having trouble,")}
                     <a href="#" @click=${this._downloadCSVSampleFile}> ${$l("Download Sample File")} </a>
                 </div>
 
+                <div
+                    class="vertical stretching spacing layout"
+                    ?hidden=${this._formatSelect && this._formatSelect.value !== imp.CSV.value}
+                >
+                    <ul>
```
BrunoBernardino, line 69: I'll later move this section to its own element.

packages/app/src/elements/import-dialog.ts
MaKleSoft: Hm, working fine for me...

packages/app/src/elements/import-dialog.ts
```diff
@@ -40,14 +58,41 @@ export class ImportDialog extends Dialog<File, void> {
...
+                                    <pl-select
+                                        id=${`itemColumnSelect-${itemColumnIndex}`}
+                                        icon=${FIELD_DEFS[itemColumn.type].icon}
+                                        .label=${itemColumn.displayName}
+                                        .options=${fieldTypeOptions}
+                                        .value=${itemColumn.type}
+                                        @change=${() => {
+                                            const thisElement = this.shadowRoot?.querySelector(
+                                                `#itemColumnSelect-${itemColumnIndex}`
+                                            ) as HTMLSelectElement;
```
MaKleSoft, line 82: I know you're still sketching and probably aware of this, but this can be done more elegantly with the `@query` decorator (see the `_formatSelect` and `_vaultSelect` properties).

MaKleSoft: Oh, nevermind... I just realised you have a select for each column so that won't work. Another option would be to use `@queryAll` and then `this._fieldTypeSelects[itemColumnIndex]`.

MaKleSoft: (One of the advantages of using `@query` and `@queryAll` is that you can define the types of the element and in this case even of the value, so you wouldn't have to do any casting here. E.g.:

```ts
@queryAll("pl-select.field-type-select")
_fieldTypeSelects: Select<FieldType>[];
```
and then yo could simply do

```ts
const newValue = this._fieldTypeSelects[itemColumnIndex]?.value; // Already has type `FieldType`
```

BrunoBernardino: Thanks, I couldn't get `@query` to work, and didn't realize `@queryAll` could also be helpful. I'll try that as cleanup.

MaKleSoft:
I like the simplicity of your design, but I think we need something slightly more complex. First of all, we need the ability to specify custom field names since we can't rely on the fact that the first row will always contain field names (we also need an option to specify that btw). It would also be nice to have a preview of the column contents, which isn't possible with a mere select element. Finally, the whole thing needs to be scrollable. What do you think about this:

Screenshot 2022-03-05 at 10 37 09

It takes up a bit more space but I think it is a little more obvious what's going on and it gives us more options overall. Here is the code I used, for reference (you'll also have to add the `fit-vertically` class to the top-level `div` to make sure the scrolling works):

```html
<pl-scroller
    class="stretch"
    ?hidden=${this._formatSelect && this._formatSelect.value !== imp.CSV.value}
>
    <ul class="vertical spacing layout">
        ${this._itemColumns.map(
            (itemColumn, itemColumnIndex) => html`
                <li class="padded box vertical spacing layout">
                    <div class="small margined spacing horizontal layout">
                        <div class="stretch">${$l("Column {0}", itemColumnIndex.toString())}</div>
                        <div class="subtle">${itemColumn.name}</div>
                    </div>

                    <div class="tiny horizontally-margined subtle mono">
                        Example Value 1, Example Value 2
                    </div>

                    <pl-input .label=${$l("Field Name")} .value=${itemColumn.displayName}></pl-input>

                    <pl-select
                        id=${`itemColumnSelect-${itemColumnIndex}`}
                        icon=${FIELD_DEFS[itemColumn.type].icon}
                        .label=${$l("Field Type")}
                        .options=${fieldTypeOptions}
                        .value=${itemColumn.type}
                        @change=${() => {
                            const thisElement = this.shadowRoot?.querySelector(
                                `#itemColumnSelect-${itemColumnIndex}`
                            ) as HTMLSelectElement;
                            const newValue = thisElement?.value as FieldType;
                            if (newValue) {
                                this._itemColumns[itemColumnIndex].type = newValue;
                                this._parseData();
                            }
                        }}
                    ></pl-select>
                </li>
            `
        )}
    </ul>
</pl-scroller>
```

BrunoBernardino:
I really like your suggestion. I was honestly trying to go for something simple because when you look at how other tools implement this sort of customization, they really only offer you "one select" per column. When more is shown, it can get visually daunting very quickly.

My goal was to implement a simpler version (similar to what I have now, but with the suggested and planned improvements), and if people still had trouble with that, eventually improve it. If you think there's no point in "splitting" this work, I'm happy to go for a version like what you suggest right away (including supporting not having a row with column names).

MaKleSoft:
I also noticed that if you select a second file for import, it still seems to be using the first file. Not sure what's up with that.

MaKleSoft:
> My goal was to implement a simpler version (similar to what I have now, but with the suggested and planned improvements), and if people still had trouble with that, eventually improve it. If you think there's no point in "splitting" this work, I'm happy to go for a version like what you suggest right away (including supporting not having a row with column names).

Totally agree with that approach in general, but CSV import is one of those things where you simply can't get around a certain level of complexity without severely limiting the usefulness. Also in this case if you make it too simply people are actually **more** likely to get confused I think since it's not obvious what is happening behind the scenes. By being more explicit about the column numbers and field names we make it a little easier to understand what's actually going on, even if the UI is a little more to take in at first.

packages/app/src/elements/import-dialog.ts
BrunoBernardino: Eh, actually worked this time. I might've had something else going on.

packages/app/src/elements/import-dialog.ts
```diff
@@ -20,15 +26,66 @@ export class ImportDialog extends Dialog<File, void> {
...
+    static styles = [
+        ...Dialog.styles,
+        css`
+            :host {
+                --pl-dialog-max-width: 40em;
+            }
+        `,
+    ];
+
+    // TODO: Fix dialog reset after closed (when re-opened)
```
BrunoBernardino, line 62: @MaKleSoft do you have any idea on where should I look at? I saw some `reset` functions in other elements but I haven't investigated yet on what's that about, etc.

packages/app/src/elements/import-dialog.ts
```diff
@@ -20,15 +26,66 @@ export class ImportDialog extends Dialog<File, void> {
...
+
     renderContent() {
+        // TODO: Fix wrong first selected value for selects
```
BrunoBernardino, line 65: @MaKleSoft also, do you have any idea on what I should look at? These two `pl-select`s just start with the first option, no matter what I set below. I've tried setting them with `.value` like the other `pl-select`s but it didn't work.

I've tried a few things around using a `string` and a `number`, in case that was causing issues.

packages/core/src/item.ts
```diff
@@ -75,7 +75,7 @@ export const FIELD_DEFS: { [t in FieldType]: FieldDef } = {
     },
     [FieldType.Url]: {
         type: FieldType.Url,
-        pattern: ".*",
+        pattern: /[(http(s)?):\/\/(www\.)?a-zA-Z0-9@:%._\+~#=]{2,256}\.[a-z]{2,8}\b([-a-zA-Z0-9@:%_\+.~#?&//=]*)/i,
```
BrunoBernardino, line 78: This won't recognize unicode URLs, but I figured it was a nice improvement anyway.

packages/app/src/elements/import-dialog.ts
```diff
@@ -40,14 +97,126 @@ export class ImportDialog extends Dialog<File, void> {
...
 
                 <div class="small padded" ?hidden=${this._formatSelect && this._formatSelect.value !== imp.CSV.value}>
-                    ${$l(
-                        "IMPORTANT: Before importing, please make sure that your CSV data " +
-                            "is structured according to {0}'s specific requirements!",
-                        process.env.PL_APP_NAME!
-                    )}
-                    <a href="#" @click=${this._downloadCSVSampleFile}> ${$l("Download Sample File")} </a>
+                    ${$l("Choose the correct column names and types for each column below. If you are having trouble,")}
+                    <a href="#" @click=${this._downloadCSVSampleFile}> ${$l("Download the Sample File")} </a>.
```
MaKleSoft, line 101: I don't really think the sample file makes sense here any more. Instead, there should be a link to the (yet to be written) manual page. So I guess just remove the second part for now, and I'll add the link to the manual page once it exists.

packages/app/src/elements/import-dialog.ts
```diff
@@ -40,14 +97,126 @@ export class ImportDialog extends Dialog<File, void> {
...
+                <pl-scroller
+                    class="stretch"
+                    ?hidden=${this._formatSelect && this._formatSelect.value !== imp.CSV.value}
+                >
+                    <ul class="vertical spacing layout">
+                        <div class="padded">
+                            <label>
+                                <input type="checkbox" id="csvHasDataOnFirstRowCheckbox" />
+                                ${$l("Check if the CSV file has data on the first row (instead of the column names)")}
+                            </label>
```
MaKleSoft, line 113: Please use `<pl-toggle-button></pl-toggle-button>` for this to make it more consistent with the rest of the UI. Also, for the label I'd prefer "First row contains field names" which should be checked by default.

packages/app/src/elements/import-dialog.ts
```diff
@@ -40,14 +97,126 @@ export class ImportDialog extends Dialog<File, void> {
...
+                                );
+                                const nameColumnIndex = this._nameColumnSelect?.value || 0;
+                                this._itemColumns[nameColumnIndex].type = "name";
+
+                                if (currentNameColumnIndex !== -1) {
+                                    this._itemColumns[currentNameColumnIndex].type = FieldType.Text;
+                                }
+
+                                this._parseData();
+                            }}
```
MaKleSoft, line 136: For anything longer than a one-liner, let's use methods instead of inline functions.

packages/app/src/lib/import.ts
```diff
@@ -48,49 +55,49 @@ export function loadPapa(): Promise<any> {
...
-    // Use first row for column names
-    const colNames = data[0];
-
-    if (nameColIndex === undefined) {
-        const i = colNames.indexOf("name");
-        nameColIndex = i !== -1 ? i : 0;
+async function fromTable(
+    data: string[][],
+    columnTypes: ImportCSVColumn[],
+    dataOnFirstRow: boolean
```
MaKleSoft, line 68: Minor nitpick, but I think the reverse makes more sense to me somehow (probably because that's what I'm used from other apps). I.e. `firstRowHasFieldNames`.

packages/app/src/lib/import.ts
```diff
@@ -102,16 +109,90 @@ export async function isCSV(data: string): Promise<Boolean> {
...
+
+                  if (lowerCaseColumnName === "name" || lowerCaseColumnName === $l("name")) {
+                      type = "name";
+                  }
+
+                  if (
+                      lowerCaseColumnName === "tags" ||
+                      lowerCaseColumnName === $l("tags") ||
+                      lowerCaseColumnName === "category" ||
+                      lowerCaseColumnName === $l("category")
```
MaKleSoft, line 150: For conditionals like this I generally prefer 

```ts
["tags", ...].includes(lowerCaseColumnName).
```
but that's just a matter of preference I guess.

MaKleSoft: Also, I'm not sure if it's worth comparing it to the translated value, since there is a fairly low chance of the translation just happening to be what the author of the file chose for the tags/category column. If you do, though, you should probably `.toLowerCase()` the translation.

packages/core/src/item.ts
```diff
@@ -52,7 +52,7 @@ export interface FieldDef {
 export const FIELD_DEFS: { [t in FieldType]: FieldDef } = {
     [FieldType.Username]: {
         type: FieldType.Username,
-        pattern: ".*",
+        pattern: /.*/,
```
MaKleSoft, line 55: Not a bad idea switching to a regex literal, but please make sure it doesn't interfere with this: https://github.com/padloc/padloc/blob/v4/packages/app/src/elements/field.ts#L341

packages/core/src/item.ts
```diff
@@ -161,7 +161,7 @@ export const FIELD_DEFS: { [t in FieldType]: FieldDef } = {
     },
     [FieldType.Totp]: {
         type: FieldType.Totp,
-        pattern: ".*",
+        pattern: /.*/,
```
MaKleSoft, line 164: This is actually very easy to recognise, since it's base32-encoded. See: https://github.com/google/google-authenticator/wiki/Key-Uri-Format#secret

packages/core/src/item.ts
```diff
@@ -296,22 +296,37 @@ export async function createVaultItem({
...
+    name: string;
+    value?: string;
+    masked?: boolean;
+}): FieldType {
+    if (masked) {
+        return FieldType.Password;
+    }
+
+    let matchedTypeByName = Object.keys(FIELD_DEFS).filter((fieldType) =>
+        new RegExp(fieldType, "i").test(name)
```
MaKleSoft, line 314: Isn't `fieldType` and object here?

packages/core/src/item.ts
```diff
@@ -296,22 +296,37 @@ export async function createVaultItem({
...
+}: {
+    name: string;
+    value?: string;
+    masked?: boolean;
+}): FieldType {
+    if (masked) {
+        return FieldType.Password;
+    }
+
+    let matchedTypeByName = Object.keys(FIELD_DEFS).filter((fieldType) =>
```
MaKleSoft, line 313: Minor nitpick, by why not `const`?

packages/core/src/item.ts
MaKleSoft: Oh, nevermind... just saw you're using `Object.keys()`.

packages/app/src/elements/import-dialog.ts
```diff
@@ -119,7 +288,9 @@ Github,"work,coding",https://github.com,john.doe@gmail.com,129lskdf93`)
                 this._items = await imp.asLastPass(file);
                 break;
             case imp.CSV.value:
-                this._items = await imp.asCSV(file);
+                const result = await imp.asCSV(file, this._itemColumns, this._csvHasDataOnFirstRowCheckbox.checked);
+                this._items = result.items;
+                this._itemColumns = result.itemColumns;
```
MaKleSoft, line 293: This seems kind of circular. Are the `itemColumns` a result or an input of this function? I think it should be either/or. Better to have a separate function that infers the initial item column definitions and another that actually parses the data into vault items (taking the column definitions as an input). So something like this:

```ts
function inferCSVColumns(file: File, firstRowHasData: boolean): CSVItemColumn[] {}
function asCSV(file: File, itemColumns: CSVItemColumn[], firstRowHasData: boolean): VaultItem[] {}
```
Or, if you don't want to read/parse the file twice:

```ts
function inferCSVColumns(data: string[][], firstRowHasData: boolean): CSVItemColumn[] {}
function asCSV(data: string[][], itemColumns: CSVItemColumn[], firstRowHasData: boolean): VaultItem[] {}
```

packages/app/src/elements/import-dialog.ts
MaKleSoft: Anything needing resetting should be done in the `show()` method.

MaKleSoft: In this case you should reset `this._itemColumns` and the value of `this._csvHasDataOnFirstRowCheckbox`.

packages/app/src/elements/import-dialog.ts
MaKleSoft: Actually I just realised my second definition for `asCSV` is exactly what `asTable` is. Well in any case I think there should be a clearer separation of concerns with these functions, but you don't have to untangle this right now if you don't want to.

packages/app/src/elements/import-dialog.ts
BrunoBernardino: Thanks for the reminder, I'll still clean this all up before I'm done.

packages/app/src/lib/import.ts
BrunoBernardino: Naming things! I personally don't like _true defaults_, but I can go with that and default to having the toggle on.

packages/core/src/item.ts
BrunoBernardino: I was making some changes before, it should totally be `const` when not reassigned.

packages/app/src/elements/import-dialog.ts
BrunoBernardino: I don't feel strongly about it right now. I do like reusing it like this: if there's something set, use it, otherwise, parse and send it back. Anyway, as this comes closer to being done, we can see what makes sense to untangle or not. I don't want to refactor unnecessarily until this is all working as expected.

packages/app/src/lib/import.ts
MaKleSoft: > Naming things!

Hahah, yeah... Well, the naming of the variable in the code I'll leave up to you, but for the user-facing option I do think "First row contains field names" is the clearer naming. Why? Because if you write "First row contains data" the natural question is: "What else would it contain?". Whereas "contains field names" makes it clear that if the option is disabled, the row will be treated as all other rows. Does that make sense?

packages/app/src/elements/import-dialog.ts
BrunoBernardino: @MaKleSoft not sure if you missed this one, but I'm more at a loss for it right now.

packages/app/src/elements/import-dialog.ts
```diff
@@ -71,28 +72,28 @@ export class ImportDialog extends Dialog<File, void> {
...
-                    ?hidden=${this._formatSelect && this._formatSelect.value !== imp.CSV.value}
-                >
+                <pl-scroller class="stretch" ?hidden=${this._formatSelect?.value !== imp.CSV.value}>
                     <ul class="vertical spacing layout">
                         <pl-toggle-button
                             class="transparent"
                             id="csvHasColumnsOnFirstRowButton"
                             .label=${$l("First row contains field names")}
                             reverse
+                            @change=${() => this._parseData(true)}
```
MaKleSoft, line 86: This is required for the change to take effect

packages/app/src/elements/import-dialog.ts
```diff
@@ -149,12 +150,22 @@ export class ImportDialog extends Dialog<File, void> {
                                     ?hidden=${itemColumn.type === "name" || itemColumn.type === "tags"}
                                 >
                                     <div class="small margined spacing horizontal layout">
-                                        <div class="stretch">${itemColumn.name}</div>
-                                        <div class="subtle">${$l("Column {0}", itemColumnIndex.toString())}</div>
+                                        <div class="stretch">
+                                            ${csvHasColumnsOnFirstRow
+                                                ? itemColumn.name
+                                                : $l("Column {0}", itemColumnIndex.toString())}
```
MaKleSoft, line 156: If the first row is not column names, we can't expect meaningful/unique names here so better to just show the column number.

packages/app/src/elements/import-dialog.ts
```diff
@@ -149,12 +150,22 @@ export class ImportDialog extends Dialog<File, void> {
...
                                     </div>
 
-                                    <div class="tiny horizontally-margined subtle mono">
-                                        ${itemColumn.exampleValues}
+                                    <div class="tiny horizontally-margined subtle mono ellipsis">
+                                        ${itemColumn.values
+                                            .filter((v) => v !== "")
+                                            .slice(0, 20)
+                                            .map((value) => (value.includes(",") ? `"${value}"` : value))
+                                            .join(", ")}
```
MaKleSoft, line 168: Why limit the number of example values? Let's use all the space we have :)

packages/app/src/elements/import-dialog.ts
```diff
@@ -224,35 +235,14 @@ export class ImportDialog extends Dialog<File, void> {
         const importFormat = (await imp.guessFormat(file)) || imp.CSV;
         this._formatSelect.value = importFormat.value;
 
-        // Select first options properly
-        if (importFormat.value === imp.CSV.value) {
-            if (
-                this._nameColumnSelect &&
-                (this._nameColumnSelect.selectedIndex === -1 || this._nameColumnSelect.selectedIndex === undefined)
```
MaKleSoft: This conditional is inappropriate since we're resetting the value so we want to we don't care if another value had been selected previously.

packages/app/src/elements/import-dialog.ts
```diff
@@ -281,9 +271,17 @@ export class ImportDialog extends Dialog<File, void> {
                 this._items = await imp.asLastPass(file);
                 break;
             case imp.CSV.value:
-                const result = await imp.asCSV(file, this._itemColumns, this._csvHasColumnsOnFirstRowButton.active);
+                const result = await imp.asCSV(
+                    file,
+                    resetCSVColumns ? [] : this._itemColumns,
+                    this._csvHasColumnsOnFirstRowButton.active
+                );
```
MaKleSoft, line 278: Item columns need to be reset whenever the `hasColumnsOnFirstRow` value changes, but resetting the value to `[]` before re-parsing causes the UI to be updated twice which leads to some ugly jumpiness.

packages/app/src/elements/import-dialog.ts
```diff
@@ -281,9 +271,17 @@ export class ImportDialog extends Dialog<File, void> {
...
                 this._items = result.items;
                 this._itemColumns = result.itemColumns;
+                console.log("itemcolumns", this._itemColumns);
+                await this.updateComplete;
```
MaKleSoft, line 282: This is what you were missing with regards to the select value not taking hold. Basically the problem was that the `option` elements associated with each column didn't have time to be rendered, which is why the value wasn't properly accepted. This allows the select element to update properly before setting the value.

packages/app/src/lib/import.ts
```diff
@@ -18,7 +18,7 @@ export interface ImportCSVColumn {
     name: string;
     displayName: string;
     type: FieldType | "name" | "tags";
-    exampleValues: string;
+    values: string[];
```
MaKleSoft, line 21: Formatting the example values for display is a concern of the UI, so I thought is was better to push the "stringifying" further up the chain.

packages/app/src/lib/import.ts
```diff
@@ -125,59 +125,52 @@ export async function asCSV(
...
-                itemColumn.type = FieldType.Text;
-            }
-        } else if (itemColumn.type === "tags") {
-            if (!hasTagsColumn) {
-                hasTagsColumn = true;
-            } else {
-                itemColumn.type = FieldType.Text;
-            }
-        }
-    });
```
MaKleSoft: Not sure why you decided to do this in a second pass, but I thought it was unnecessary so I added the deduplication to the initial loop. I hope I didn't miss anything!

packages/app/src/lib/import.ts
```diff
@@ -125,59 +125,52 @@ export async function asCSV(
...
             : columnNames.map((columnName, columnIndex) => {
+                  const values = rows.map((row) => row[columnIndex] || "");
+
+                  // Guess field type based on first non-empty value
+                  // TODO: Sample all values for more reliable results?
                   let type = guessFieldType({
-                      name: columnName,
-                      value: rows[1] ? rows[1][columnIndex] : "",
+                      name: columnsOnFirstRow ? columnName : "",
+                      value: values.find((v) => Boolean(v)),
```
MaKleSoft, line 150: There is no guarantee that the first row has data in all columns.

MaKleSoft:
Fixed the problem with setting the select values and made a couple of other improvements as well while I was at it. Hope that's ok!

packages/app/src/elements/import-dialog.ts
BrunoBernardino: I just figured it could be confusing, but you're right.

packages/app/src/elements/import-dialog.ts
BrunoBernardino: Gotcha.

packages/app/src/elements/import-dialog.ts
BrunoBernardino: Thanks. I didn't see it in the documentation, but I found it now!

BrunoBernardino:
Thank you so much, @MaKleSoft ! The changes you made look great and I'll address your comments and cleanup I still wanted to do next week.

BrunoBernardino:
Alright, I tested some more complex CSVs and found and fixed a bunch of issues, right now there are just a couple of things missing:
- Add option to ignore/skip column
- Refactor CSV UI into its own component

packages/app/src/elements/import-dialog.ts
```diff
@@ -92,8 +92,11 @@ export class ImportDialog extends Dialog<File, void> {
                             .label=${$l("Name Column")}
                             .options=${this._itemColumns.map((itemColumn, itemColumnIndex) => ({
                                 label: csvHasColumnsOnFirstRow
-                                    ? `${itemColumn.displayName} (${$l("Column {0}", itemColumnIndex.toString())})`
-                                    : $l("Column {0}", itemColumnIndex.toString()),
+                                    ? `${itemColumn.displayName} (${$l(
+                                          "Column {0}",
+                                          (itemColumnIndex + 1).toString()
```
BrunoBernardino, line 97: We shouldn't be starting with "Column 0" but "Column 1"

packages/core/src/item.ts
```diff
@@ -32,8 +32,10 @@ export enum FieldType {
 export interface FieldDef {
     /** content type */
     type: FieldType;
-    /** regular expression describing pattern of field contents (not used for validation) */
+    /** regular expression describing pattern of field contents (used for validation) */
     pattern: RegExp;
+    /** regular expression describing pattern of field contents (used for matching) */
+    matchPattern: RegExp;
```
BrunoBernardino, line 38: I've created this other one that won't affect validation, and can be "stricter" (like forcing a `\n` to exist for notes) for matching.

packages/core/src/item.ts
```diff
@@ -318,9 +332,13 @@ export function guessFieldType({
         return matchedTypeByName;
     }
 
-    const matchedTypeByValue = Object.keys(FIELD_DEFS).filter((fieldType) =>
-        FIELD_DEFS[fieldType].pattern.test(value)
-    )[0] as FieldType;
+    // We skip some because they can match anything, and are only really valuable when matched by name
+    const fieldTypesToSkipByValue = [FieldType.Username, FieldType.Password];
```
BrunoBernardino, line 336: If I didn't skip these, it would always be matched as them because they allow anything.

MaKleSoft (approved):
Looks good to merge! I haven't tested it extensively, but I trust that you've already done that 😉. I'm really happy with this turned out! This is could potentially make a big difference for onboarding new users. Thanks!

BrunoBernardino:
I have tested the sample.csv, a more complex CSV with different kinds of fields (including columns to skip and no columns row), and an older 1Password csv export I had. I'm confident this will work in most scenarios!

Issue #342, opened 2021-12-23
Password Audit Page


MaKleSoft:
- [ ] Checks against HIBP
- [ ] Checks against password blacklist
- [ ] Expired Passwords
- [ ] Duplicate Passwords
- [ ] Passwords which used to be accessible to former Org members
- [ ] Password entropy
- [ ] Unused Vault Items (stale accounts)

MaKleSoft:
Closing this, as we'll track future improvements/additions in separate issues

PR #414, opened 2022-03-21
Security Audit v1

BrunoBernardino:
This adds a password audit page with some mocked elements and routing, plus some thoughts in inline comments.

Related to #342


packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,98 @@
...
+@customElement("pl-audit")
+export class Audit extends Routing(LitElement) {
+    @state()
+    private _reusedPasswordCount: number = 0;
+
+    @state()
+    private _weakPasswordCount: number = 0;
+
+    @state()
+    private _compromisedPasswordCount: number = 0;
```
BrunoBernardino, line 22: I'm thinking for a "v1" of Password Audit page, to show these 3 counts, and allow seeing items filtered by them.

packages/app/src/elements/items-list.ts
```diff
@@ -37,6 +37,9 @@ export interface ItemsFilter {
     attachments?: boolean;
     recent?: boolean;
     host?: boolean;
+    auditWeak?: boolean;
+    auditReused?: boolean;
+    auditCompromised?: boolean;
```
BrunoBernardino, line 42: These are the 3 filters I plan on adding, to allow just routing to `/items` and display them.

BrunoBernardino:
@MaKleSoft no need to review any code, but I have some comments/questions, and just want to make sure we're on the same page for this.

packages/app/src/elements/items-list.ts
MaKleSoft: I had envisioned simply listing the items in the audit page but I like this idea!

packages/app/src/elements/menu.ts
```diff
@@ -548,6 +548,16 @@ export class Menu extends Routing(StateMixin(LitElement)) {
                         <div class="stretch">${$l("Password Generator")}</div>
                     </div>
 
+                    <div
+                        class="menu-item"
+                        @click=${() => this._goTo("audit")}
+                        aria-selected=${this.selected === "audit"}
+                    >
+                        <pl-icon icon="info-round"></pl-icon>
```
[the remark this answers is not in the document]
MaKleSoft: You can use any icons from the fontawesome icon set: https://fontawesome.com/icons You just have to add them to    https://github.com/padloc/padloc/blob/v4/packages/app/src/elements/icon.ts (if they haven't been yet). I like this one for the audit page: https://fontawesome.com/icons/shield-check?s=solid

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,98 @@
...
+        `,
+    ];
+
+    render() {
+        const { _reusedPasswordCount, _weakPasswordCount, _compromisedPasswordCount } = this;
+        return html`
+            <div class="padded layout">
+                <div class="vertical spacing">
+                    <h3>
+                        Here you can see some counts for passwords you should change, as they might be easier to be
```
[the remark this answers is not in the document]
MaKleSoft: I like the idea of using filters and routing to the list page, but I think it would be cool to display at least the first 5 of each. You could use the same layout as in https://github.com/padloc/padloc/blob/v4/packages/app/src/elements/org-dashboard.ts, which a "box" for weak, reused, compromised and then up to 5 items in each box (that route to the item directly) and a "View All" button that routes to the list view.

packages/app/src/elements/audit-view.ts
```diff
@@ -0,0 +1,57 @@
...
+                    >
+                        <div class="half-margined horizontal spacing center-aligning layout text-left-aligning">
+                            <pl-icon icon="info-round"></pl-icon>
+                            <div class="stretch ellipsis">${$l("Password Audit")}</div>
+                        </div>
+                    </pl-button>
+                </header>
+                <pl-scroller class="stretch">
+                    <div class="layout padded">
+                        <pl-audit></pl-audit>
```
MaKleSoft, line 51: Any specific reason why you decided to use a separate component for this?

BrunoBernardino: I do expect it to have a lot of complex logic, and this would allow it to be eventually reused, but I only separated it _now_ because it was how the component I was copying was already setup. I'm happy to bring it all in here, if you'd like that.

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,214 @@
...
+
+        // TODO: Remove this
+        console.log("======== rendering!");
+        console.log(
+            JSON.stringify({
+                reusedPasswordsCount: _reusedPasswords.length,
+                weakPasswordsCount: _weakPasswords.length,
+                compromisedPasswordsCount: _compromisedPasswords.length,
+            })
+        );
```
BrunoBernardino, line 112: @MaKleSoft I'm sure this is a noob-in-lit kind of problem, but I can get new values here, while the rendered component only shows whatever came the first time around. I even tried `this.stateChanged();` in `this.audit()` to force it to update, but no go. I also played around with `@property` and `@state` inside the `pl-audit-list-item`, in case that was the thing not updating.

What am I doing wrong?

packages/app/src/elements/icon.ts
```diff
@@ -545,6 +545,10 @@ export class PlIcon extends LitElement {
             :host([icon="celebrate"]) > div::before {
                 content: "\\e383";
             }
+
+            :host([icon="shield-check"]) > div::before {
```
BrunoBernardino, line 549: Nice tip @MaKleSoft looks really nice with this icon.

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,214 @@
...
+                        Here you can see some counts for passwords you should change, as they might be easier to be
+                        easily guessed.
+                    </h3>
+                    <p class="subtle">Click a count to view the affected items.</p>
+                    <div class="counts">
+                        <pl-audit-list-item
+                            .listItems=${_reusedPasswords}
+                            .label=${$l("Reused Passwords")}
+                            .routeParams=${{ auditReused: "true" }}
+                        ></pl-audit-list-item>
```
MaKleSoft, line 127: Again, no reason to create a separate element here. If you want to avoid repetition, you can use a partial render function, like so:

```ts
private _renderSection(listItems: ListItem[], label: string) {
    return html`
            <section class="box count" ?hidden=${listItems.length === 0}>
                <h2 class="uppercase bg-dark border-bottom semibold center-aligning spacing horizontal layout">
                    <div></div>
                    <div>${label}</div>
                    <div class="tiny bold tag">${listItems.length}</div>
                    <div class="stretch"></div>
                </h2>
                <pl-list>
                    ${listItems.slice(0, 5).map(
                        (listItem) => html`
                            <div class="list-item hover click" @click=${() => this.go(`items/${listItem.item.id}`)}>
                                <pl-vault-item-list-item
                                    .item=${listItem.item}
                                    .vault=${listItem.vault}
                                    class="stretch collapse"
                                ></pl-vault-item-list-item>
                            </div>
                        `
                    )}
                </pl-list>
                <pl-button
                    class="slim margined transparent"
                    @click=${() => this.go("items", routeParams)}
                    ?hidden=${listItems.length < 6}
                >
                    <div>${$l("Show All")}</div>
                    <pl-icon icon="arrow-right" class="small left-margined"></pl-icon>
                </pl-button>
            </section>
    `;
}

render() {
    return html`
        ...
        ${this._renderSection(this._reusedPasswords, $l("Reused Passwords")}
        ${this._renderSection(this._weakPassword, $l("Weak Passwords")}
        ${this._renderSection(this._compromisedPasswords, $l("Compromised Passwords")}
        ...
    `;
}
```

But even that might be "premature generalisation". I found that often it's better to repeat yourself a little than unnecessarily complicating things by trying to deduplicate code.

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,214 @@
...
+                    </div>
+                </div>
+            </div>
+        `;
+    }
+
+    connectedCallback() {
+        super.connectedCallback();
+        this.addEventListener("change", () => this.audit());
```
MaKleSoft, line 146: Not sure what you're trying to achieve with this? The `@change` event is usually used with user input, of which there is none here.

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,214 @@
...
+                // TODO: Store + save (+ sync) last audit date and audit match booleans in items
+            }
+        }
+
+        this._countElements.forEach((countElement) => {
+            animateElement(countElement, { animation: "bounce" });
+        });
+
+        // TODO: This doesn't work, and shouldn't be necessary
+        this.stateChanged();
```
MaKleSoft, line 202: Nope, I've never used this method and don't think this is meant to be called explicitly.

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,214 @@
...
+                        vault,
+                        item,
+                    });
+
+                    checkedItemIds.add(item.id);
+                } else if (randomNumber % 5 === 2) {
+                    this._compromisedPasswords.push({
+                        vault,
+                        item,
+                    });
```
MaKleSoft, line 188: This is where you're going wrong. Reactive properties are implemented using simple getters and setters, and don't pick up on manipulations of complex objects. I.e., in order to trigger an update, you have to update the object (reference) itself. So the pattern you should use looks something like this:

```ts
const reused = [];
const weak = [];
const compromised = [];

// Fill arrays...

this._reusedPasswords = reused;
this._weakPasswords = weak;
this._compromisedPasswords = compromised;
```

MaKleSoft: Alternatively, you *can* trigger updates manually as well, using `requestUpdate`. E.g.:

```ts
this._reusedPasswords.push({vault, item});
this.requestUpdate();
```

See also: https://lit.dev/docs/components/lifecycle/#reactive-update-cycle-triggering

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+    async audit() {
+        this._reusedPasswords = [];
+        this._weakPasswords = [];
+        this._compromisedPasswords = [];
+
+        const { vaults } = this.state;
+
+        // This prevents the same item from being considered more than once for different sections/audits
```
MaKleSoft, line 156: Why? I don't see any reason why an item should not be allowed to appear in multiple sections. The reused/weak/compromised classifications are not mutually exclusive after all!

packages/app/src/elements/audit.ts
BrunoBernardino: I got this from another component and hadn't gotten around to seeing if it was necessary yet.

packages/app/src/elements/audit.ts
BrunoBernardino: Gotcha, I'm happy with the functions stuff. Refactoring this piece actually helped me a lot with tweaking the design, so I wouldn't call it premature :blush:

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+                        isReused = true;
+
+                        // Don't add the same item twice to the list, if there are more than one reused password fields in it
+                        if (!reusedPasswordItemIds.has(item.id)) {
+                            reusedPasswords.push({ item, vault });
+                            reusedPasswordItemIds.add(item.id);
+                        }
+
+                        // TODO: Save audit match boolean in field
+                        // passwordField.auditResult = { ...(passwordField.auditResult || {}), isReused };
```
BrunoBernardino, line 132: @MaKleSoft does something like this (field schema-wise) look alright to you? I'm not sure we need this to be stored in the DB, but at least it should be kept in memory.

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+
+                        // TODO: Save audit match boolean in field
+                        // passwordField.auditResult = { ...(passwordField.auditResult || {}), isWeak };
+                    }
+
+                    // TODO: Perform compromised audit
+                }
+
+                // TODO: Save audit match booleans in item
+                // item.auditResult = { lastAudited: new Date(), isReused, isWeak, isCompromised };
```
BrunoBernardino, line 165: @MaKleSoft and this one for the item itself (so we can filter by).

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+
+                console.log({ item, isReused, isWeak });
+            }
+        }
+
+        // TODO: Sync?
```
BrunoBernardino, line 171: @MaKleSoft I haven't looked into if this is necessary after saving/updating the `auditResult` property in the items and fields, but let me know if there's any gotchas.

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+
+        // This makes the UI update
+        this._reusedPasswords = reusedPasswords;
+        this._weakPasswords = weakPasswords;
+        this._compromisedPasswords = compromisedPasswords;
```
BrunoBernardino, line 176: @MaKleSoft thanks, this worked perfectly!

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+                >
+                    <div>${$l("Show All")}</div>
+                    <pl-icon icon="arrow-right" class="small left-margined"></pl-icon>
+                </pl-button>
+            </section>
+        `;
+    }
+}
+
+async function sha1(stringToHash: string) {
```
BrunoBernardino, line 218: I couldn't find this anywhere (and I'll need it for the HIBP as well), and since it isn't exactly "secure", I didn't think it was worth putting it in `core`, but I can be persuaded otherwise.

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+    const hashedPasswordData = await crypto.subtle.digest("SHA-1", new TextEncoder().encode(stringToHash));
+
+    const hashedPassword = Array.from(new Uint8Array(hashedPasswordData))
+        .map((byte) => byte.toString(16).padStart(2, "0"))
+        .join("");
+
+    return hashedPassword;
+}
+
+function isPasswordWeak(password: string) {
```
BrunoBernardino, line 228: Let me know if there are any obvious "this password is weak" checks I can do here that I missed.

packages/app/src/elements/audit.ts
MaKleSoft: The [crypto](https://github.com/padloc/padloc/blob/master/packages/core/src/crypto.ts) module has got you covered 😉

```ts
import { getCryptoProvider } from "@padloc/core/src/platform";
import { HashParams } from "@padloc/core/src/crypto";
import { stringToBytes } from "@padloc/core/src/encoding";

...

const hashed = await getCryptoProvider()
    .hash(stringToBytes(password), new HashParams({ algorithm: "SHA-1" }));
```

packages/app/src/elements/audit.ts
MaKleSoft: Got [something](https://github.com/padloc/padloc/blob/master/packages/app/src/lib/util.ts#L25) for that as well 😊

```ts
const { score } = await passwordStrength(password);
return score < 2;
```

This uses [zxcvbn](https://github.com/dropbox/zxcvbn), which to this day is the most reliable tool for guessing password entropy (or "strength") that I know of. If you want you can do a little research into whether there is something better out there by now but it's served me pretty well so far!

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,239 @@
...
+    async audit() {
+        const reusedPasswords: ListItem[] = [];
+        const weakPasswords: ListItem[] = [];
+        const compromisedPasswords: ListItem[] = [];
+
+        const { vaults } = this.state;
+
+        const firstMatchPerPasswordHash: {
+            [passwordHash: string]: { item: VaultItem; vault: Vault; passwordField: Field };
+        } = {};
```
MaKleSoft, line 101: Why not use a `Map<string, {...}>` for this?

packages/app/src/elements/audit.ts
MaKleSoft: I think it would be better to store audit results on the item, instead of the field, as some results may affect multiple fields (e.g. the HIBP result would affect a username/password pair) and some may not affect any particular fields at all (stale/unused/expired items). How about something like this :

```ts
// @padloc/core/src/item.ts

export enum AuditResultType {
    WeakPassword = "weak_password",
    ReusedPassword = "reused_password",
    LeakedCredentials = "leaked_credentials",
}

export interface AuditResult {
    type: AuditResultType,
    /** Indexes of affected fields */
    fields: number[],
}

export class VaultItem {
    ...

    auditResults: AuditResult = [];
}
```

This gives us enough flexibility to add more audit types later without doing any major refactoring.

MaKleSoft: With that model in mind, I think I'd also use a different schema for the filter:
 
```ts
export interface ItemsFilter {
    vault?: VaultID;
    tag?: Tag;
    favorites?: boolean;
    attachments?: boolean;
    recent?: boolean;
    host?: boolean;
    audit?: AuditType;
}
```

packages/app/src/elements/audit.ts
MaKleSoft: This is enough for persisting the changes:
```ts
app.save();
```
If you want to manually trigger a sync, you can call:
```ts
app.synchronize();
```
But that's not really necessary, as the app will periodically sync everything automatically.

packages/app/src/elements/audit.ts
MaKleSoft: FYI: You'll also need to call `Vault::updateItem`:

```ts
item.auditResults = ...
item.lastAudited = new Date();
vault.updateItem(item);
```

For updating invidual items, there is also the convenience method `app.updateItem`, but that will also trigger `app.save()` and `app.synchronize()` on every call, so it shouldn't be used for bulk updates:

```ts
app.updateItem(item, { auditResults: ..., lastAudited: new Date() })
```

packages/app/src/elements/audit.ts
BrunoBernardino: @MaKleSoft Thanks for sharing your thoughts here. Maybe I'm misunderstanding something, but HIBP doesn't give you a way to check a combination of username/password. A given account/email was either leaked or not, and a given password was compromised or not.

I like your `AuditResult[]` suggestion better than mine, but I still think we should have store it on the field level (so we can show some visual indicators) as well as the item level (so we can filter). If you feel strongly (since you know the db, app, and use cases much better than me) that we should only store it in the item, we can always run an audit when an item is open and it has an audit result, to see which fields "fail".

packages/app/src/elements/audit.ts
BrunoBernardino: I've never found it to be much more advantageous than a plain object for <1k entries, but I see no harm here, and thinking about it now, I suppose there will be many people with thousands of items. Good suggestion, thanks!

MaKleSoft: I don't think there is a measurable difference in performance, but I think it's semantically more appropriate. Being able to use plain objects as dictionaries is a neat feature in javascript but for me it always felt kinda wrong 😅. Using `Map` makes it clear what kind of data structure we're creating and for what purpose. This is, once again, just a style question though, so I don't feel too strongly either way.

packages/app/src/elements/audit.ts
MaKleSoft: >  Maybe I'm misunderstanding something, but HIBP doesn't give you a way to check a combination of username/password.

Nope, I'm the one who misunderstood 😅 I guess I misremembered how the HIBP api works. Nevermind! I guess in that case we can use

```ts
fieldIndex?: number;
```
instead of 
```ts
fields: number[];
```

Since I can't think of any other audit types that might affect multiple fields.

MaKleSoft: > I like your AuditResult[] suggestion better than mine, but I still think we should have store it on the field level (so we can show some visual indicators) as well as the item level (so we can filter). If you feel strongly (since you know the db, app, and use cases much better than me) that we should only store it in the item, we can always run an audit when an item is open and it has an audit result, to see which fields "fail".

I don't see the point in storing the result in both, since we can always do:

```ts
fieldResults = item.auditResults.filter(res => res.fieldIndex === fieldIndex);
```

No need to re-run the audit.

BrunoBernardino: Ok. I can get behind that! It does mean we need to re-run the audit when an item changes the order of fields, _but_ we'd have to run it anyway because the password fields might change on an update, so I think this works well.

packages/app/src/elements/audit.ts
BrunoBernardino: I don't see a `vault.updateItem()` nor anything similar on `vault` or `item` , only `app.save()`. Is that not enough, or should I use something else?

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,273 @@
...
+                    }
+                }
+
+                item.auditResults = auditResults;
+                item.lastAudited = new Date();
+                vault.items.update(item);
+            }
+        }
+
+        app.save();
```
BrunoBernardino, line 193: @MaKleSoft this updates the items, but after a sync it's reset, so it doesn't persist. What am I missing?

packages/app/src/elements/field.ts
```diff
@@ -391,6 +398,33 @@ export class FieldElement extends LitElement {
                         >
                             <div class="spacer" slot="before"></div>
                             <pl-icon icon="${this._fieldDef.icon}" slot="before"></pl-icon>
+                            <pl-icon
```
BrunoBernardino, line 401: I really liked how this looked!

packages/app/src/elements/audit.ts
MaKleSoft: Ah, I forgot about `vault.commit`. Actually, better use `app.saveVault(vault)` (further up, in the outer loop).

BrunoBernardino:
@MaKleSoft this is coming along really nicely. The two remaining things right now before this is ready are:

1. Filter audit items
2. Run audit information on a schedule

For 1, I think I already know everything I need to, but for 2, do you think I should implement something like `auto-lock.ts` and the alarms in `background.ts`, or do you have another idea? My thinking was to run the audit once a week tops, but the check to see if it needs to run would run either daily, or when the app opens (to avoid a timeout/alarm).

MaKleSoft:
> for 2, do you think I should implement something like auto-lock.ts and the alarms in background.ts, or do you have another idea? My thinking was to run the audit once a week tops, but the check to see if it needs to run would run either daily, or when the app opens (to avoid a timeout/alarm).

I think it would indeed be good to implement this as a mixin, but I don't think we need a scheduled action like in `auto-lock.ts` or even a period run like in `auto-sync.ts` - it should be sufficient to have a method like this, that is called after the app is unlocked and whenever an item is saved:

```ts
// Basically the same as what you have in the `<pl-audit>` component now, 
// but with checks of whether a given audit needs to be run
runAudits() {
    for (const vault of app.vaults){
        for (const item of vault.items) {
            if (!item.lastAudited || item.updated > item.lastAudited) {
                // Check for weakness
                // Check for reuse 
            }

            if (
                !item.lastAudited ||
                item.updated > item.lastAudited ||
                item.lastAudited.getTime() < Date.now() - checkHIBPPeriod
            ) {
                // Check for leaks via HIBP
            }
      }
      ...
}
```

packages/app/src/elements/app.ts
```diff
@@ -126,6 +129,11 @@ export class App extends ServiceWorker(StateMixin(AutoSync(ErrorHandling(AutoLoc
             return;
         }
 
+        if (page === "items") {
+            // Run audits, if necessary
+            this._maybeRunAudits();
+        }
```
BrunoBernardino, line 135: I don't like having this here, but maybe it's fine. I initially tried to do it in `core/src/app:_unlocked`, but that meant I'd have to import `@padloc/app` inside for `auditItems`, or refactor that to not need access to `app.save()` and `app.saveVault()` (or pass those functions inside, which would be doable, but ugly), and I didn't like that.

I also tried hooking into `stateChanged()` in this element, but that triggered more frequently.

I hope you have a better suggestion @MaKleSoft ?

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,311 @@
...
+        for (const item of vault.items) {
+            if (updateOnlyItemWithId) {
+                if (item.id !== updateOnlyItemWithId) {
+                    // TODO: Remove this
+                    console.log("Skipped, not the matching item.");
+                    continue;
+                }
+            }
+
+            if (updateOnlyIfOutdated && item.lastAudited && item.lastAudited >= oneWeekAgo) {
```
BrunoBernardino, line 218: I chose to not run _any_ audit if it's been audited recently, because we already run an audit if the item changes, so it seems very unnecessary.

MaKleSoft: The idea was to also call this method even when a single item is updated. This may seem a bit inefficient but it shouldn't take up too many cycles since we'll be skipping most items anyway.

MaKleSoft: Never mind... I misread your comment.

packages/app/src/elements/app.ts
MaKleSoft: I'd also move `auditVaults` from `src/elements/audit.ts` into a method on a new mixin under `mixins/audit.ts` and add that mixin to the `<pl-app>` element. (The audit element doesn't need to have any of the actual audit logic but should only take care of displaying results).

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,311 @@
...
+    const compromisedPasswordItemIds: Set<string> = new Set();
+
+    const oneWeekAgo = sub(new Date(), { weeks: 1 });
+
+    // TODO: Remove this
+    console.log(`Running audit! ${JSON.stringify({ updateOnlyItemWithId, updateOnlyIfOutdated })}`);
+
+    for (const vault of vaults) {
+        for (const item of vault.items) {
+            if (updateOnlyItemWithId) {
```
MaKleSoft, line 210: I think you can get rid of that check and parameter, since we can always ignore items without an id.

packages/app/src/elements/audit.ts
```diff
@@ -0,0 +1,311 @@
...
+                        reusedPasswords.push({ item, vault });
+                        reusedPasswordItemIds.add(item.id);
+                    }
+
+                    auditResults.push({
+                        type: AuditResultType.ReusedPassword,
+                        fieldIndex: passwordField.fieldIndex,
+                    });
+
+                    // Also tag the first matching item as reused, once
```
MaKleSoft, line 253: Why only tag the first matching item, and not all of them? If a password is reused accross more than 2 items, we'd want to tag all of them as reused, no?

packages/app/src/elements/audit.ts
BrunoBernardino: Not running the audit when an item is updated makes it seem like a now "good" password still looks bad (with failed audits), that's why I wanted to run it there. It'll also _only_ run for that item (no matter when the last time it ran), not for everything.

Am I missing something on why that would be bad?

packages/app/src/elements/audit.ts
BrunoBernardino: This is a check to make sure we can run the audit for a single item (without refactoring the code a bunch).

packages/app/src/elements/audit.ts
BrunoBernardino: It would look strange to show the _same item_ multiple times on the same block (if it had more than one "failed" password field), that's why this check is in here. Do note it'll still show up in different blocks (weak and compromised), though.

packages/app/src/elements/app.ts
BrunoBernardino: I hear you, that makes sense.

packages/app/src/elements/audit.ts
MaKleSoft: The audit should definitely be run whenever an item is updated, the question is only how that action should be triggered. Since after saving an item the condition `item.updated  > item.lastAudited` is always true, the execution path I described would cause the audit to re-run automatically on the next "tick" (and after whatever timeout we set for the debounce). I hadn't considered the use case of "fixing" an audit warning, where you'd want to see the warning disappear immediately, but this could easily be solved in a few different ways:

- Use [throttle](https://github.com/padloc/padloc/blob/v4/packages/core/src/util.ts#L123) instead of [debounce](https://github.com/padloc/padloc/blob/v4/packages/core/src/util.ts#L101) (other than debounce, throttle executes the first invocation immediately).
- Delete the `auditResults` and `lastAudited` properties from the item when updating fields (this would make the warnings disappear immediately. Any new warnings would appear after a short delay).
- Call `runAudits` explicitly at the end of [`ItemView::save()`](https://github.com/padloc/padloc/blob/v4/packages/app/src/elements/item-view.ts#L493) (No `updateOnlyIfOutdated` parameter needed though since, like I said `item.updated  > item.lastAudited` will always be true after an item was saved).
- A combination of the above.

So there is nothing that speaks against explicitly triggering a re-audit (other than that it may be redundant), I just don't want to have to rely on "remembering" to trigger a re-audit whenever we make changes to an item, since there may be places other than the item view where that may happen.

packages/app/src/elements/audit.ts
MaKleSoft: Ah, I thought this was a boolean to filter out items with no ids at all 🤦‍♂️. Why not  `if (updateOnlyItemWithId && item.id !== updateOnlyItemWithId)` though, instead of two nested conditionals?

packages/app/src/elements/audit.ts
MaKleSoft: If an item has multiple reused password fields, both should be highlighted as such though, right? Besides, if you filter by `item.auditResults.some(res => res.type === AuditResultType.PasswordReused)` The item would still only appear once, even if it has multiple audit results of that type.

packages/app/src/elements/audit.ts
BrunoBernardino: It could be that and I had that initially, but it felt harder to read than this. I can change it if you want.

packages/app/src/elements/audit.ts
BrunoBernardino: All fields will be highlighted in the item view, not in the password audit. The password audit page shows the items in each bucket/box, so it doesn't matter if they have one or 10 fields failing the audit. Does that make sense, or are you thinking of doing something else?

We're not currently filtering like you mentioned, though, just adding items per bucket/box as the audit runs, so it's one less loop over everything.

packages/app/src/elements/audit.ts
BrunoBernardino: Gotcha, I'll try to implement it a bit differently, then (removing the warnings would help even if any new warning eventually shows up again).

BrunoBernardino:
@MaKleSoft this should be now ready for a review. Your `item.updated > item.lastAudited` recommendation threw me off for a while because when we set `lastAudited`, it also updates `item.updated`, and there isn't an easy way to just "unset" it, so I simply went with setting `lastAudited` way in the past, to make it run. It's worked really well in my testing.

packages/app/src/elements/items-list.ts
```diff
@@ -657,6 +658,11 @@ export class ItemsList extends StateMixin(LitElement) {
                   icon: "time",
                   text: $l("You don't have any recently used items!"),
               }
+            : audit
+            ? {
+                  icon: "shield-check",
+                  text: noItemsTextForAudit(audit),
```
BrunoBernardino, line 664: I didn't see this pattern anywhere, but I needed a readable/easy/simple way to set different texts here, and didn't want to `: ?` a few more.

packages/app/src/elements/menu.ts
```diff
@@ -558,6 +558,16 @@ export class Menu extends Routing(StateMixin(LitElement)) {
...
                     </div>
 
+                    <div
+                        class="menu-item"
+                        @click=${() => this._goTo("audit")}
+                        aria-selected=${this.selected === "audit"}
+                    >
+                        <pl-icon icon="shield-check"></pl-icon>
+
+                        <div class="stretch">${$l("Security Audit")}</div>
```
BrunoBernardino, line 568: I like this name, but wouldn't oppose `Security Check` (I've seen your comment). If you want me to change it, let me know if I should change the file names/filter, etc.

packages/app/src/mixins/audit.ts
```diff
@@ -0,0 +1,195 @@
...
+            const compromisedPasswords: ListItem[] = [];
+
+            // Don't try to run if the app has locked
+            if (app.state.locked) {
+                return {
+                    reusedPasswords,
+                    weakPasswords,
+                    compromisedPasswords,
+                };
+            }
```
BrunoBernardino, line 35: This was causing the errors in Cypress. I'm happy it was reporting a real/useful error!

packages/app/src/mixins/audit.ts
```diff
@@ -0,0 +1,195 @@
...
+                { item: VaultItem; vault: Vault; passwordField: { field: Field; fieldIndex: number } }
+            >();
+            const reusedPasswordItemIds: Set<string> = new Set();
+            const weakPasswordItemIds: Set<string> = new Set();
+            const compromisedPasswordItemIds: Set<string> = new Set();
+
+            const oneWeekAgo = sub(new Date(), { weeks: 1 });
+
+            // TODO: Remove this
+            console.log(`Running audit! ${JSON.stringify({ updateOnlyItemWithId, updateOnlyIfOutdated })}`);
```
BrunoBernardino, line 48: Will remove this before merging, just kept it in so you could see how frequently it ran.

packages/app/src/mixins/audit.ts
```diff
@@ -0,0 +1,195 @@
...
+
+            for (const vault of vaults) {
+                for (const item of vault.items) {
+                    if (updateOnlyItemWithId) {
+                        if (item.id !== updateOnlyItemWithId) {
+                            // TODO: Remove this
+                            console.log("Skipped, not the matching item.");
```
BrunoBernardino, line 55: Same as above.

packages/app/src/mixins/audit.ts
```diff
@@ -0,0 +1,195 @@
...
+                            continue;
+                        }
+                    }
+
+                    if (updateOnlyIfOutdated && item.lastAudited && item.lastAudited >= oneWeekAgo) {
+                        // TODO: Remove this
+                        console.log("Skipped, already up to date.");
```
BrunoBernardino, line 62: Same as above.

packages/app/src/elements/item-view.ts
```diff
@@ -508,11 +513,15 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
         if (!this._nameInput.reportValidity()) {
             return;
         }
+        const oneMonthAgo = sub(new Date(), { months: 1 });
         app.updateItem(this._item!, {
```
BrunoBernardino, line 517: Is there a way to _unset_ something here? I couldn't find it, so I just set `lastAudited` to a date far in the past, which will also work.

packages/app/src/elements/menu.ts
MaKleSoft: Another idea: `Security Scanner`?

BrunoBernardino: I still prefer `Security Audit` or `Security Check`.

packages/app/src/elements/item-view.ts
MaKleSoft: Have you tried `{ lastAudited: undefined }`? Otherwise, `{ lastAudited: new Date(0) }` should also work.

packages/app/src/elements/item-view.ts
```diff
@@ -508,11 +513,15 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
         if (!this._nameInput.reportValidity()) {
             return;
         }
+        const oneMonthAgo = sub(new Date(), { months: 1 });
```
MaKleSoft, line 516: Going back a month seems kind of arbitrary. Why not use the beginning of the Unix epoch (`new Date(0)`)?

packages/app/src/elements/items-list.ts
MaKleSoft: This pattern is fine 👍

packages/app/src/elements/app.ts
```diff
@@ -31,10 +31,10 @@ import "./rich-content";
 import { alertDisabledFeature, displayProvisioning, getDefaultStatusLabel } from "../lib/provisioning";
 import { ItemsView } from "./items";
 import { wait, throttle } from "@padloc/core/src/util";
-import { AuditMixin } from "../mixins/audit";
+import { auditVaults } from "../lib/audit";
```
MaKleSoft, line 34: I realised that the way we're doing things now, there is really no value in implementing this as a mixin as opposed to just exporting a function.

packages/app/src/elements/audit.ts
```diff
@@ -54,8 +36,7 @@ export class Audit extends AuditMixin(StateMixin(Routing(View))) {
     ];
 
     render() {
-        const { _reusedPasswords, _weakPasswords, _compromisedPasswords } = this;
-
+        let items = app.auditedItems;
```
MaKleSoft, line 39: There is no reason to re-run the audit every time we open this page since the information we want is already stored in the items themselves. This is more efficient and also guarantees that this view is always updated, even while open.

packages/app/src/elements/audit.ts
```diff
@@ -91,46 +69,45 @@ export class Audit extends AuditMixin(StateMixin(Routing(View))) {
...
-        const { reusedPasswords, weakPasswords, compromisedPasswords } = await this.auditVaults(vaults);
-
-        // This makes the UI update
-        this._reusedPasswords = reusedPasswords;
-        this._weakPasswords = weakPasswords;
-        this._compromisedPasswords = compromisedPasswords;
-
-        this._countElements.forEach((countElement) => {
-            animateElement(countElement, { animation: "bounce" });
-        });
```
MaKleSoft: See my comment above for why I removed this. You can add the animation back in if you want - the best place for it that would then be in `handleRoute`, though.

packages/app/src/elements/audit.ts
```diff
@@ -91,46 +69,45 @@ export class Audit extends AuditMixin(StateMixin(Routing(View))) {
...
-    }
-
-    private _renderSection(listItems: ListItem[], label: string, routeParams: any) {
+    private _renderSection(listItems: { item: VaultItem; vault: Vault }[], type: AuditResultType) {
         return html`
-            <section class="box count" ?hidden=${listItems.length === 0}>
```
MaKleSoft: I thought it was better to always display all sections, even if there are no results in them (yet).

packages/app/src/elements/audit.ts
```diff
@@ -91,46 +69,45 @@ export class Audit extends AuditMixin(StateMixin(Routing(View))) {
...
-                    <div></div>
-                    <div>${label}</div>
-                    <div class="tiny bold tag">${listItems.length}</div>
+                    <pl-icon icon="${iconForAudit(type)}" class="left-margined"></pl-icon>
+                    <div>${titleTextForAudit(type)}</div>
+                    <div class="bold subtle">${listItems.length}</div>
                     <div class="stretch"></div>
+                    <pl-button class="subtle skinny transparent half-margined">
+                        <pl-icon icon="info-round"></pl-icon>
+                    </pl-button>
```
MaKleSoft, line 82: TODO: Add a popover with a short description of what this section means.

packages/app/src/elements/audit.ts
```diff
@@ -91,46 +69,45 @@ export class Audit extends AuditMixin(StateMixin(Routing(View))) {
...
+                              ${listItems.slice(0, 5).map(
+                                  (listItem) => html`
+                                      <div
+                                          class="list-item hover click"
+                                          @click=${() => this.go(`items/${listItem.item.id}`)}
+                                      >
+                                          <pl-vault-item-list-item
+                                              .item=${listItem.item}
+                                              .vault=${listItem.vault}
+                                              class="none-interactive"
```
MaKleSoft, line 96: I thought it was better to deactivate interaction with fields and such here.

packages/app/src/lib/audit.ts
```diff
@@ -0,0 +1,241 @@
...
+        case AuditResultType.WeakPassword:
+            return $l("Weak Passwords");
+        case AuditResultType.ReusedPassword:
+            return $l("Reused Passwords");
+        case AuditResultType.CompromisedPassword:
+            return $l("Compromised Passwords");
+        default:
+            return $l("Insecure");
+    }
+}
```
MaKleSoft, line 228: Moved these two functions here along with the `auditVaults` function so they can easily be imported and used throughout components.

MaKleSoft: Also, for this kind of thing I prefer switch/case, simply because it looks nicer :)

packages/core/src/app.ts
```diff
@@ -367,6 +367,18 @@ export class App {
...
+        let items: { item: VaultItem; vault: Vault }[] = [];
+        for (const vault of this.vaults) {
+            for (const item of vault.items) {
+                if (item.auditResults?.length) {
+                    items.push({ item, vault });
+                }
+            }
+        }
+        return items;
+    }
```
MaKleSoft, line 380: Just added for convenience.

packages/app/src/lib/audit.ts
```diff
@@ -0,0 +1,241 @@
...
+        case AuditResultType.WeakPassword:
+            return "weak";
+        case AuditResultType.ReusedPassword:
+            return "reused";
+        case AuditResultType.CompromisedPassword:
+            return "compromised";
+        default:
+            return "shield-x";
+    }
+}
```
MaKleSoft, line 241: I thought it would be nice to have distinct icons for each audit type. Looks really cool in the item view!

packages/app/src/elements/icon.ts
```diff
@@ -558,13 +558,25 @@ export class PlIcon extends LitElement {
                 content: "\\f2dc";
             }
 
-            :host([icon="shield-check"]) > div::before {
+            :host([icon="audit-pass"]) > div::before {
                 content: "\\f2f7";
             }
 
-            :host([icon="shield-xmark"]) > div::before {
+            :host([icon="audit-fail"]) > div::before {
```
MaKleSoft, line 565: Icon names in this component are usually chosen by their meaning in the context in the app, as opposed to the actual thing they are showing.

MaKleSoft:
Looks really good so far! I added a couple of tweaks of my own - hope that's ok! I also noticed that all audits are run for empty password fields as well - I think those should be skipped.

packages/app/src/elements/menu.ts
MaKleSoft: Ok, let's stick with `Security Audit` then 😊

BrunoBernardino:
@MaKleSoft I just fixed the date thing (setting `undefined` worked), the empty passwords check (empty password fields are ignored now), and a new bug I had found while testing: the first item in a vault would never get a "reused password" audit fail, if it didn't repeat the password in itself (because other items wouldn't make a change to it), or if it was skipped (when updating another item, for example). Basically the simplest solution was to just check the number of times a password hash is used and if > 1, tag it for audit if the item is being audited.

I'm comfortable with how this looks and works now, thanks for the UI polish with the icons, explanations, etc.

MaKleSoft (approved):
I'd say this is ready to be merged. Great work!

PR #415, opened 2022-03-22
Lock down tauri apis / whitelist dependencies

MaKleSoft:
Upgraded tauri to the latest release candidate and updated the tauri config to disable all tauri/rust apis, since we're currently not using any of them (we did use `shell` but it was only used to open the help page from the application menu and I decided that wasn't a good enough reason to whitelist that api).


packages/tauri/package.json
```diff
@@ -16,7 +17,8 @@
     "license": "GPL-3.0",
     "devDependencies": {
         "@iarna/toml": "2.2.5",
-        "@tauri-apps/cli": "1.0.0-beta.10",
+        "@tauri-apps/cli": "1.0.0-rc.7",
+        "@tauri-apps/tauricon": "github:tauri-apps/tauricon",
```
MaKleSoft, line 21: `tauri icon` appears to be a separate packages now.

packages/tauri/src-tauri/src/main.rs
```diff
@@ -20,21 +19,7 @@ fn main() {
...
-        "get_help" => {
-          shell::open(
-            "https://padloc.app/help/".to_string(),
-            None,
-          )
-          .unwrap();
-        }
-        _ => {}
-      }
-    })
```
MaKleSoft, line 37: Couldn't find a different way to open a url and whitelisting the `shell` api just for that didn't seem justified.

packages/tauri/src-tauri/tauri.conf.json
```diff
@@ -63,7 +63,7 @@
             "pubkey": "dW50cnVzdGVkIGNvbW1lbnQ6IG1pbmlzaWduIHB1YmxpYyBrZXk6IDM5NDMxQ0I3MTA3NDU3RjIKUldUeVYzUVF0eHhET1ZpNUZDYWY3Zlk1aUN1ZkpXcU5wS0pSTERSay91ODBpWlBIcTViNk11RHgK"
         },
         "allowlist": {
-            "all": true
+            "all": false
```
MaKleSoft, line 66: Lock down all tauri apis for now since we're not using any of them.

packages/tauri/src-tauri/tauri.conf.json
```diff
@@ -75,7 +75,7 @@
             }
         ],
         "security": {
-            "csp": "default-src blob: data: filesystem: ws: wss: http: https: tauri: 'unsafe-eval' 'unsafe-inline' 'self'; style-src 'self' 'unsafe-inline'; object-src 'self' blob:; frame-src 'self'; img-src 'self' blob: data: https:;"
+            "csp": null
```
MaKleSoft, line 78: Apparently tauri now automatically adds nonces to inline `<script>` and `<style>` tags to white-list them explicitly, which conflicts with how lit injects style sheets... So I had to disable it altogether. Hopefully in the future they'll offer a way to opt out of the nonce stuff but including in a csp tag in the `index.html` directly works just as well I think (not sure if there are any differences security-wise).

packages/tauri/package.json
BrunoBernardino: I would recommend we pin this to a specific tag/version, but since they have none yet, I think this is fine.

BrunoBernardino (approved):
I've read the code and I haven't pulled this down to test further. I like the changes, and they seem sensible and clear. Nice work!

PR #417, opened 2022-03-31
Implement Stripe Integration for v4

MaKleSoft:
This adds a proper implementation of `StripeProvisioner`, which builds on `BasicProvisioner` and derives provisioning profiles from Stripe subscriptions. There's been some refactoring of the core provisioning logic as well so this may break compatibility with previous v4 revisions.

Since I was doing a lot of explorative programming the individual commits don't really have any rhyme or reason to them so we'll probably just squash everything before merging.


MaKleSoft:
@BrunoBernardino This is a pretty huge PR which touches a lot of different places, so I don't expect you to do a full review. Maybe just skim through it and see if something stands out to you 😊

.github/workflows/build-tauri.yml
```diff
@@ -67,45 +67,35 @@ jobs:
             - name: Archive AppImage
               uses: actions/upload-artifact@v2
               if: matrix.platform == 'ubuntu-latest'
-              env:
-                  TARGET_DIR: ${{ github.event.inputs.environment == 'Production' && 'release' || 'debug' }}
               with:
                   name: padloc-linux-${{ github.sha }}-unsigned.AppImage
-                  path: packages/tauri/src-tauri/target/${{ env.TARGET_DIR }}/bundle/linux/padloc*.AppImage
+                  path: packages/tauri/src-tauri/target/release/bundle/appimage/padloc*.AppImage
```
MaKleSoft, line 72: I just cherry-picked those changes from your open PR to fix the tauri build.

BrunoBernardino:
I skimmed through it and it all looked clear, safe, and good. Nothing jumped out as being confusing or potentially dangerous/problematic. Nice work!

PR #425, opened 2022-04-08
Add timingSafeEqual to CryptoProvider as better alternative for @padloc/core/encoding/equalCT

MaKleSoft:
Replaces [@padloc/core/util/equalCT](https://github.com/padloc/padloc/blob/v4/packages/core/src/encoding.ts#L507) (which is not really constant-time) with true timing-safe comparison method as part of [CryptoProvider](https://github.com/padloc/padloc/blob/v4/packages/core/src/crypto.ts#L148).

### DIFF
--- a/packages/core/src/server.ts
+++ b/packages/core/src/server.ts
@@ -1,4 +1,4 @@
-import { equalCT, Serializable, stringToBase64 } from "./encoding";
+import { Serializable, stringToBase64 } from "./encoding";
 import {
     API,
     StartCreateSessionParams,
@@ -51,7 +51,7 @@ import {
     Messenger,
 } from "./messenger";
 import { Server as SRPServer, SRPSession } from "./srp";
-import { DeviceInfo } from "./platform";
+import { DeviceInfo, getCryptoProvider } from "./platform";
 import { getIdFromEmail, uuid } from "./util";
 import { loadLanguage } from "@padloc/locale/src/translate";
 import { Logger, VoidLogger } from "./logging";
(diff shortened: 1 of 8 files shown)


packages/core/src/server.ts
```diff
@@ -599,7 +599,7 @@ export class Controller extends API {
         // Add device to trusted devices
         if (
             this.context.device &&
-            !auth.trustedDevices.some(({ id }) => equalCT(id, this.context.device!.id)) &&
+            !auth.trustedDevices.some(({ id }) => id === this.context.device!.id) &&
```
MaKleSoft, line 602: We don't actually need a constant-time comparison here, since we're already authenticated and both values are (in theory) already known to the caller.

packages/core/src/stub-crypto-provider.ts
```diff
@@ -142,4 +142,14 @@ export class StubCryptoProvider implements CryptoProvider {
...
+
+    async timingSafeEqual(a: Uint8Array, b: Uint8Array): Promise<boolean> {
+        let match = true;
+
+        for (let i = 0; i < a.length; i++) {
+            match = match && a[i] === b[i];
+        }
+
+        return a.length === b.length && match;
+    }
```
MaKleSoft, line 154: This is **NOT** timing safe, but merely a stub implementation for test purposes.

BrunoBernardino (approved):
I haven't pulled this down but read the code carefully and it looks good! Nice job.

PR #429, opened 2022-04-14
SCIM v1

BrunoBernardino:
Basic thoughts and structure.

Not shown here yet is an option to show, per org, in their settings, this SCIM URL with their secret token, like `http://localhost:5000/Users?token=asdrtyghj&org=e0bb91b4-2b35-4ba7-ba60-f4ac8470e7a3` and you can test it locally by sending a POST request like:

```
curl --request POST \
  --url 'http://localhost:5000/Users?token=asdrtyghj&org=e0bb91b4-2b35-4ba7-ba60-f4ac8470e7a3' \
  --header 'Content-Type: application/scim+json' \
  --data '{
	"schemas": ["1", "2"],
	"externalId": "some-id",
	"userName": "bruno",
	"active": true,
	"meta": {
		"resourceType": "User"
	},
	"name": {
		"formatted": "Bruno SCIM",
		"familyName": "SCIM",
		"givenName": "Bruno"
	},
	"email": "bruno.scim@example.com"
}'
```


packages/core/src/provisioning.ts
```diff
@@ -1,10 +1,17 @@
+import { createServer, IncomingMessage, ServerResponse } from "http";
+import { Config, ConfigParam } from "./config";
 import { Account, AccountID } from "./account";
 import { AsSerializable, Serializable } from "./encoding";
 import { Err, ErrorCode } from "./error";
 import { Org, OrgID, OrgInfo } from "./org";
 import { Storable, Storage } from "./storage";
 import { getIdFromEmail } from "./util";
 
+export class BasicProvisionerConfig extends Config {
```
BrunoBernardino, line 10: @MaKleSoft I was going to implement SCIM in `BasicProvisioning`, but I'm running into some problems like:

1. `core` needs to run on browser (I "mocked" `http` via `webpack`)
2. The methods I find for inviting/adding members require the app to be unlocked, which can't be done via server like this

I initially looked into hooking up to the `api.ts` provisioning in `server`, but it seems like that's not used? (only `basic` and `stripe` are allowed) I could add this to the `stripe` provider, though, and just allow only some provisioners to support it, which would make sense.

packages/core/src/provisioning.ts
```diff
@@ -351,4 +378,154 @@ export class BasicProvisioner implements Provisioner {
...
+        const url = new URL(`http://localhost${httpReq.url || ""}`);
+        const secretToken = url.searchParams.get("token") || "";
+
+        // TODO: find secret token in the DB
+        if (secretToken === "asdrtyghj") {
+            return {
+                accountId: "472478c5-17e8-4ed5-8f51-05a9c5deaebb",
+                orgId: "e0bb91b4-2b35-4ba7-ba60-f4ac8470e7a3",
+            };
+        }
```
BrunoBernardino, line 392: This is just some local test data for me, but this is what I'd envision we return from a token.

packages/core/src/provisioning.ts
```diff
@@ -351,4 +378,154 @@ export class BasicProvisioner implements Provisioner {
...
+            const organization = provisioning.orgs.find((org) => org.id === orgId);
+
+            if (!organization) {
+                throw new Error("Organization not found");
+            }
+
+            // TODO: create user
+            // const invite = new Invite(email, purpose);
+            // await invite.initialize(org, this.account!);
+            // await org.addOrUpdateMember(invite.invitee);
```
BrunoBernardino, line 459: None of these would work here.

packages/core/src/provisioning.ts
```diff
@@ -351,4 +378,154 @@ export class BasicProvisioner implements Provisioner {
 
         return prov;
     }
+
+    getAccountIdFromScimRequest(httpReq: IncomingMessage) {
+        const url = new URL(`http://localhost${httpReq.url || ""}`);
+        const secretToken = url.searchParams.get("token") || "";
+
+        // TODO: find secret token in the DB
```
MaKleSoft, line 386: I think it's ok to include the organisation id in the url, which will make it easier to find the `Org` and `OrgProvisioning` objects in the database. I.e.:

```sh
const orgId = url.searchParams.get("orgId");
const orgProvisioning = await this.storage.get(Org, orgId);
```

packages/core/src/provisioning.ts
```diff
@@ -351,4 +378,154 @@ export class BasicProvisioner implements Provisioner {
...
+
+        const validationError = this.validateScimUser(newUser);
+        if (validationError) {
+            httpRes.statusCode = 400;
+            httpRes.end(validationError);
+            return;
+        }
+
+        try {
+            const provisioning = await this.getProvisioning({ email: newUser.email, accountId });
```
MaKleSoft, line 448: FYI: You don't need the `accountId` here, and in fact there is no guarantee that there even is one at this point, so just the email is fine.

packages/core/src/provisioning.ts
MaKleSoft: I'll probably add a `members` property to `OrgProvisioning` where you can add these.

packages/core/src/provisioning.ts
```diff
@@ -351,4 +378,154 @@ export class BasicProvisioner implements Provisioner {
...
+
+    async startScimServer() {
+        // TODO: Remove this
+        console.log("======== provisioning.startScimServer");
+        const server = createServer((req, res) => this.handleScimRequest(req, res));
+        server.listen(this.config!.port);
+    }
+}
+
+// TODO: Maybe it makes sense to move this from the http server transport to core's transport as readHttpBody?
```
MaKleSoft, line 509: As discussed, simply move the class/module into the `@padloc/server` package and then you can simply import the function from `@padloc/server/src/transport/http`.

packages/core/src/provisioning.ts
MaKleSoft: As discussed, this doesn't need to live in core. Let's just make the whole thing it's own class in `@padloc/server` for now.

BrunoBernardino:
I already moved it to its own class. I'll keep working on it in a little bit.

packages/server/src/scim.ts
```diff
@@ -0,0 +1,126 @@
...
+            if (!orgProvisioning) {
+                throw new Error("Organization not found");
+            }
+
+            const newProvisioningMember = new MemberProvisioning();
+            newProvisioningMember.email = newUser.email;
+
+            orgProvisioning.members.push(newProvisioningMember);
+
+            // TODO: Save?
```
BrunoBernardino, line 75: @MaKleSoft do I need to expose something to allow this, or is there something I'm missing?

MaKleSoft: You'll need to pass the storage reference to the class, just like with `BasicProvisioner`. Then you can do `this.storage.save(orgProvisioning)`.

packages/server/src/scim.ts
```diff
@@ -0,0 +1,126 @@
...
+
+            // TODO: create auth provisioning
+            // TODO: create invite provisioning
```
BrunoBernardino, line 78: @MaKleSoft I believe these are superseded by the `MemberProvisioning` above, correct?

packages/server/src/scim.ts
```diff
@@ -0,0 +1,126 @@
...
+    async init() {
+        await this._startScimServer();
+    }
+
+    private _getDataFromScimRequest(httpReq: IncomingMessage) {
+        const url = new URL(`http://localhost${httpReq.url || ""}`);
+        const secretToken = url.searchParams.get("token") || "";
+        const orgId = url.searchParams.get("org") || "";
+
+        // TODO: find account/org based on token
```
BrunoBernardino, line 21: @MaKleSoft any idea of where should this (token) be? I'm thinking of automatically generating + adding it to the org when it's created, but how would it be searchable here?

packages/server/src/scim.ts
MaKleSoft: The "invite provisioning" yes, but you'll still have to get or create the AccountProvisioning object for that email and make sure the `ProvisioningStatus` is set to `ProvisioningStatus.Active`.

MaKleSoft: In that regard it kind of overlaps with `BasicProvisioner`, which is why I think it should be a subclass/extension.

MaKleSoft: In that regard it kind of overlaps with `BasicProvisioner`, which is why I think it should be a subclass/extension.

packages/server/src/scim.ts
MaKleSoft: You can find the `OrgProvisioning` and `Org` objects based on `orgId`:

```sh
const orgProv = await this._getOrCreateOrgProvisioning(orgId);
const org = await this.storage.get(Org, orgId);
```

And the `AccountProvisioning`:

```sh
const accountProv = this._getOrCreateAccountProvisioning({ email });
```

To use `_getOrCreateOrgProvisioning` and `_getOrCreateAccountProvisioning` you'll have to subclass `BasicProvisioner` and make those methods `protected` instead of `private`

MaKleSoft: You can find the `OrgProvisioning` and `Org` objects based on `orgId`:

```sh
const orgProv = await this._getOrCreateOrgProvisioning(orgId);
const org = await this.storage.get(Org, orgId);
```

And the `AccountProvisioning`:

```sh
const accountProv = this._getOrCreateAccountProvisioning({ email });
```

To use `_getOrCreateOrgProvisioning` and `_getOrCreateAccountProvisioning` you'll have to subclass `BasicProvisioner` and make those methods `protected` instead of `private`

packages/server/src/scim.ts
MaKleSoft: So specifically:

```sh
const accountProv = await this._getOrCreateAccountProvisioning({ email });
accountProv.status = newUser.active ? ProvisioningStatus.Active : ProvisioningStatus.Suspended;
```

MaKleSoft: 👆This is the part that clashes with `StripeProvisioner`, basically.

packages/server/src/scim.ts
MaKleSoft: So the secret token wouldn't be used for the lookup but just as an additional check. You can store it on the `OrgProvisioning` object for now.

```ts
if (orgProv.scimSecret !== secret) {
     throw new Err(ErrorCode.AUTHENTICATION_FAILED, "...")
}
```

packages/server/src/provisioning/scim.ts
```diff
@@ -0,0 +1,175 @@
...
+            const newProvisioningMember = new MemberProvisioning();
+            newProvisioningMember.email = newUser.email;
+
+            provisioning.account.status = newUser.active ? ProvisioningStatus.Active : ProvisioningStatus.Suspended;
+
+            orgProvisioning.members.push(newProvisioningMember);
+
+            // TODO: Save
+            // await this.storage.save(orgProvisioning);
+            // await this.storage.save(provisioning);
```
BrunoBernardino, line 127: @MaKleSoft I suppose this would be enough (didn't want to save yet as then I'll have to delete data on every test)

MaKleSoft:
Ok, so I've thought about it for a while and I've come to the conclusion that we're dealing with two distinct concerns here:

**Directory Provisioning**: Provisioning / whitelisting emails/accounts based on the users in a directory. This is only relevant in an on-premise/enterprise setting

**Directory Sync**: Adding/removing organisation members, adding/removing groups and assigning group memberships. This can be used both in the enterprise and SaaS context.

Since both of these may need to exist in parallel and draw data from the same source(s), we need a **Directory Provider** that both of these can hook into. **Directory Providers** and **Directory Subscribers** interact with a simple interface:

```ts
export interface DirectorySubscriber {
    userCreated(user: DirectoryUser, orgId?: OrgID): Promise<void>;
    userUpdated(user: DirectoryUser, orgId?: OrgID): Promise<void>;
    userDeleted(user: DirectoryUser, orgId?: OrgID): Promise<void>;

    groupCreated(group: DirectoryGroup, orgId: OrgID): Promise<void>;
    groupUpdated(group: DirectoryGroup, orgId: OrgID): Promise<void>;
    groupDeleted(group: DirectoryGroup, orgId: OrgID): Promise<void>;
}

export interface DirectoryProvider {
    subscribe(sub: DirectorySubscriber): void;
}
```

An example for a **Directory Provider** would be [`ScimServer`](https://github.com/padloc/padloc/pull/429/files#diff-173a31fa3c4302dcbd000042d1a739ef9c4f0657d24670172e98825bebbf0b06R31). Examples for a **Directory Subscriber** would be [`DirectoryProvisioner`](https://github.com/padloc/padloc/pull/429/files#diff-c1796e7fb5357168d4beb2daa0fef71408eb5b91d81302da24970bcf0e376540) and [`DirectorySync`](https://github.com/padloc/padloc/pull/429/files#diff-e48f537ea43df65f2f733b68e321624fd34237257fab618884ade5c6a77ef438R30).

packages/core/src/org.ts
```diff
@@ -31,16 +31,27 @@ export enum OrgRole {
      * Member information (like public key and email address) of suspended members
      * are considered unverified, and need to be updated and verified via a
      * membership confirmation [[Invite]].
+     * @deprecated Use `OrgMemberStatus.Suspended` instead
      */
     Suspended,
 }
 
+export enum OrgMemberStatus {
+    Provisioned = "provisioned",
```
MaKleSoft, line 40: This seemed like a more ergonomic way to "provision" new organisation members, rather than storing them on the `OrgProvisioning` object.

packages/core/src/org.ts
```diff
@@ -31,16 +31,27 @@ export enum OrgRole {
...
  * Represents an [[Account]]s membership to an [[Org]]
  */
 export class OrgMember extends Serializable {
+    get id() {
+        return this.accountId;
+    }
+
     /** id of the corresponding [[Account]] */
-    id: AccountID = "";
+    accountId?: AccountID = undefined;
```
MaKleSoft, line 54: Organization members are no longer guaranteed to have a corresponding account (yet), so we'll use the email address as a primary way of reference now.

packages/core/src/org.ts
```diff
@@ -50,15 +61,15 @@ export class OrgMember extends Serializable {
 
     /** public key of the corresponding [[Account]] */
     @AsBytes()
-    publicKey!: RSAPublicKey;
+    publicKey?: RSAPublicKey;
```
MaKleSoft, line 64: These won't be defined until the Member is fully onboarded (i.e. an Account has been created and the invite flow has been completed).

packages/core/src/org.ts
```diff
@@ -238,45 +281,61 @@ export class Org extends SharedContainer implements Storable {
     @AsSerializable(Invite)
     invites: Invite[] = [];
 
+    @AsSerializable(OrgDirectorySettings)
+    directory: OrgDirectorySettings = new OrgDirectorySettings();
```
MaKleSoft, line 285: This is where users will be able to store their directory sync preferences.

packages/core/src/app.ts
```diff
@@ -1889,6 +1890,29 @@ export class App {
         );
     }
 
+    async autoHandleInvites(): Promise<void> {
```
MaKleSoft, line 1893: This is where the magic happens. Basically whenever an organisation owner unlocks their app, invites will be sent out automatically to all provisioned members.

packages/core/src/directory.ts
```diff
@@ -0,0 +1,67 @@
...
+        const org = (orgId && (await this.storage.get(Org, orgId))) || null;
+        if (org && org.directory.syncProvider === "scim" && org.directory.syncMembers) {
+            org.members.push(
+                new OrgMember({
+                    email: user.email,
+                    status: OrgMemberStatus.Provisioned,
+                    updated: new Date(),
+                })
+            );
+            org.revision = await uuid();
```
MaKleSoft, line 47: FYI: It's important to update the `revision` property every time you update an org object, so the client knows to pull the changes.

packages/app/src/elements/org-settings.ts
```diff
@@ -116,6 +119,57 @@ export class OrgSettingsView extends Routing(StateMixin(LitElement)) {
...
+
+        await app.updateOrg(this._org!.id, async (org) => {
+            org.directory.syncProvider = "scim";
+            org.directory.syncGroups = true;
+            org.directory.syncMembers = true;
+            org.directory.scim = new ScimSettings();
+            org.directory.scim.secret = stringToBytes(await uuid());
+        });
+
+        await app.synchronize();
```
BrunoBernardino, line 144: @MaKleSoft this isn't working for me, do you know why? The org keeps the previous/default `directory` values.

packages/app/src/elements/org-settings.ts
```diff
@@ -116,6 +119,57 @@ export class OrgSettingsView extends Routing(StateMixin(LitElement)) {
...
+        }
+
+        // TODO: In the future, ask if only groups/members should be synchronized
+
+        await app.updateOrg(this._org!.id, async (org) => {
+            org.directory.syncProvider = "scim";
+            org.directory.syncGroups = true;
+            org.directory.syncMembers = true;
+            org.directory.scim = new ScimSettings();
+            org.directory.scim.secret = stringToBytes(await uuid());
```
MaKleSoft, line 141: Better: `await getCryptoProvider().randomBytes(16);`

packages/app/src/elements/org-settings.ts
MaKleSoft: You don't need `synchronise`, since `app.updateOrg` already takes care of that. What you'll need to do is to update the `updateOrg` endpoint  here: https://github.com/padloc/padloc/blob/v4/packages/core/src/server.ts#L915

packages/core/src/server.ts
```diff
@@ -1014,14 +1028,22 @@ export class Controller extends API {
...
+            directory,
         });
 
-        if (org.owner !== owner) {
-            await this.provisioner.orgOwnerChanged(
-                org,
-                org.getMember({ id: org.owner })!,
-                org.getMember({ id: owner })!
-            );
+        if (org.directory.syncProvider === "scim") {
```
BrunoBernardino, line 1034: @MaKleSoft I initially implemented this in the frontend, but `directory` wasn't seeing `directory.scim` (even though it was properly being sent as serialized in the HTTP request). After a lot of struggling, I decided it wouldn't be bad to implement this in the backend, as if it was ever updated in some other way, it would not have to be a concern of the client, but it still doesn't work, somehow. The other settings work and get saved, my logging below sees the secret, but it's just not saved.

packages/core/src/org.ts
```diff
@@ -163,10 +163,12 @@ export class ScimSettings extends Serializable {
 }
 
 export class OrgDirectorySettings extends Serializable {
-    scim?: ScimSettings;
     syncProvider: "scim" | "none" = "none";
     syncGroups: boolean = true;
     syncMembers: boolean = true;
+
+    @AsSerializable(ScimSettings)
```
MaKleSoft, line 170: @BrunoBernardino This is what was missing I think. My bad, sorry!

BrunoBernardino: :facepalm: I was looking for something like this and missed it! Thanks, it worked!

packages/core/src/directory.ts
```diff
@@ -0,0 +1,91 @@
...
+            org.updated = new Date();
+
+            await this.storage.save(org);
+
+            // TODO: Remove this
+            console.log("---- core/src/directory. userCreated");
+            console.log(org.members);
```
BrunoBernardino, line 54: @MaKleSoft now this seems to be saving, but when I load up the org again, it doesn't have the new org member (I also see no invites or anything that should've been auto-created), and no errors. I haven't had a chance to debug that, though.

MaKleSoft: I think I know what the problem is. Let me take a closer look.

MaKleSoft: Ok, so updating the `revision` property on the `Org` isn't enough, you actually have to update the `orgs` property for all member accounts as well. Actually that's what [`updateMetaData`](https://github.com/padloc/padloc/blob/master/packages/core/src/server.ts#L1352) in the `server.ts` module is for. In fact, I'm thinking it may be better to pass the `Server` instance to `DirectorySync` instead of just the storage. Then you can do:

```ts
this.server.updateMetaData(org);
this.server.storage.save(org);
```

MaKleSoft: > Ok, so updating the revision property on the Org isn't enough, you actually have to update the orgs property for all member accounts as well.

To understand why this is necessary, take a look [here](https://github.com/padloc/padloc/blob/acbc931b510df53dfaaf0b96fb27cad8463e99f1/packages/core/src/app.ts#L1478) and [here](https://github.com/padloc/padloc/blob/acbc931b510df53dfaaf0b96fb27cad8463e99f1/packages/core/src/app.ts#L1489).

BrunoBernardino: I'm afraid this wasn't enough. I've investigated this deeper and I believe the problem comes from [here](https://github.com/padloc/padloc/blob/feature/scim-v1/packages/core/src/server.ts#L1999-L2001), because there is no `account.orgs` to save the data into and it doesn't get saved?

MaKleSoft: @BrunoBernardino Can you give me an example SCIM payload so I can test this locally?

BrunoBernardino: Sure, check the initial comment, it's got one, though you'll need to change the URL to the one you get in your org's settings after enabling SCIM.

If it helps, I boot the app with `PL_PROVISIONING_BACKEND=directory PL_DIRECTORY_PROVIDERS=scim PL_DIRECTORY_SCIM_PORT=5000 npm run dev`

packages/core/src/server.ts
```diff
@@ -2023,7 +2023,7 @@ export class Server {
         await Promise.all(promises);
 
         org.vaults = org.vaults.filter((v) => !deletedVaults.has(v.id));
-        org.members = org.members.filter((m) => m.accountId && !deletedMembers.has(m.accountId));
+        org.members = org.members.filter((m) => !m.accountId || !deletedMembers.has(m.accountId));
```
MaKleSoft, line 2026: @BrunoBernardino This is where the new member was being filtered out (made a mistake here when refactoring the `OrgMember` model).

packages/core/src/org.ts
```diff
@@ -95,9 +95,10 @@ export class OrgMember extends Serializable {
         orgSignature,
         role,
         updated,
+        status,
     }: Partial<OrgMember> = {}) {
         super();
-        Object.assign(this, { accountId, name, email, publicKey, signature, orgSignature, updated });
+        Object.assign(this, { accountId, name, email, publicKey, signature, orgSignature, updated, status });
```
MaKleSoft, line 101: I noticed the status wasn't applied correctly (again, an oversight on my part), so I fixed that, too.

packages/core/src/directory.ts
```diff
@@ -0,0 +1,107 @@
...
+                return;
+            }
+
+            await org.removeMember(existingUser, false);
+
+            // Remove any existing invites
+            const existingInvite = org.invites.find((invite) => invite.email === existingUser.email);
+            if (existingInvite) {
+                org.removeInvite(existingInvite);
+            }
```
BrunoBernardino, line 92: @MaKleSoft this was enough for me to get it working locally. Let me know if there's anything else I'm missing.

packages/core/src/org.ts
```diff
@@ -69,17 +80,52 @@ export class OrgMember extends Serializable {
     /** the members organization role */
     role: OrgRole = OrgRole.Member;
 
+    status: OrgMemberStatus = OrgMemberStatus.Active;
+
+    /** external id of a sync service, if it exists */
+    directoryExternalId?: string;
```
BrunoBernardino, line 86: @MaKleSoft I created this so we can allow updating emails, for example.

packages/app/src/elements/org-settings.ts
```diff
@@ -169,4 +210,55 @@ export class OrgSettingsView extends Routing(StateMixin(LitElement)) {
...
+
+    private _renderDirectorySettings() {
+        const org = this._org!;
+
+        let sectionHtml: TemplateResult<1>;
+
+        if (org.directory.syncProvider !== "none") {
+            const scimSecret = bytesToBase64(org.directory.scim!.secret, true);
+            // TODO: Make this section more helpful and pretty
+            // TODO: Get proper SCIM host + port
```
BrunoBernardino, line 222: @MaKleSoft I haven't figured out how to get the server's URL and port for the frontend by "forcefully inspecting" :sweat_smile: let me know if there's a way or if I need to come up with something.

packages/core/src/directory.ts
```diff
@@ -0,0 +1,107 @@
...
+    }
+
+    async userCreated(user: DirectoryUser, orgId: string) {
+        const org = (orgId && (await this.server.storage.get(Org, orgId))) || null;
+        if (org && org.directory.syncProvider === "scim" && org.directory.syncMembers) {
+            const memberExists = org.members.some((member) => member.directoryExternalId === user.externalId);
+
+            if (memberExists) {
+                return;
+            }
```
BrunoBernardino, line 44: Without this we just kept creating the same user, which isn't a good idea since AD can send the same request more than once (if it failed or something).

packages/core/src/directory.ts
```diff
@@ -0,0 +1,107 @@
...
+        }
+    }
+    async userDeleted(user: DirectoryUser, orgId: string): Promise<void> {
+        const org = (orgId && (await this.server.storage.get(Org, orgId))) || null;
+        if (org && org.directory.syncProvider === "scim" && org.directory.syncMembers) {
+            const existingUser = org.members.find((member) => member.directoryExternalId === user.externalId);
+
+            if (!existingUser) {
+                return;
+            }
```
BrunoBernardino, line 84: Same as above.

packages/core/src/directory.ts
MaKleSoft: Unfortunately there is a little bit more work involved. See https://github.com/padloc/padloc/blob/v4/packages/core/src/server.ts#L1115-L1133

packages/app/src/elements/org-settings.ts
MaKleSoft: I think this should be provided as `org.directory.scim.url` or something, that is set on the server side, probably [here](https://github.com/padloc/padloc/blob/feature/scim-v1/packages/core/src/server.ts#L1037).

packages/core/src/directory.ts
```diff
@@ -0,0 +1,115 @@
...
+            if (existingInvite) {
+                org.removeInvite(existingInvite);
+
+                try {
+                    const auth = await this.server.getAuth(existingInvite.email);
+                    auth.invites = auth.invites.filter((invite) => invite.id !== existingInvite.id);
+                    await this.server.storage.save(auth);
+                } catch (_error) {
+                    // Ignore
+                }
```
BrunoBernardino, line 99: @MaKleSoft I had to create and expose the `getAuth` (a simplified version) to allow to do this. I think this is enough now? It seemed harder to expose/inject the controller here, though perhaps better?

packages/core/src/server.ts
```diff
@@ -1014,14 +1028,24 @@ export class Controller extends API {
...
-                org,
-                org.getMember({ id: org.owner })!,
-                org.getMember({ id: owner })!
-            );
+        if (org.directory.syncProvider === "scim") {
+            if (!org.directory.scim) {
+                org.directory.scim = new ScimSettings();
+                org.directory.scim.secret = await getCryptoProvider().randomBytes(16);
+                // TODO: This works locally for now, but needs to be updated to be more bullet-proof
+                org.directory.scim.url = this.config.clientUrl.replace("8080", "5000");
```
BrunoBernardino, line 1039: @MaKleSoft Why wouldn't this work until I made the `url: string` default to `""` in `ScimSettings`? It would be set but not stored.

packages/core/src/server.ts
```diff
@@ -1996,7 +2027,27 @@ export class Server {
         await Promise.all(promises);
 
         org.vaults = org.vaults.filter((v) => !deletedVaults.has(v.id));
-        org.members = org.members.filter((m) => !deletedMembers.has(m.id));
+        org.members = org.members.filter((m) => !m.accountId || !deletedMembers.has(m.accountId));
+    }
+
+    // Simplified version of the controller one
```
BrunoBernardino, line 2033: @MaKleSoft This is the simplified version I created. Let me know if there's a better way you'd prefer (like injecting/exposing the `controller`).

packages/server/src/init.ts
```diff
@@ -245,6 +278,8 @@ async function init(config: PadlocConfig) {
         legacyServer
     );
 
+    new DirectorySync(server, directoryProviders);
```
BrunoBernardino, line 281: I think I'll need to pass the `config.directory` additionally, and add a `scim.url` property to it, to properly build up the URL.

packages/core/src/org.ts
```diff
@@ -119,10 +165,29 @@ export class OrgSecrets extends Serializable {
...
+        accountId?: AccountID;
+    };
     revision: string;
 }
 
+export class ScimSettings extends Serializable {
+    @AsBytes()
+    secret!: Uint8Array;
+
+    url: string = "";
```
BrunoBernardino, line 179: @MaKleSoft this is the thing that wouldn't be saved when it was set as `url!: string;`, which I don't quite understand.

packages/core/src/directory.ts
MaKleSoft: We may want to think of a cleaner/better way to do this in the future but for now this should be fine.

packages/core/src/server.ts
MaKleSoft: Actually, since we're not using this anywhere else, why not just implement this on `ScimServer`? The only important bit here is this:

```ts
auth = await this.storage.get(Auth, await getIdFromEmail(email));
```

and `storage` is already available to `ScimServer`.

packages/core/src/org.ts
MaKleSoft: Explained in my other comment ;)

packages/core/src/directory.ts
```diff
@@ -0,0 +1,188 @@
...
+            if (existingGroupIndex === -1) {
+                return;
+            }
+
+            org.groups.splice(existingGroupIndex, 1);
+
+            await this.server.updateMetaData(org);
+            await this.server.storage.save(org);
+
+            // TODO: Go through users and remove membership?
```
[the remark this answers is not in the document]
MaKleSoft: Nope, groups are pretty simple that sense. Of course access to vaults has to be updated based on the group changes but that happens asynchronously on the client side: https://github.com/padloc/padloc/blob/v4/packages/core/src/app.ts#L1239-L1246

packages/core/src/directory.ts
```diff
@@ -0,0 +1,263 @@
...
+                            existingOrgMember = member;
+                        }
+                    }
+
+                    if (existingOrgMember) {
+                        const existingMemberIndex = existingGroup.members.findIndex(
+                            ({ email }) => existingOrgMember!.email === email
+                        );
+                        if (existingMemberIndex !== -1) {
+                            existingGroup.members.splice(existingMemberIndex, 1);
```
BrunoBernardino, line 230: @MaKleSoft any idea why this isn't enough to persist a group membership removal? Adding by mutating this array works, but removing doesn't persist. It's the final piece to start testing with real AD.

MaKleSoft: Hm, no idea. I can try debugging this once we have a test setup though.

BrunoBernardino: So you're saying it should be enough to simply update the `members` array and save the org, right?

packages/app/src/elements/org-settings.ts
```diff
@@ -169,4 +220,64 @@ export class OrgSettingsView extends Routing(StateMixin(LitElement)) {
...
+
+        let scimGroupsUrl = "";
+        let scimUsersUrl = "";
+
+        const syncEnabled = org.directory.syncProvider !== "none";
+
+        if (syncEnabled && org.directory.scim) {
+            const scimSecret = bytesToBase64(org.directory.scim.secret, true);
+            scimGroupsUrl = `${org.directory.scim.url}/Groups?org=${org.id}&token=${scimSecret}`;
+            scimUsersUrl = `${org.directory.scim.url}/Users?org=${org.id}&token=${scimSecret}`;
```
MaKleSoft, line 235: Two things:
1. I think it would be better if the urls were fully constructed on the server side. I.e. instead of `scim.url` we should have `scim.groupsUrl` and `scim.usersUrl`. The url structure is something that the client should not have to concern itself with, and harcoding this into the client will make it a lot harder to change the url structure in the future if we need to. Which brings me to my second point
2. I think for the url I this better: `${baseUrl}/${orgId}/${"users" | groups}/?token=${scimSecret}`

packages/core/src/directory.ts
```diff
@@ -0,0 +1,263 @@
...
+    memberIdsToRemove?: string[];
+}
+
+export interface DirectorySubscriber {
+    userCreated(user: DirectoryUser, orgId: OrgID): Promise<OrgMember | void>;
+    userUpdated(user: DirectoryUser, orgId: OrgID, userId: string): Promise<OrgMember | void>;
+    userDeleted(user: DirectoryUser, orgId: OrgID, userId: string): Promise<void>;
+
+    groupCreated(group: DirectoryGroup, orgId: OrgID): Promise<Group | void>;
+    groupUpdated(groupChanges: DirectoryGroupChanges, orgId: OrgID, groupId: string): Promise<Group | void>;
```
MaKleSoft, line 31: I think I prefer `groupUpdated` to have the same signature as the other functions. Changes can (and probably should) be inferred by the directory subscriber by comparing the existing group with the new one.

packages/core/src/directory.ts
```diff
@@ -0,0 +1,263 @@
...
+        if (org && org.directory.syncMembers) {
+            let existingUser: OrgMember | null = null;
+            for (const member of org.members) {
+                if (existingUser) {
+                    continue;
+                }
+
+                const provisioningId = await getIdFromEmail(member.email);
+
+                if (member.accountId === userId || member.id === userId || provisioningId === userId) {
```
MaKleSoft, line 82: I don't like this ambiguity in how we deal with ids here. When I was testing the endpoint it wasn't clear whether the expected user id was the email, the account id or something else. We should decide on a single identifier and use that throughout. The account id is not a good choice for this since it's not guaranteed to be defined yet, so it must be one of:

1. The email, or a hash of it
2. The external id, mapped to the email
3. A random id, mapped to the email

2 and 3 have the advantage that they allow the id to be constant, which should be a hard requirement for a public id anyway.

packages/core/src/directory.ts
```diff
@@ -0,0 +1,263 @@
...
+
+export interface DirectoryGroupChanges {
+    newName?: string;
+    memberIdsToAdd?: string[];
+    memberIdsToRemove?: string[];
+}
+
+export interface DirectorySubscriber {
+    userCreated(user: DirectoryUser, orgId: OrgID): Promise<OrgMember | void>;
+    userUpdated(user: DirectoryUser, orgId: OrgID, userId: string): Promise<OrgMember | void>;
```
MaKleSoft, line 27: I don't think we need the third `userId` argument, since that is what `DirectoryUser.externalId` is for.

BrunoBernardino: `externalId` is supposed to be given to us (provider), not generated by us, as per my understanding of the RFC: https://datatracker.ietf.org/doc/html/rfc7643#section-3.1 so we need the `DirectoryUser.id` which is what I was using as `userId` here.

packages/core/src/directory.ts
BrunoBernardino: Agreed. I had the impression the `getIdFromEmail()` would match these, so that's what I was going for here. I prefer your idea of generating something else and keeping that map outside of this.

packages/core/src/directory.ts
BrunoBernardino: I actually would prefer `userUpdated` to have the same signature as this, because a `PATCH` should always have changes and not the full object. Also, for example, arrays can be very costly to "infer" when something has been removed, for example, especially after a few thousands of items. When a member is added, you need to fetch all previous members to build the "updated" version, and when it's removed, you need to fetch everything and remove the right item as well.

packages/app/src/elements/org-settings.ts
BrunoBernardino: Agreed!

packages/core/src/directory.ts
MaKleSoft: If you look at my changes, you'll see that there is nothing to infer on the `DirectorySubscriber` side. We simply mirror whatever we get from the directory provider. Updating and inferring changes should happen on the `ScimServer` side of things.

MaKleSoft: > When a member is added, you need to fetch all previous members to build the "updated" version, and when it's removed, you need to fetch everything and remove the right item as well.

I agree that it's less efficient, but I don't think the difference big enough to make a difference. Simply rebuilding the member lists on the other hand is much more robust and guarantees that everything is always in sync, even if an intermediate update fails.

MaKleSoft: You can also see that the new code is much simpler and therefore less error-prone: https://github.com/padloc/padloc/pull/429/commits/00b33e801ee964dafbb84b96f49bf30a0d5e1bcc#diff-e48f537ea43df65f2f733b68e321624fd34237257fab618884ade5c6a77ef438L181-R178

packages/core/src/provisioning.ts
```diff
@@ -116,7 +116,7 @@ export class AccountProvisioning extends Storable {
 
     accountId?: AccountID = undefined;
 
-    status: ProvisioningStatus = ProvisioningStatus.Unprovisioned;
```
BrunoBernardino: @MaKleSoft I've reverted this because otherwise the default `simple` provisioning backend won't work, and overriding the status for tests seemed brittle. I'll instead add a new server setting to allow/disallow signup, which should give more flexibility to people who need automatic provisioning and also allow non-provisioned accounts to be created (assuming the server is behind a VPN, for example).

MaKleSoft: Makes sense. I would add maybe add a setting like `PTC_PROVISIONING_BASIC_DEFAULT_STATUS` for the `BasicProvisioner`, but for `DirectoryProvisioner` it is better to "hardcode" the default status by overriding [`_getOrCreateAccountProvisioning`](https://github.com/padloc/padloc/blob/v4/packages/core/src/provisioning.ts#L332)

MaKleSoft: Or, even better, add a protected method `_getDefaultAccountProvisioning` to `BasicProvisioner`, which would be consumed [here](https://github.com/padloc/padloc/blob/feature/scim-v1/packages/core/src/provisioning.ts#L328) and  could then be overridden by `DirectoryProvisioner`.

BrunoBernardino: I've added the new setting to disable signup, and tests should be passing again. Do you still want me to add this new method for the default?

packages/core/src/server.ts
```diff
@@ -79,6 +79,10 @@ export class ServerConfig extends Config {
     @ConfigParam("boolean")
     verifyEmailOnSignup = true;
 
+    /** Whether or not to disable creating an account via the signup form */
+    @ConfigParam("boolean")
+    disableSignup = false;
```
MaKleSoft, line 84: Doesn't this prevent signup altogether? That's not what we want as it would prevent onboarding even for provisioned members (even provisioned members need to complete the signup process). What we want instead is a setting to change the default provisioning status, i.e. the provisioning status for users for which no status has been set explicitly. Also, this setting should be on the provisioner, not the server itself.

MaKleSoft: I'll make that change myself later today. You'll see what I mean.

BrunoBernardino: Gotcha, and yes, it prevents signup. This was meant to be a setting that wasn't directly related to provisioning, just something that someone might want to do on their server, for example, and helped if you just wanted directory provisioning, for example.

MaKleSoft: I guess there was a misunderstanding on how provisioning works. The provisioner simply "whitelists" a given email address, but doesn't actually create the account. So the user still has to go through the signup flow, even if their provisioning status has been set to `active`.

packages/core/src/provisioning.ts
```diff
@@ -314,6 +350,19 @@ export class BasicProvisioner implements Provisioner {
...
+            () =>
+                new AccountProvisioning({
+                    status: this.config.default.status,
+                    statusLabel: this.config.default.statusLabel,
+                    statusMessage: this.config.default.statusMessage,
+                    actionUrl: this.config.default.actionUrl,
+                    actionLabel: this.config.default.actionLabel,
+                })
+        );
+    }
```
MaKleSoft, line 367: The default account provisioning is what is returned for an unknown email, i.e. somebody for whom no provisioning status has been set yet. This can now be configured via environment variables or through a database entry with the id `[default]`, which takes precedence. The default provisioning status basically determines whether an unknown user should be allowed to sign up for the service or not, or in other words, whether you need to whitelist or blacklist users. For example, if you want to only allow users that have been whitelisted/provisioned explicitly, the default status should be set to `unprovisioned`. If you want the instance to be open to anyone, the default status should be `active` (which is the default).

packages/core/src/server.ts
BrunoBernardino: Ohh, indeed. This makes more sense, then. I guess this isn't necessary at all, then.

packages/server/src/scim.ts
```diff
@@ -124,7 +124,10 @@ export class ScimServer implements DirectoryProvider {
     }
 
     private async _getScimOrg(orgId: OrgID) {
-        const prov = await this.storage.get(OrgProvisioning, orgId);
+        const prov = await this.storage.get(OrgProvisioning, orgId).catch(() => null);
```
BrunoBernardino, line 127: Why not wrap this in `try/catch` returning null instead? Like:

```suggestion
        try {
          const prov = await this.storage.get(OrgProvisioning, orgId);
        } catch (_error) {
          return null;
        }
```

Since that seems easier to read and maintain.

packages/server/src/scim.ts
```diff
@@ -241,18 +244,39 @@ export class ScimServer implements DirectoryProvider {
         return null;
     }
 
+    private _sendErrorResponse(httpRes: ServerResponse, status: number, detail: string) {
```
BrunoBernardino, line 254: Thanks, I was planning to do this and forgot.

packages/server/src/scim.ts
MaKleSoft: Because it's 5 lines instead of 4 😄. JK, I have no preference either way. We can make that change, but then we should also remove the conditional return below.

Honestly I kind of regret having `Storage::get` throw and error instead of simply returning `null` if nothing is found, but it's a little late to change that now.

MaKleSoft:
Looks like I broke the v3 compatibility tests. It's a little late now, but I'll fix that tomorrow.

packages/server/src/scim.ts
BrunoBernardino: Yes, I couldn't easily suggest removing the conditional below.

MaKleSoft (approved):

MaKleSoft:
Alright, I think this is ready to merge! Great job!

Issue #362, opened 2021-12-29
[Safari] Padloc logo looks blurry and doesn't animate


MaKleSoft:
<img width="515" alt="Screenshot 2021-12-29 at 09 33 09" src="https://user-images.githubusercontent.com/1270940/147642645-f2be67b8-f730-4514-a643-603a5f4e7e0b.png">

PR #443, opened 2022-05-06
Update logo and icon

MaKleSoft:
Ok, so after going back and forth for a while and showing the both versions of the logo and icon to a couple of people, I decided to go with the "stacked" one instead of the gradient and drop shadow. The former one simply looks sharper/cleaner and works better in small sizes. It should also fix this issue: https://github.com/padloc/padloc/issues/362

I may keep tweaking it a bit so this is not necessarily the final version, but I still wanted to get this in so I can put it off my mind for now.


MaKleSoft:
@BrunoBernardino Hm, the web extension build is failing but I'm not sure why. Can you take a look?

BrunoBernardino:
@MaKleSoft looks like we're running into https://github.com/cardinalby/webext-buildtools-chrome-crx-action/issues/10 see there for a quick/hotfix in `package.json` if you need it.

MaKleSoft:
Hm, I guess we could also just ignore it, right? Since I assume this is going to be fixed soon, anyway.

BrunoBernardino:
Definitely not a blocker, just can't build web/chrome extension automatically for now.

PR #446, opened 2022-05-08
Use different strategy for dealing with scim resources:

MaKleSoft:
- Store "virtual" scim-based representation of org in provisioning meta data
- Assign random ids to all scim resources
- use "virtual" scim model to resolve ids and track changes

### DIFF
--- a/packages/core/src/directory.ts
+++ b/packages/core/src/directory.ts
@@ -1,35 +1,30 @@
 import { Org, OrgID, OrgMember, OrgMemberStatus, Group } from "./org";
 import { Server } from "./server";
-import { getIdFromEmail } from "./util";
+import { getIdFromEmail, uuid } from "./util";
 import { Auth } from "./auth";
+import { Err, ErrorCode } from "./error";
 
 export interface DirectoryUser {
     externalId?: string;
-    email?: string;
+    email: string;
+    active: boolean;
     name?: string;
-    active?: boolean;
 }
 
 export interface DirectoryGroup {
     externalId?: string;
     name: string;
-    members?: DirectoryUser[];
-}
-
-export interface DirectoryGroupChanges {
-    newName?: string;
-    memberIdsToAdd?: string[];
-    memberIdsToRemove?: string[];
+    members: DirectoryUser[];
 }
 
 export interface DirectorySubscriber {
-    userCreated(user: DirectoryUser, orgId: OrgID): Promise<OrgMember | void>;
-    userUpdated(user: DirectoryUser, orgId: OrgID, userId: string): Promise<OrgMember | void>;
-    userDeleted(user: DirectoryUser, orgId: OrgID, userId: string): Promise<void>;
+    userCreated(user: DirectoryUser, orgId: OrgID): Promise<void>;
+    userUpdated(user: DirectoryUser, orgId: OrgID, previousEmail?: string): Promise<void>;
+    userDeleted(user: DirectoryUser, orgId: OrgID): Promise<void>;
 
-    groupCreated(group: DirectoryGroup, orgId: OrgID): Promise<Group | void>;
-    groupUpdated(groupChanges: DirectoryGroupChanges, orgId: OrgID, groupId: string): Promise<Group | void>;
-    groupDeleted(group: DirectoryGroup, orgId: OrgID, groupId: string): Promise<void>;
+    groupCreated(group: DirectoryGroup, orgId: OrgID): Promise<void>;
+    groupUpdated(group: DirectoryGroup, orgId: OrgID, previousName?: string): Promise<void>;
+    groupDeleted(group: DirectoryGroup, orgId: OrgID): Promise<void>;
 }
 
 export interface DirectoryProvider {
@@ -53,6 +48,7 @@ export class DirectorySync implements DirectorySubscriber {
             }
 
             const newOrgMember = new OrgMember({
+                id: await uuid(),
                 name: user.name,
                 email: user.email,
                 status: OrgMemberStatus.Provisioned,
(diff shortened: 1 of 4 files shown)


packages/core/src/directory.ts
```diff
@@ -63,74 +59,59 @@ export class DirectorySync implements DirectorySubscriber {
...
+                return;
             }
 
-            if (!existingUser) {
-                return;
+            // Changing the email address of a user is a little more
+            // involved than that. Let's disable it for now.
+
+            if (previousEmail && previousEmail !== user.email) {
+                throw new Err(ErrorCode.NOT_SUPPORTED, "Updating user emails is not supported at this time");
```
MaKleSoft, line 80: Actually changing the email of a user is actually fairly complicated and currently not supported. Let's tackle this at a later time.

packages/server/src/init.ts
```diff
@@ -267,6 +267,10 @@ async function init(config: PadlocConfig) {
         });
     }
 
+    if (config.directory.scim && !config.server.scimServerUrl) {
+        config.server.scimServerUrl = config.directory.scim.url;
+    }
```
MaKleSoft, line 272: We need the url setting in `ScimServer` as well to construct the resource urls. This way we don't have to set it twice.

packages/server/src/scim.ts
```diff
@@ -83,6 +88,11 @@ interface ScimGroupPatch {
     Operations: ScimGroupPatchOperation[];
 }
 
+interface ScimOrg {
+    users: ScimUser[];
+    groups: ScimGroup[];
+}
```
MaKleSoft, line 94: This is the "virtual" scim model of the org. We'll use this to store and resolve ids and track changes.

packages/server/src/scim.ts
```diff
@@ -96,6 +106,41 @@ export class ScimServer implements DirectoryProvider {
...
 
+    private async _getScimOrg(orgId: OrgID) {
+        const prov = await this.storage.get(OrgProvisioning, orgId);
+        return (prov.metaData?.scim || { users: [], groups: [] }) as ScimOrg;
+    }
+
+    private async _saveScimOrg(orgId: OrgID, scimOrg: ScimOrg) {
+        const prov = await this.storage.get(OrgProvisioning, orgId);
+        prov.metaData = prov.metaData || {};
+        prov.metaData.scim = scimOrg;
```
MaKleSoft, line 117: I figured the `metaData` property on `OrgProvisioning` was a good place as any to store this. We **could** also store this in a separate table though.

packages/server/src/scim.ts
```diff
@@ -113,14 +158,10 @@ export class ScimServer implements DirectoryProvider {
...
         }
 
-        if (!user.name.formatted) {
+        if (!user.name?.formatted) {
             return "User must contain name.formatted";
         }
 
-        if (user.meta.resourceType !== "User") {
-            return 'User meta.resourceType must be "User"';
-        }
```
MaKleSoft, line 122: [From the spec](https://datatracker.ietf.org/doc/html/rfc7643#section-3.1): 
```
All "meta"
      sub-attributes are assigned by the service provider (have a
      "mutability" of "readOnly"), and all of these sub-attributes have
      a "returned" characteristic of "default".  This attribute SHALL be
      ignored when provided by clients.
   ```

File: packages/server/src/scim.ts [hunk 6]
```diff
@@ -141,10 +182,6 @@ export class ScimServer implements DirectoryProvider {
             return "Group must contain displayName";
         }
 
-        if (group.meta.resourceType !== "Group") {
-            return 'Group meta.resourceType must be "Group"';
-        }
```
MaKleSoft, line 146: See above

packages/server/src/scim.ts
```diff
@@ -182,11 +219,6 @@ export class ScimServer implements DirectoryProvider {
 
         let objectId = (objectIdMatches && objectIdMatches[1]) || "";
 
-        // The id of groups is their name, so we base64 encode it, and need to decode here
-        if (url.pathname.startsWith("/Groups")) {
-            objectId = base64ToString(objectId);
-        }
```
MaKleSoft, line 188: Using group names as resource ids is a bad idea, because...

```
      It MUST be a stable, non-reassignable identifier that does not change when the
      same resource is returned in subsequent requests.
```

(See https://datatracker.ietf.org/doc/html/rfc7643#section-3.1)

MaKleSoft:
@BrunoBernardino I've gone over your implementation carefully and thought about it for a while, and I've come to the conclusion that we need a more robust way of assigning/resolving ids and tracking changes. Here is what I've come up with:

- Instead of applying changes to the underlying `Org` object directly, we now use an intermediate model representing the entire organisation and all of it's SCIM resources. When a resource is created, modified or deleted, we first perform this operation on the [intermediate SCIM model](https://github.com/padloc/padloc/blob/feature/scim-v1-revised/packages/server/src/scim.ts#L91), and then pass the changes on to all directory subscribers.
- Instead of deriving ids from user emails and group names, we now generate random ids during resource creation. The `ScimOrg` model can the be used to resolve those unique ids and find the corresponding SCIM resources, which can then be transformed into the standarised `DirectoryUser` and `DirectoryGroup` models.
- On the `DirectorySubscriber` side, we continue to use the email and group names to uniquely identify org members and groups respectively. The mapping to the SCIM id happens in `ScimServer` and is of no concern to directory subscribers.
- Conflicts like duplicate emails or group names are detected and resolved on the `ScimServer` side. Directory subscribers should either ignore or gracefully handle those conflicts.

I believe this approach is more robust and future-proof, and better aligns with the spec, but I didn't want to just make those changes without consulting you first. That's why I've implemented this as a PR to your PR. Please review and merge if you are ok with the changes 😊

BrunoBernardino (approved):
Thank you so much! I read through these code changes and they make sense and look good. I didn't pull this down but will still merge. If I find any issues with my testing, I'll just tweak over in the other PR.

Thanks again!

PR #451, opened 2022-05-20
Fix scrolling in Security Report

BrunoBernardino:
Also add an automatic action to update the test deployment on push/merge.


MaKleSoft (approved):
👍

PR #452, opened 2022-05-20
Fix automatic deployment update

BrunoBernardino:
Couldn't fix it before because I couldn't run it yet.

### DIFF
--- a/.github/workflows/update-deployment.yml
+++ b/.github/workflows/update-deployment.yml
@@ -25,6 +25,8 @@ on:
 jobs:
     trigger-gh-action:
         runs-on: ubuntu-latest
+        env:
+            DEPLOY_BRANCH: ${{ github.event.inputs.instance || 'test' }}
         steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/update-deployment.yml
```diff
@@ -34,7 +36,7 @@ jobs:
                     --header 'Content-Type: application/json' \
                     --data '{
                         "inputs": {
-                            "instance": "${{ github.event.inputs.instance || \'test\' }}",
+                            "instance": "${{ env.DEPLOY_BRANCH }}",
```
BrunoBernardino, line 39: Well, this works now, but I'm getting a 404... :thinking:

BrunoBernardino: This is working now (calling the dispatch) even though on the other repo that's failing, but it'll be a separate PR in that repo.

MaKleSoft (approved):

PR #456, opened 2022-06-02
Enable SAML SSO for SCIM (Google Workspaces)

BrunoBernardino:
This adds support for integrating SCIM with Google Workspaces (or Cloud Identity) via SAML SSO.

Basically, when someone initiates SSO from their provider, it'll go to the right URL with the authorization info (once authorized in the provider), and we'll take that information, transform to a SCIM (new user) request, and parse that in.

It's not very pretty, but it works. This needs to be tested and still needs some polish (redirect to the start page once the user has been created or already exists).

As a reference, some things I tried that did not work, and some other interesting bits of information:
- [Cloud Identity](https://cloud.google.com/identity) (it's different from Google Workspaces, and does similar tihngs)
- And [the only supported architectures to integrate with their directory](https://cloud.google.com/architecture/identity/reference-architectures)
- Their [AD instructions](https://cloud.google.com/architecture/identity/preparing-your-g-suite-or-cloud-identity-account) are similar to ours, which is nice.
- [LDAP would be quite convoluted to setup](https://support.google.com/a/answer/9048434?authuser=2), and requires Business Plus or greater, which starts at roughly €15 / month / user.

### DIFF
--- a/docs/examples/config/example.env
+++ b/docs/examples/config/example.env
@@ -287,7 +287,8 @@
 
 # PL_PROVISIONING_BACKEND=directory
 # PL_DIRECTORY_PROVIDERS=scim
-# PL_DIRECTORY_SCIM_PORT=[required]
+# PL_DIRECTORY_SCIM_URL=http://localhost:5000
+# PL_DIRECTORY_SCIM_PORT=5000
 
 # =============================================================================
 # SERVER
@@ -301,6 +302,7 @@
 # PL_SERVER_VERIFY_EMAIL_ON_SIGNUP=true
 # PL_SERVER_DEFAULT_AUTH_TYPES=email
 # PL_SERVER_DISABLE_SIGNUP=false
+# PL_SERVER_SCIM_SERVER_URL=http://localhost:5000
 
 
 # =============================================================================
--- a/packages/server/src/scim.ts
+++ b/packages/server/src/scim.ts
@@ -180,9 +180,6 @@ export class ScimServer implements DirectoryProvider {
     }
 
     private _validateScimUser(user: ScimUser) {
-        // TODO: Remove this
-        console.log(JSON.stringify({ user }, null, 2));
-
         if (!this._getScimUserEmail(user)) {
             return "User must contain email";
         }
@@ -813,7 +810,7 @@ export class ScimServer implements DirectoryProvider {
         }
 
         // TODO: Remove this
-        console.log(JSON.stringify({ newUser }, null, 2));
+        console.log(JSON.stringify({ newUser, orgId }, null, 2));
 
         return this._createScimUser(newUser, orgId, secretToken, httpRes, true);
     }
--- a/packages/server/src/scim.ts
+++ b/packages/server/src/scim.ts
@@ -373,6 +373,7 @@ export class ScimServer implements DirectoryProvider {
 
             return this._sendResponse(httpRes, 201, newUser);
         } catch (error) {
+            console.error(error);
             return this._sendErrorResponse(httpRes, 500, "Unexpected Error");
         }
     }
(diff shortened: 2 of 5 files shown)


packages/server/src/scim.ts
```diff
@@ -699,6 +703,96 @@ export class ScimServer implements DirectoryProvider {
...
+                          firstNamePropertyEndIndex
+                      )
+                    : "";
+            const lastName =
+                lastNamePropertyStartIndex !== -1 && lastNamePropertyEndIndex !== -1
+                    ? xmlSamlResponse.substring(
+                          lastNamePropertyStartIndex + lastNamePropertyStart.length,
+                          lastNamePropertyEndIndex
+                      )
+                    : "";
```
BrunoBernardino, line 764: @MaKleSoft your thoughts on this "mess" are welcome. I think it's still better than adding an xml parser dependency.

BrunoBernardino:
This is working fine locally now (with the proper redirect UX and updated docs), but I still ran into some unrelated issues with the test deployment to verify this in an "external deploy" setting.

BrunoBernardino:
Alright, it's now working for the test instance, @MaKleSoft (the main remaining issue was a config one, for non-local setups)! You should be able to test it and auto-provision your own user from https://admin.google.com/u/2/ac/apps/saml/868009575075/sp

docs/examples/config/example.env
```diff
@@ -287,7 +287,8 @@
 
 # PL_PROVISIONING_BACKEND=directory
 # PL_DIRECTORY_PROVIDERS=scim
-# PL_DIRECTORY_SCIM_PORT=[required]
+# PL_DIRECTORY_SCIM_URL=http://localhost:5000
```
BrunoBernardino, line 290: We _might_ want to consider eventually not requiring this one and just use the `serverScimServerUrl` isntead, but it seemed out-of-scope to pass the server config in as well, and it's not terrible to have two separate variables that need to have the same value (even if not ideal).

packages/server/src/scim.ts
```diff
@@ -337,6 +327,14 @@ export class ScimServer implements DirectoryProvider {
             const email = this._getScimUserEmail(newUser);
 
             if (scimOrg.users.some((user) => this._getScimUserEmail(user) === email)) {
+                // Just redirect and don't error out for SAML if this user already exists
+                if (isComingFromSaml) {
+                    httpRes.statusCode = 302;
+                    httpRes.setHeader("Location", `/start?email=${email}`);
```
BrunoBernardino, line 333: This way we get some benefit from SSO (not having to fill in the email).

packages/server/src/scim.ts
```diff
@@ -699,6 +725,96 @@ export class ScimServer implements DirectoryProvider {
...
+        }
+
+        try {
+            const body = await readBody(httpReq);
+            const params = new URLSearchParams(body);
+            const samlResponse = params.get("SAMLResponse") || "";
+            const xmlSamlResponse = Buffer.from(samlResponse, "base64").toString("ascii");
+
+            // TODO: Remove this
+            console.log(JSON.stringify({ xmlSamlResponse }, null, 2));
```
BrunoBernardino, line 743: Keeping this for now, like the other SCIM data, so we can debug in case something strange happens.

packages/server/src/scim.ts
```diff
@@ -699,6 +725,96 @@ export class ScimServer implements DirectoryProvider {
...
+                        primary: true,
+                    },
+                ],
+            };
+        } catch (e) {
+            return this._sendErrorResponse(httpRes, 400, "Failed to read request body.");
+        }
+
+        // TODO: Remove this
+        console.log(JSON.stringify({ newUser, orgId }, null, 2));
```
BrunoBernardino, line 813: Same as above on potential future debugging.

PR #461, opened 2022-06-06
Support exporting all vaults

### LINKED ISSUE #253: Feature Request: Allow exporting ALL vaults
It would be very helpful to have an option of "All Vaults" when creating an encrypted backup of your vaults.

BrunoBernardino:
Fixes #253

This basically creates a `.zip` file with all the vault export files inside.

I also updated the dependency version because on 3.8.0 there was an important security fix.

### DIFF
--- a/packages/app/src/elements/export-dialog.ts
+++ b/packages/app/src/elements/export-dialog.ts
@@ -2,6 +2,7 @@ import { Vault } from "@padloc/core/src/vault";
 import { translate as $l } from "@padloc/locale/src/translate";
 import { saveFile } from "@padloc/core/src/platform";
 import { stringToBytes } from "@padloc/core/src/encoding";
+import JSZip from "jszip";
 import { CSV, PBES2 } from "../lib/import";
 import { supportedFormats, asCSV, asPBES2Container } from "../lib/export";
 import { app } from "../globals";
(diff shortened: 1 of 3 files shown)


packages/app/src/elements/export-dialog.ts
```diff
@@ -40,15 +54,13 @@ export class ExportDialog extends Dialog<void, void> {
 
                 <div class="small padded" ?hidden=${this._formatSelect && this._formatSelect.value !== CSV.value}>
                     ${$l(
-                        "WARNING: Exporting to CSV format will save your data without encyryption of any " +
-                            "kind which means it can be read by anyone. We strongly recommend exporting your data as " +
-                            "a secure, encrypted file, instead!"
+                        `WARNING: Exporting to CSV format will save your data without encryption of any kind which means it can be read by anyone. We strongly recommend exporting your data as "a secure, encrypted file, instead!`
```
BrunoBernardino, line 57: There was a typo here (`encyryption`).

packages/app/src/elements/export-dialog.ts
```diff
@@ -81,7 +91,72 @@ export class ExportDialog extends Dialog<void, void> {
 
             case PBES2.value:
                 this.open = false;
-                const password = await prompt($l("Please choose a password to protect this backup with!"), {
+                if (!password) {
```
BrunoBernardino, line 94: This is the same password logic, but inside this `if`, and looks funky in the git diff, as if it was completely added.

packages/app/src/elements/export-dialog.ts
```diff
@@ -11,21 +12,34 @@ import { Dialog } from "./dialog";
...
+
+        if (this._vaultSelect?.value !== ALL_VAULTS_OPTION) {
+            const vault = this._vaultSelect?.value as Vault;
+            if (vault) {
+                itemCount = vault.items.size.toString();
+            }
+        } else if (this._vaultSelect?.value === ALL_VAULTS_OPTION) {
+            itemCount = $l("All");
+        }
+
```
MaKleSoft, line 35: I know you don't care as much about terseness as I do, but you could write this in two lines:

```ts
const vault = this._vaultSelect?.value;
const itemCount = vault instanceof Vault ? vault.items.size : $l("All");
```

MaKleSoft: Also, why not show the actual item count instead of "All"?

packages/app/src/elements/export-dialog.ts
MaKleSoft: Why not just require the `password` argument and remove this whole code block? There's really no reason to have this exact same code in there twice.

MaKleSoft (changes requested):
I think instead of just having an "All" option, it would be better to allow exporting all vaults for a single org. @BrunoBernardino  What do you think?

packages/app/src/elements/export-dialog.ts
BrunoBernardino: You're right, I can simply force this earlier.

BrunoBernardino:
> I think instead of just having an "All" option, it would be better to allow exporting all vaults for a single org. @BrunoBernardino What do you think?

I think that would defeat the purpose here. The idea is to make it easier to have a backup of everything, and "per org" would still mean multiple clicks.

MaKleSoft (approved):
> I think that would defeat the purpose here. The idea is to make it easier to have a backup of everything, and "per org" would still mean multiple clicks.

You're assuming that full backups are the only use case here 😉. In any case, "All Vaults" and "All Vaults for this org" don't have to be mutually exclusive, and I definitely think there would be value in adding support for exporting all vaults for a specific organisation. But I'll leave it up to you whether you want to add that to this PR or save it for a later one.

BrunoBernardino:
Thanks, that makes sense. I think we can consider all by org later, as this resolves #253

PR #466, opened 2022-06-08
Update automatic migration of legacy (v2) accounts.

### LINKED ISSUE #416: Update automatic migration from legacy Padlock Cloud for v4

MaKleSoft:
Update signup flow to include check for and automatic migration of legacy (v2) accounts. Fixes #416

### DIFF
--- a/packages/app/src/elements/login-signup.ts
+++ b/packages/app/src/elements/login-signup.ts
@@ -23,6 +23,8 @@ import { displayProvisioning } from "../lib/provisioning";
 import { StartAuthRequestResponse } from "@padloc/core/src/api";
 import { Confetti } from "./confetti";
 import { singleton } from "../lib/singleton";
+import { PBES2Container } from "@padloc/core/src/container";
+import { importLegacyContainer } from "../lib/import";
 
 @customElement("pl-login-signup")
 export class LoginOrSignup extends StartForm {
@@ -158,10 +160,12 @@ export class LoginOrSignup extends StartForm {
         authenticatorIndex?: number;
         pendingRequest?: StartAuthRequestResponse;
     }): Promise<{
+        email: string;
         token: string;
         accountStatus: AccountStatus;
         provisioning: AccountProvisioning;
         deviceTrusted: boolean;
+        legacyData?: PBES2Container;
     } | null> {
         try {
             if (!req) {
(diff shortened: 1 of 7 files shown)


packages/app/src/elements/login-signup.ts
```diff
@@ -489,6 +489,95 @@ export class LoginOrSignup extends StartForm {
...
+    }): Promise<boolean> {
+        const legacyData = authResponse.legacyData!;
+
+        this._submitEmailButton.start();
+
+        const choice = await alert(
+            $l(
+                "You don't have a Padloc 4 account yet but we've found " +
+                    "an account from an older version. " +
+                    "Would you like to migrate your account to Padloc 4 now?"
```
BrunoBernardino, line 505: I'd recommend switching this to a template literal instead, as it's easier to search for and less problematic than concatenations.

packages/app/src/elements/login-signup.ts
```diff
@@ -489,6 +489,95 @@ export class LoginOrSignup extends StartForm {
...
+            ),
+            {
+                title: "Account Migration",
+                icon: "user",
+                options: [$l("Migrate"), $l("Learn More"), $l("Cancel")],
+            }
+        );
+
+        if (choice === 1) {
+            window.open("https://padloc.app/help/migrate-v3", "_system");
```
BrunoBernardino, line 515: Should this be a new guide, or this old one will work?

packages/app/src/elements/login-signup.ts
```diff
@@ -489,6 +489,95 @@ export class LoginOrSignup extends StartForm {
...
+            if (deleteLegacy) {
+                await this.app.api.deleteLegacyAccount();
+            }
+
+            await alert(
+                $l(
+                    "All done! Please note that you won't be able to access your Padloc 4 account " +
+                        "with older versions of the app, so please make sure you have the latest version installed " +
+                        "on all your devices! (You can find download links for all platforms at " +
+                        "https://padloc.app/downloads/). Enjoy using Padloc 4!"
```
BrunoBernardino, line 560: Same as above.

packages/app/src/elements/login-signup.ts
MaKleSoft: The reason why I'm not using template literals within `$l()` is that they're not yet supported by our string extraction tool (https://github.com/padloc/padloc/blob/master/packages/locale/src/extract.ts#L7-L17). At some point I want to improve on that (and on our localisation tools in general) but for now it's easier to just use concatenated strings.

packages/app/src/elements/login-signup.ts
MaKleSoft: Good question! Since there is no new guide yet, the old one will have to do for now. I'll probably just update the old article when I work on the updated website.

packages/app/src/elements/login-signup.ts
BrunoBernardino: I'm not saying to inject the variables on them just yet, so it'll just be a regular string to work with, no?

Issue #331, opened 2021-12-21
Reproducible builds


MaKleSoft:
Provide reproducible builds for all targets along with clear instructions on how to verify that published binaries match with what is published on Github. Not sure to what degree that is even possible for iOS and Android but at the very least it should be possible for the PWA, desktop apps and web extension.

PR #467, opened 2022-06-08
Checksums - Allow anyone to confirm source and delivered code

BrunoBernardino:
For now this only has the initial step on making CSP stricter so we can use it to parse through used files.

Very much WIP for now, but now it should just be a matter of writing up concise docs on how to go through the process, after generating and publishing the checksums via CI as well.

Related to #331


packages/pwa/webpack.config.js
```diff
@@ -66,16 +70,78 @@ module.exports = {
...
+            },
+        },
         new HtmlWebpackPlugin({
             title: name,
             template: resolve(__dirname, "src/index.html"),
             meta: {
                 "Content-Security-Policy": {
                     "http-equiv": "Content-Security-Policy",
-                    content: `default-src 'self' ${serverUrl} https://api.pwnedpasswords.com blob:; style-src 'self' 'unsafe-inline'; object-src 'self' blob:; frame-src 'self'; img-src 'self' blob: data: https:;`,
+                    content: `default-src 'self' ${serverUrl};`, // NOTE: This will be overwritten below, but we need the tag to exist
```
MaKleSoft, line 98: Can you explain what we need `@melloware/csp-webpack-plugin` for? Or in other words, why don't we just build that string ourselves?

BrunoBernardino: Doing the hashing + nonces + type safety for the CSP is not super easy or trivial work. This is also _only_ a dev dependency, so it's not going to impact customers or the end product.

MaKleSoft: Got it. Let's just make sure to properly audit the dependency before using it in production.

BrunoBernardino: It's not really used in "production", only while building the app. If you think that's a problem, I'm happy to just not do the nonces and hashes (which does ensure a lot more security), which will remove the need for the dependency.

BrunoBernardino: I just ran into a problem with the hashes nonces. Since they'll vary on every build, we can't then reliably use checksums as they'll fail for `index.html` and `main.js` when trying to compare locally. I'll remove this dependency and manually build the CSP.

BrunoBernardino:
So, everything _should_ be working now, _but_ the `sha256sums.txt` file I'm getting isn't matching for 3 files:

```txt
./main.js: FAILED
./zxcvbn.chunk.js: OK
./ua-parser.chunk.js: OK
./locale_res_wordlists_pt_json.chunk.js: OK
./locale_res_wordlists_en_json.chunk.js: OK
./index.html: FAILED
./app_src_elements_app_ts.chunk.js: OK
./sw.js: FAILED
./locale_res_translations_fr_json.chunk.js: OK
./locale_res_translations_pl_json.chunk.js: OK
./locale_res_translations_es_json.chunk.js: OK
./locale_res_translations_de_json.chunk.js: OK
./locale_res_translations_ru_json.chunk.js: OK
./vendors-app_node_modules_date-fns_esm_sub_index_js.chunk.js: OK
./locale_res_wordlists_de_json.chunk.js: OK
./vendors-app_node_modules_autosize_src_autosize_js-app_node_modules_dompurify_dist_purify_js-a-10f8da.chunk.js: OK
./app_src_lib_1pux-parser_ts.chunk.js: OK
./papaparse.chunk.js: OK
./locale_res_wordlists_es_json.chunk.js: OK
./locale_res_translations__template_json.chunk.js: OK
./locale_res_wordlists_fr_json.chunk.js: OK
./jsqr.chunk.js: OK
./date-fns.chunk.js: OK
sha256sum: WARNING: 3 computed checksums did NOT match
```

I can't quite tell why, since `PL_PWA_URL` and `PL_SERVER_URL` should match. I'll have to debug this directly in CI later (inspecting those files to see where the differences are).

BrunoBernardino:
Everything should be ready and I just need to do a final round of testing with all the docs and outputs of a release that's being built at https://github.com/padloc/padloc/actions/runs/2488417441

docs/checksums/android/README.md
```diff
@@ -0,0 +1,56 @@
...
+though.
+
+## Verify what you're using has the same source code
+
+1. Download the apk file.
+
+You can do that from
+[our releases page](https://github.com/padloc/padloc/releases) or from the store
+you've downloaded it from, to make sure that wasn't tampered with in the process
+of uploading there.
```
MaKleSoft, line 27: Minor nitpick: I think for that text  to be recognised as part of the markdown list item, it needs to be indented.

.github/workflows/publish-release.yml
```diff
@@ -323,3 +485,40 @@ jobs:
...
+                  cd packages/pwa/dist
+                  find . -type f ! \( -name "*.txt" -or -name "*.png" -or -name "*.json" -or -name "*.svg" -or -name "*.woff2" -or -name "*.map" \) -exec sha256sum {} > sha256sums-web.txt \;
+            - name: Upload checksums
+              uses: svenstaro/upload-release-action@2.2.1
+              with:
+                  repo_token: ${{ secrets.GITHUB_TOKEN }}
+                  tag: "v${{ env.PL_VENDOR_VERSION }}"
+                  file: packages/pwa/dist/sha256sums-web.txt
+                  asset_name: sha256sums-web.txt
+                  prerelease: true
```
MaKleSoft, line 524: Actually, why not release the pwa dist folder as well? For completeness' sake?

BrunoBernardino: I actually thought about it, but it'll be built for production/beta, so it won't be "download + usable" by people, which is weird to have, I suppose.

.github/workflows/publish-release.yml
```diff
@@ -323,3 +485,40 @@ jobs:
...
+            - name: Build
+              run: |
+                  npm run pwa:build
+              env:
+                  PL_PWA_URL: ${{ secrets.PL_PWA_URL }}
+                  PL_SERVER_URL: ${{ secrets.PL_SERVER_URL }}
+            - name: Generate checksums
+              run: |
+                  cd packages/pwa/dist
+                  find . -type f ! \( -name "*.txt" -or -name "*.png" -or -name "*.json" -or -name "*.svg" -or -name "*.woff2" -or -name "*.map" \) -exec sha256sum {} > sha256sums-web.txt \;
```
MaKleSoft, line 516: I don't see `*.js`? Am I missing something?

MaKleSoft: Ah, ok, missed the `not`. Which leads to to a different question: Why exclude those files? Any of those could potentially the source of a vuln if compromised (unlikely, but you never know...). So while we're at it, why not check everything?

.github/workflows/publish-release.yml
BrunoBernardino: Ok, I'll add it.

.github/workflows/publish-release.yml
BrunoBernardino: I'm excluding them because they're not relevant from a security point of view and they're prone to being changed from system to system (imagine someone's using this documentation to check on some Padloc install other than Padloc's web app). I could've instead just included *.js and *.html indeed. Also I needed to exclude the `sha256sums-web` file itself otherwise running the command locally would fail on itself.

MaKleSoft: > because they're not relevant from a security point of view

I would disagree. `json` and `text` files can be read dynamically from the app, and *could* change certain behaviours.  `.map` files can be used to obfuscate the actual source when viewed trough the dev tools. Even things like `png` files [have been known](https://www.mozilla.org/en-US/security/advisories/mfsa2010-41/) to be used as attack vectors. 
So unless there is a very good reason against it, I would simply include all of them.

> they're prone to being changed from system to system

Really? Which ones and how/why?

BrunoBernardino: > So unless there is a very good reason against it, I would simply include all of them.

I understand, but they're not allowed via CSP, and since they're not included in the CSP, we can't really get their list from looking at a web app. Also, since they're not in the CSP, they can't be loaded, even dynamically.

> Really? Which ones and how/why?

`robots.txt` will vary from host to host, and likely the `png` and `svg` files as well. Some people might have slightly different fonts too, and any small change will generate a completely different filename hash and contents, which will fail the checksum.

MaKleSoft: > I understand, but they're not allowed via CSP

Pretty sure all of these file types (with the possible exception of the source maps) can (and maybe should) be whitelisted explicitly through the CSP (`font-src` for fonts, `img-src` for pngs etc.).

> robots.txt will vary from host to host

Where do we have a `robots.txt`? 

> and likely the png and svg files as well. Some people might have slightly different fonts too, and any small change will generate a completely different filename hash and contents, which will fail the checksum.

I don't understand - they're included with the source code so why should the vary? Are you talking about white-labeling scenarios?

packages/pwa/webpack.config.js
```diff
@@ -66,13 +68,39 @@ module.exports = {
...
+            },
+        },
         new HtmlWebpackPlugin({
             title: name,
             template: resolve(__dirname, "src/index.html"),
             meta: {
                 "Content-Security-Policy": {
                     "http-equiv": "Content-Security-Policy",
-                    content: `default-src 'self' ${serverUrl} https://api.pwnedpasswords.com blob:; style-src 'self' 'unsafe-inline'; object-src 'self' blob:; frame-src 'self'; img-src 'self' blob: data: https:;`,
+                    content: `default-src 'self' ${serverUrl}; base-uri 'self'; script-src blob:; connect-src ${serverUrl} https://api.pwnedpasswords.com; style-src 'self' 'unsafe-inline'; object-src 'self' blob:; frame-src 'self'; img-src 'self' blob: data: https:; manifest-src 'self'; worker-src ${pwaUrl}/sw.js;`,
```
MaKleSoft, line 103: `'self'` needs to be removed from all of these directives. Otherwise anything can be loaded from the same origin which defeats the purpose of the whole exercise ;)

MaKleSoft: When you do, don't forget to add `https://icons.duckduckgo.com/ip3/*` to `img-src` though, which is used for loading the favicons.

BrunoBernardino: Good point! :facepalm:

.github/workflows/publish-release.yml
BrunoBernardino: > Pretty sure all of these file types (with the possible exception of the source maps) can (and maybe should) be whitelisted explicitly through the CSP

Ok, I can appreciate that plan, even if it makes this take a bit longer.

> Where do we have a `robots.txt`?

When you try to download a website with `wget -r -p` you'll get "common" files like `robots.txt` that are generally expected, and our beta environment exposes that (any deployment should, really, just to make sure things are/aren't indexed properly).

> I don't understand - they're included with the source code so why should the vary? Are you talking about white-labeling scenarios?

I'm talking about scenarios where people might install on their own server and likely change the logo, among other small customizations. If you don't want to cater to those scenarios, then it's fine to include them in the checks, and let people judge if/when it's fine if they're not matching the checksums, I assume.

MaKleSoft: > When you try to download a website with wget -r -p you'll get "common" files like robots.txt that are generally expected, and our beta environment exposes that (any deployment should, really, just to make sure things are/aren't indexed properly).

But we're only checking the files that are included in the CSP, so we don't have to worry about those extraneous files, right?

> I'm talking about scenarios where people might install on their own server and likely change the logo, among other small customizations. If you don't want to cater to those scenarios, then it's fine to include them in the checks, and let people judge if/when it's fine if they're not matching the checksums, I assume.

Why would people check their custom builds against the official checksums? If they're making customisations, then of course the checksums won't match - that's the whole point 😊. If people create custom builds and want those to be verifiable by others, they should publish their own checksums.

BrunoBernardino: > But we're only checking the files that are included in the CSP, so we don't have to worry about those extraneous files, right?

We weren't, but we can. These files will also be downloaded by default, so I figured it would make sense to get them, as it would be the "default browser behavior" we talked about trying to emulate before. I'm fine ignoring them.

> Why would people check their custom builds against the official checksums?

If they said "the code is the same, trust me, I just changed the images and fonts", they could actually check that and it would be fine/safe. I'll tweak all of this stuff then.

MaKleSoft: > If they said "the code is the same, trust me, I just changed the images and fonts", they could actually check that and it would be fine/safe.

I don't think that's a use case we want to support. With this stuff it's generally better to err on the side of more strictness.

MaKleSoft: > We weren't, but we can. These files will also be downloaded by default, so I figured it would make sense to get them, as it would be the "default browser behavior" we talked about trying to emulate before. I'm fine ignoring them.

If they're not included in the CSP, they can't/won't be loaded in the context of the PWA, so these are actually irrelevant.

MaKleSoft: > I don't think you can prevent a browser from trying to get robots.txt and favicon.ico, for example, but that's ok.

Pretty sure the CSP will prevent those from loading as well if blocked / not whitelisted.

BrunoBernardino: Not `robots.txt`. `favicon.ico` _maybe_. It doesn't matter here, anyway.

packages/pwa/webpack.config.js
```diff
@@ -66,13 +68,95 @@ module.exports = {
...
+                                        "icon_512x512.e3175643e8fe0d95175a493da5201480.png",
+                                        "icon_384x384.971e45062e4d601a3014dc16ee3ed27b.png",
+                                        "icon_256x256.9a47fba2857d94939047064f37cd075f.png",
+                                        "icon_192x192.8dfb7236c7e6b6591567173b18eaa144.png",
+                                        "icon_128x128.f620784d1682c9fbb033d3b018e7d998.png",
+                                        "icon_96x96.eda9f98be1c35dabab77f9d2ab7be538.png",
+                                    ],
+                                ],
+                                // TODO: This should to be dynamically added, but it's not available at this point
+                                ["manifest-src", ["manifest.623e2268f17398ec7f19225e281e4056.json"]],
```
MaKleSoft, line 102: Have you tried simply adding your plugin **after** all the other plugins?

BrunoBernardino: Good idea. My only bet would be to use the `afterEmit` hook, since the manifest and inject ones use `emit`, _but_ then I'd be stuck having to read the file and write it, while some other plugin might also try to do that. Just very error-prone stuff.

BrunoBernardino:
This has the new CSP parsing script now, and everything seems to be working, except the checkums fail on `index.html` and `sw.js` when comparing with a local build, and fail on that and 4 more (`main.js`, `app_src_elements_app_ts.chunk.js` and both those map files). I'll have to investigate this further tomorrow. Beta has this branch at the moment.

BrunoBernardino:
@MaKleSoft all the checksums are passing now for beta, CI (the packaged thing), and locally. I won't delete the release until you've had a chance to test it as well, if you want.

packages/pwa/webpack.config.js
MaKleSoft: I think we should still try and find a solution for this, since these files **will** change sooner or later, plus anyone trying to customise anything in `assets` will run into problems here. I'm ok with deferring this until later, though. If we do so, let's create an issue for this.

BrunoBernardino: Deferred. Issue is at https://github.com/padloc/padloc/issues/471

MaKleSoft (approved):
Haven't gotten around to retesting this yet, but I trust you've tested it thoroughly and I don't want to block merging this. Merge at your discretion!

PR #469, opened 2022-06-09
Allow hiding/disable directory sync feature via provisioning layer

MaKleSoft:
- Allow hiding/disable directory sync feature via provisioning layer
- [Stripe Provisioner] Enable directory sync only for Business plan


BrunoBernardino (approved):
I didn't pull this down but reviewed the code carefully. Looks good! Nice job.

PR #475, opened 2022-06-20
Fix default TLS Config for PostgreSQL

### LINKED ISSUE #459: Set rejectUnauthorized to true by default for Postgres connections using tls certificates
Current `rejectUnauthorized` is always set to false. I don't recall why I set that flag but I assume it was because there were problems connecting without it. Let's set this to true by default and maybe add a configuration option to set it to false for those who want/need to.
https://github.com/padloc/padloc/blob/v4/packages/server/src/storage/postgres.ts#L49

BrunoBernardino:
Fixes #459

### DIFF
--- a/docs/examples/config/example.env
+++ b/docs/examples/config/example.env
@@ -110,7 +110,7 @@
 # PL_DATA_POSTGRES_PASSWORD=""
 # PL_DATA_POSTGRES_TLS=false
 # PL_DATA_POSTGRES_TLS_CAFILE=""
-# PL_DATA_POSTGRES_TLS_REJECT_UNAUTHORIZED=false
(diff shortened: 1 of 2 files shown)


docs/examples/config/example.env
```diff
@@ -110,6 +110,7 @@
 # PL_DATA_POSTGRES_PASSWORD=""
 # PL_DATA_POSTGRES_TLS=false
 # PL_DATA_POSTGRES_TLS_CAFILE=""
+# PL_DATA_POSTGRES_TLS_REJECT_UNAUTHORIZED=false
```
MaKleSoft, line 113: Minor nitpick: This should match the default (currently `true`)

MaKleSoft (approved):
Looks good (except my one comment, but that's definitely not a blocker)!

PR #477, opened 2022-06-21
Limit input field length for email, name, and org name

### LINKED ISSUE #458: Limit length of organisation names, user display names and emails
These limits should be checked/enforced both on the server side and client side.

BrunoBernardino:
Fixes #458

### DIFF
--- a/packages/app/src/elements/login-signup.ts
+++ b/packages/app/src/elements/login-signup.ts
@@ -460,7 +460,7 @@ export class LoginOrSignup extends StartForm {
                     this._accountExists();
                     return;
                 default:
-                    alert(e.message, { type: "warning" });
+                    alert(e.message || $l("Server error."), { type: "warning" });
                     throw e;
             }
         }
@@ -720,8 +720,9 @@ export class LoginOrSignup extends StartForm {
                             <div class="spacer"></div>
 
                             <div class="hint">
-                                Hi there, <strong>${this._nameInput?.value || "Stranger"}</strong>! Let's set up your
-                                brand new ${process.env.PL_APP_NAME} account! (This will only take a few moments.)
+                                Hi there, <strong class="break-words">${this._nameInput?.value || "Stranger"}</strong>!
+                                Let's set up your brand new ${process.env.PL_APP_NAME} account! (This will only take a
+                                few moments.)
                             </div>
 
                             <pl-input
--- a/packages/app/src/elements/login-signup.ts
+++ b/packages/app/src/elements/login-signup.ts
@@ -25,6 +25,7 @@ import { Confetti } from "./confetti";
 import { singleton } from "../lib/singleton";
 import { PBES2Container } from "@padloc/core/src/container";
 import { importLegacyContainer } from "../lib/import";
+import { ACCOUNT_EMAIL_MAX_LENGTH, ACCOUNT_NAME_MAX_LENGTH } from "@padloc/core/src/account";
 
 @customElement("pl-login-signup")
 export class LoginOrSignup extends StartForm {
@@ -685,7 +686,7 @@ export class LoginOrSignup extends StartForm {
                                     type="email"
                                     required
                                     select-on-focus
-                                    maxlength="255"
+                                    maxlength=${ACCOUNT_EMAIL_MAX_LENGTH}
                                     .label=${$l("Email Address")}
                                     @enter=${() => this._submitEmail()}
                                     ?disabled=${this._page !== "start"}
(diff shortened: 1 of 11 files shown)


packages/app/src/lib/auth/email.ts
```diff
@@ -19,7 +19,9 @@ export class EmailAuthClient implements AuthClient {
         const sent = new Date(sentAt);
         const message = async () => {
             const sentString = await formatDateFromNow(sent);
-            return html` <div>${$l(`Please enter the six digit verification code sent to {0}!`, email)}</div>
+            return html` <div class="break-words">
```
BrunoBernardino, line 22: As I was testing a really long email, I noticed it looked terrible (all in one line and not visible past 50 characters or so). This way it's visible and inside the box.

packages/core/src/server.ts
```diff
@@ -646,11 +646,13 @@ export class Controller extends API {
         authToken,
         verify,
     }: CreateAccountParams): Promise<Account> {
-        // For compatibility with v3 clients, which still use the depracated `verify` property name
+        // For compatibility with v3 clients, which still use the deprecated `verify` property name
         if (verify && !authToken) {
             authToken = verify;
         }
 
+        account.validateOrThrowInputValues();
```
BrunoBernardino, line 654: This will take in consideration the server-side validation, which only happens after the email, name, and master password have been set, whereas the client-side happens right on field submission/keyup.

packages/core/src/util.ts
```diff
@@ -183,7 +183,7 @@ export function truncate(str: string, len: number) {
 export async function getIdFromEmail(email: string) {
     const id = bytesToHex(
         await getCryptoProvider().hash(
-            stringToBytes(email.trim().toLocaleLowerCase()),
+            stringToBytes((email || "").trim().toLocaleLowerCase()),
```
BrunoBernardino, line 186: I ran into this error when trying to delete an account. I didn't investigate further if it could/should be stopped before as it was out of scope. I'm happy to look further into it if you think it's a good idea.

packages/core/src/account.ts
```diff
@@ -227,6 +227,25 @@ export class Account extends PBES2Container implements Storable {
         this._key = account._key;
     }
 
+    validateOrThrowInputValues() {
+        const NAME_MAX_LENGTH = 100;
+        const EMAIL_MAX_LENGTH = 255;
```
BrunoBernardino, line 232: There's no _great_ reason for me to set these values. The numbers just seemed sensible to me, probably from decades-old indoctrination about database limits that are irrelevant now.

packages/core/src/account.ts
```diff
@@ -227,6 +227,25 @@ export class Account extends PBES2Container implements Storable {
         this._key = account._key;
     }
 
+    validateOrThrowInputValues() {
```
MaKleSoft, line 230: `Serializable` actually already has a [`validate`](https://github.com/padloc/padloc/blob/v4/packages/core/src/encoding.ts#L170-L172) method that is automatically called when objects are deserialized. Might as well use that 😉

MaKleSoft (changes requested):
Looks good, except that you should probably use [Serializable::validate](https://github.com/padloc/padloc/blob/v4/packages/core/src/encoding.ts#L170-L172) for the validation logic. That one will be called automatically so you can remove the explicit calls in `server.ts`.

packages/app/src/elements/login-signup.ts
```diff
@@ -460,7 +460,7 @@ export class LoginOrSignup extends StartForm {
                     this._accountExists();
                     return;
                 default:
-                    alert(e.message, { type: "warning" });
+                    alert(e.message || $l("Server error."), { type: "warning" });
```
BrunoBernardino, line 463: I added this default message because an encoding error has no message, and it seemed out of scope to implement a way to return a message on `validate()` (and I'm not even sure if you'd want to). We're not expecting people to hit that anyway, since the client will have validated the data, but still.

packages/app/src/elements/login-signup.ts
```diff
@@ -719,12 +720,14 @@ export class LoginOrSignup extends StartForm {
                             <div class="spacer"></div>
 
                             <div class="hint">
-                                Hi there, <strong>${this._nameInput?.value || "Stranger"}</strong>! Let's set up your
-                                brand new ${process.env.PL_APP_NAME} account! (This will only take a few moments.)
+                                Hi there, <strong class="break-words">${this._nameInput?.value || "Stranger"}</strong>!
```
BrunoBernardino, line 723: People usually have names with spaces, but I figured it wouldn't hurt to keep things looking good here, too.

MaKleSoft (changes requested):
Sorry, one more change 😅

BrunoBernardino:
Alright, here you go!

MaKleSoft (approved):
**Now** I'm happy 😊 Approved!

PR #485, opened 2022-06-30
Fix occasional non-focus extension popup in Firefox

### LINKED ISSUE #482: v4: extension in Firefox sometimes doesn't properly focus
I haven't been able to replicate this exactly since I'm usually in the middle of doing something when I notice it, but it's "visible" because clicking on a field doesn't copy it, and trying to drag also doesn't do anything.

It happens to me maybe once every 20 times I open it, but I just decided to log it since it's been annoying me and my wife recently, and will certainly make the extension seem broken, even if it's just in Firefox (I'll also have to test Chrome).

BrunoBernardino:
Fixes #482

While the solution isn't ideal, there isn't much more to do, other than not using a JS framework to render HTML, as per https://bugzilla.mozilla.org/show_bug.cgi?id=1324255

This also fixes some text on the web extension build, to make it clearer Edge is also supported.

I tested this throughout the day without ever getting the same problem (I did get it with just the `100` timeout), and did not notice any side effects (auto-focus for the password field works fine and if I start typing really fast, also noticed no issues in Firefox nor Chrome).

### DIFF
--- a/.github/workflows/build-web-extension.yml
+++ b/.github/workflows/build-web-extension.yml
@@ -54,7 +54,7 @@ jobs:
                   name: padloc-web-extension-${{ github.sha }}-signed.xpi
                   path: packages/extension/dist/web-ext-artifacts/*.xpi
                   if-no-files-found: error
-            - name: Archive Unsigned Web Extension (Chrome)
+            - name: Archive Unsigned Web Extension (Chrome/Edge)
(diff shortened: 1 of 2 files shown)


BrunoBernardino:
Beta-ready versions of this branch: https://github.com/padloc/padloc/actions/runs/2589520254

PR #486, opened 2022-07-01
Automatically include manifest files in CSP when building PWA

### LINKED ISSUE #471: Automatically add manifest-related files to CSP
Right now they're manually hardcoded because they don't change often, and automating it will require changing the HTML file _after_ it's been generated.

More info: https://github.com/padloc/padloc/pull/467#discussion_r900079941

BrunoBernardino:
Fixes #471

The hardest part here was finding the right hook to tap into, and the silent error thing threw me off for a while as well.

### DIFF
--- a/packages/pwa/webpack.config.js
+++ b/packages/pwa/webpack.config.js
@@ -1,11 +1,12 @@
 const { resolve, join } = require("path");
+const { readFileSync, writeFileSync } = require("fs");
 const { EnvironmentPlugin } = require("webpack");
 const { InjectManifest } = require("workbox-webpack-plugin");
 const HtmlWebpackPlugin = require("html-webpack-plugin");
 const { CleanWebpackPlugin } = require("clean-webpack-plugin");
 const WebpackPwaManifest = require("webpack-pwa-manifest");
-const { version } = require("../../package.json");
 const sharp = require("sharp");
+const { version } = require("../../package.json");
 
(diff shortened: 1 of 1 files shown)


packages/pwa/webpack.config.js
```diff
@@ -70,66 +73,25 @@ module.exports = {
         new CleanWebpackPlugin(),
         {
             apply(compiler) {
-                compiler.hooks.compilation.tap("Store Built Files for CSP", (compilation) => {
+                compiler.hooks.compilation.tap("Update CSP - dev", (compilation) => {
```
BrunoBernardino, line 76: This only runs for dev now (I didn't want to remove CSP entirely, so I just limited to the localhost for somethings, and kept the other limitations as well).

packages/pwa/webpack.config.js
```diff
@@ -191,6 +151,77 @@ module.exports = {
                         size: () => Buffer.byteLength(icon),
                     };
 
+                    return true;
+                });
+            },
+        },
+        {
+            apply(compiler) {
+                compiler.hooks.afterEmit.tapPromise("Store Built Files for CSP - non-dev", async (compilation) => {
```
BrunoBernardino, line 160: And this is for "non-dev", where everything is dynamic now.

MaKleSoft (approved):
I just skimmed the changes, but looks good to me!

PR #487, opened 2022-07-01
Make sure postgres `tlsRejectUnauthorized` setting is parsed as boolean

MaKleSoft:


packages/server/src/storage/postgres.ts
```diff
@@ -28,7 +28,7 @@ export class PostgresConfig extends Config {
     @ConfigParam()
     tlsCAFile?: string;
 
-    @ConfigParam()
+    @ConfigParam("boolean")
```
MaKleSoft, line 31: @BrunoBernardino FYI: Non-string params need to be marked as such, i.e.: "number" for numbers, "boolean" for bools etc. Simple arrays are possible, too via the `string[]` etc.

BrunoBernardino: Oooh, nice to know!

BrunoBernardino (approved):

PR #488, opened 2022-07-04
Add snap and flatpak options to electron publishing and release

### LINKED ISSUE #201: Support of flatpaks (or/and snap)
Hello,

Even if appimage is a nice format, It isn't natively integrated in any major gnu/linux distributions, as Fedora and Ubuntu, using respectively Flatpaks and Snap. Beyond that, you could ship your app on a repo, as flathub. 

I was initially going to give a shot to your app, and this package is very efficient for that, as it don't need installation, not any package manager, but the fact that it would be upgradable only manually, alone in a folder, and not integrated at all in my system isn't really enjoying me to keep it long term.

So the Is there any reason not to try to package it with other universal packages ?

Doc : 
- Flatpak : 
  Packaging : http://docs.flatpak.org/en/latest/first-build.html
  Flathub submission : https://github.com/flathub/flathub/wiki/App-Submission
- Snap : 
  Packaging : https://docs.snapcraft.io/build-snaps/languages
  Snapcraft submission : https://docs.snapcraft.io/build-snaps/publish

BrunoBernardino:
Also adds `deb` and tweaks the v3 compatibility test a bit more since it has failed a couple of times in the last week.

Fixes #201


MaKleSoft (approved):

.github/workflows/build-electron.yml
```diff
@@ -48,13 +48,18 @@ jobs:
...
+                  sudo apt install flatpak flatpak-builder
+                  flatpak remote-add --if-not-exists --user flathub https://flathub.org/repo/flathub.flatpakrepo
             - name: Install dependencies
               run: |
                   npm i -g npm@8.2.0
                   npm ci
             - name: Build
               run: |
                   npm run electron:build
+                  npm run electron:build:flatpak
```
BrunoBernardino, line 62: This had to be a separate process because of the special `appId` requirements.

package.json
```diff
@@ -55,7 +56,7 @@
         "test:e2e": "concurrently --prefix=name --prefix-length=30 --kill-others --success=first -n app,v3-app,maildev,cypress \"PL_DATA_BACKEND=memory PL_DISABLE_SW=true PL_EMAIL_BACKEND=smtp PL_EMAIL_SMTP_HOST=localhost PL_EMAIL_SMTP_PORT=1025 PL_EMAIL_SMTP_IGNORE_TLS=true npm start\" \"npm run start:v3\" \"npx maildev\" \"./node_modules/.bin/wait-on tcp:localhost:8080 && CYPRESS_CRASH_REPORTS=0 cypress run\"",
         "test:e2e:dev": "concurrently --prefix=name --prefix-length=30 --kill-others --success=first -n app,v3-app,cypress \"PL_DATA_BACKEND=memory PL_DISABLE_SW=true PL_EMAIL_BACKEND=smtp PL_EMAIL_SMTP_HOST=localhost PL_EMAIL_SMTP_PORT=1025 PL_EMAIL_SMTP_IGNORE_TLS=true npm run dev\" \"npm run start:v3\" \"npx maildev\" \"./node_modules/.bin/wait-on tcp:localhost:8080 && CYPRESS_CRASH_REPORTS=0 cypress open\"",
         "locale:extract": "lerna run extract --scope '@padloc/locale'",
-        "add": "lerna add $1 --scope=@padloc/$scope",
+        "add": "lerna add $1 --exact --scope=@padloc/$scope",
```
BrunoBernardino, line 59: This just makes sure that whenever we add a dependency, we don't have to remember to pin it to a specific version, it'll already be.

packages/electron/prepare-build.js
```diff
@@ -17,6 +17,7 @@ async function main() {
     const { width } = await baseIcon.metadata();
     const padding = Math.floor(width / 20);
     await baseIcon
+        .resize(512 - padding * 2, 512 - padding * 2, { fit: "inside" }) // Some targets don't deal well with images > 512x512
```
BrunoBernardino, line 20: I _thought_ `.resize(512, 512)` was enough, but the `extend` in the same "pipeline" would make it always increase in size.

packages/electron/prepare-build.js
```diff
@@ -49,12 +50,29 @@ async function main() {
             schemes: [scheme],
         },
         linux: {
+            target: ["AppImage", "snap", "deb", "dir"],
```
BrunoBernardino, line 53: I took the opportunity to also add `deb` here, because why not.

PR #492, opened 2022-07-14
Support better fuzzy-matching, with subdomains

### LINKED ISSUE #397: Improve hostname matching for showing suggested items
Imagine I have an item with the URL of `https://google.com`, for a Google account. When I try to sign in, that suggestion doesn't show, because the URL for logging in (right now) is `https://accounts.google.com`. While it's possible for me to change the URL in the item to match the new subdomain, it's not impossible for Google to eventually change that (to `https://auth.google.com`, for example), and then my items will stop showing as suggested.

While we have a matching logic for `www.{domain}`, we should also consider `*.{domain}`. This is especially bad for Apple accounts, because they frequently change the URL for the authentication form.

@MaKleSoft let me know if this is too complicated or impossible for v4 (not necessarily 4.0.0).

BrunoBernardino:
Fixes #397

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -119,12 +119,12 @@ export class Index extends Serializable {
         return this.items.filter((item) => item.hosts.some((h) => h === hashedHost)).length;
     }
 
-    splitHostBySubdomains(host: string) {
+    getHostnameVariants(host: string) {
         const parts = host.split(".");
 
         // Ignore single domains
         if (parts.length <= 2) {
-            return [];
+            return [host, `*.${host}`];
         }
 
         // Remove the tld and domain from the parts, build it separately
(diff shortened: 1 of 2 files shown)


packages/core/src/app.ts
```diff
@@ -119,11 +119,44 @@ export class Index extends Serializable {
...
+            return [];
+        }
+
+        // Remove the tld and domain from the parts, build it separately
+        const domain: string[] = [];
+
+        domain.unshift(parts.pop()!);
+        domain.unshift(parts.pop()!);
+
+        // Build list of subdomains to match, so given 'login.accounts.google.com', we'd see ['login', 'accounts'] as parts and ['google', 'com'] as domain, which should return ['google.com', 'accounts.google.com', and 'login.accounts.google.com']
```
BrunoBernardino, line 136: This comment is mainly here to help quickly test the lines of code below, but I also think it makes it clearer as to what's happening. I've always disliked how `reduce` makes code less understandable, at least for me, but its' the best option, here, really.

packages/core/src/app.ts
```diff
@@ -119,11 +119,44 @@ export class Index extends Serializable {
...
+                return currentDomainParts;
+            },
+            [domain.join(".")]
+        );
+    }
+
     async fuzzyMatchHost(host: string) {
+        const domains = this.splitHostBySubdomains(host);
+
+        const domainsMatches = (await Promise.all(domains.map(async (domain) => await this.matchHost(domain)))).reduce(
```
BrunoBernardino, line 150: Running this logic below would've been better (not checking if there was a direct host match), but a lot less readable. Your call.

packages/core/src/app.ts
```diff
@@ -119,11 +119,44 @@ export class Index extends Serializable {
...
+            (previousCount, currentCount) => previousCount + currentCount,
+            0
+        );
+
         // Try exact match first, then try to add/remove "www."
         return (
             (await this.matchHost(host)) ||
-            (host.startsWith("www.") ? this.matchHost(host.slice(4)) : this.matchHost("www." + host))
+            (await (host.startsWith("www.") ? this.matchHost(host.slice(4)) : this.matchHost("www." + host))) ||
```
MaKleSoft, line 158: I think we can include that part in the `splitHostBySubdomains` function.

packages/core/src/app.ts
```diff
@@ -119,11 +119,44 @@ export class Index extends Serializable {
         return this.items.filter((item) => item.hosts.some((h) => h === hashedHost)).length;
     }
 
+    splitHostBySubdomains(host: string) {
```
MaKleSoft, line 122: Let's rename this function to something like `getHostnameVariants` and let it return **all** variants, including the actual host name, and the variants with and without `www`. I think that makes it cleaner and easier to understand what's going on.

packages/core/src/app.ts
```diff
@@ -1545,8 +1578,14 @@ export class App {
...
-                        // If host doesn't match exactly, try with/without "www."
-                        return h === host || (host.startsWith("www.") ? host.slice(4) === h : "www." + host === h);
+                        const domainOptions = this.state.index.splitHostBySubdomains(host);
+
+                        // If host doesn't match exactly, try with/without "www." and all subdomain variations
+                        return (
+                            h === host ||
+                            (host.startsWith("www.") ? host.slice(4) === h : "www." + host === h) ||
+                            domainOptions.some((domain) => domain === h)
+                        );
```
MaKleSoft, line 1588: With the suggested change from above (making `splitHostBySubdomains` function return **all** variants), this could be simplified to:

```ts
return this.state.index.getHostnameVariants(host).includes(h)
```

MaKleSoft (changes requested):
As discussed in my comments, I would suggest the following changes:

- Only match subdomain variants if the url value is prefixed with `*.` (I think simply adding `*.` to all subdomain variants should do it). This is important because being too liberal with hostname matching can open us up to phishing attacks.
- Make `splitHostBySubdomains` return **all** hostname variants (sorted by "distance" from the actual hostname) and then match against those instead of treating the exact match and `www` cases separately. Maybe also rename the function accordingly.
- To improve "discoverability" of the wildcard option, include it in the [field suggestions](https://github.com/padloc/padloc/blob/v4/packages/app/src/elements/field.ts#L161)

MaKleSoft (approved):
Thanks for implementing my suggestions! Good to merge 👍

PR #493, opened 2022-07-18
Feature/richtext markdown notes

MaKleSoft:
This PR actually contains a couple of changes:

- New notes editor with support for headings, lists, block quotes markdown support and even a full screen edit mode!
- TOTP and notes fields now require a paid plan in the SaaS version
- Pricing for the family plan is now displayed as the minimum package price (5 seats) to avoid confusion. Additional seats can still be purchased if required.

localhost_8080_items_eda651af-c27f-46e0-bad8-59f1cecefbdd_new_ (1)
localhost_8080_items_f1dc4c22-8f01-41f3-b196-b2e1dc0ccfe2_edit_ (1)
localhost_8080_items_f1dc4c22-8f01-41f3-b196-b2e1dc0ccfe2_


packages/core/src/container.ts
```diff
@@ -117,6 +117,10 @@ export class PBES2Container extends BaseContainer {
             this.keyParams.salt = await getProvider().randomBytes(16);
         }
         this._key = await getProvider().deriveKey(stringToBytes(password), this.keyParams);
+        // If this container has data already, make sure the derived key properly decrypts it.
+        if (this.encryptedData) {
+            await this.getData();
+        }
```
MaKleSoft, line 123: This is a little unrelated fix I snuck in. Basically this just makes sure that an error is thrown if `unlock` is called with an invalid key.

packages/core/src/item.ts
```diff
@@ -193,6 +179,17 @@ export const FIELD_DEFS: { [t in FieldType]: FieldDef } = {
...
+        type: FieldType.Text,
+        pattern: /.*/,
+        matchPattern: /.*/,
+        mask: false,
+        multiline: true,
+        icon: "text",
+        get name() {
+            return $l("Plain Text");
+        },
+    },
```
MaKleSoft, line 192: Renamed the `text` field to "Plain Text" and enabled multi line editing. This basically replaces what the `note` field was previously.

packages/server/src/provisioning/stripe.ts
```diff
@@ -90,6 +90,8 @@ export class StripeProvisioner extends BasicProvisioner {
                 "Shared vaults",
                 "Encrypted file storage",
                 "Security Report",
+                "Built-in Authenticator",
+                "Advanced Note-Taking",
```
MaKleSoft, line 94: If you have a better idea on how to name these features, let me know. For the notes, I've also considered:
- Richtext / Markdown Notes
- Full-featured Notes Editor

From what I've seen elsewhere "Authenticator" is commonly understood to mean "Google Authenticator / Authy type functionality". "Multi-Factor Authentication" is ambiguous because it can also mean MFA for logging into Padloc.

BrunoBernardino (changes requested):
Alright, I reviewed this carefully (but didn't pull down). It seems like there's a failing test that needs to be addressed, and other than that, I only have a couple of minor suggestions that aren't blocking.

While I don't like seeing so many new runtime dependencies, I love the end result, great job!

BrunoBernardino (approved):

PR #500, opened 2022-07-20
Desktop lock on suspend

### LINKED ISSUE #211: OSX desktop app should lock when device goes into sleep mode.
Should use [power-monitor](https://electronjs.org/docs/api/power-monitor) electron module to listen for `suspend` event and lock the app when the device goes to sleep.

BrunoBernardino:
Related to #211

Note that this attempt doesn't quite work as the app doesn't load properly in Electron, and I eventually stopped trying to, since it's turning into a lot of effort. I just wanted it "documented".

There are to options to close #211:

1. When we detect the system is suspending/locking, shutdown/quit the app. This won't require any communication with the PWA.
2. We implement something like [`electron-safe-ipc`](https://www.npmjs.com/package/electron-safe-ipc) directly in the PWA, so electron can send the lock event to it.

Option 1 is not a great UX, but does handle the security aspect.

Option 2 is a much better UX, but introduces a new dependency that is only _relevant_ when the app's used via Electron.

What do you think @MaKleSoft ?

### DIFF
--- a/packages/electron/src/main.ts
+++ b/packages/electron/src/main.ts
@@ -1,10 +1,10 @@
-import { app, BrowserWindow, Menu, dialog, shell } from "electron";
+import { app, BrowserWindow, Menu, dialog, shell, powerMonitor, ipcMain } from "electron";
 import { autoUpdater, UpdateInfo } from "electron-updater";
 // import * as os from "os";
 import ElectronStore from "electron-store";
 
-const debug = process.argv.includes("--dbg");
-const pwaUrl = process.env.PL_PWA_URL!.replace(/(\/*)$/, "");
+const debug = true; //process.argv.includes("--dbg");
+// const pwaUrl = process.env.PL_PWA_URL!.replace(/(\/*)$/, "");
 const appName = process.env.PL_APP_NAME!;
 const appScheme = process.env.PL_APP_SCHEME!;
 
@@ -125,6 +125,8 @@ function createWindow(path: string = "") {
         autoHideMenuBar: true,
         webPreferences: {
             devTools: debug,
+            nodeIntegration: true,
+            contextIsolation: false,
         },
         minWidth,
         minHeight,
@@ -134,8 +136,8 @@ function createWindow(path: string = "") {
         win.webContents.openDevTools();
     }
 
-    // win.loadFile("index.html");
-    win.loadURL(`${pwaUrl}/${path}`);
+    win.loadFile("index.html");
+    // win.loadURL(`${pwaUrl}/${path}`);
 
     win.once("ready-to-show", () => {
         win.show();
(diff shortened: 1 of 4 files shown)


packages/electron/src/main.ts
```diff
@@ -1,10 +1,10 @@
-import { app, BrowserWindow, Menu, dialog, shell } from "electron";
+import { app, BrowserWindow, Menu, dialog, shell, powerMonitor, ipcMain } from "electron";
 import { autoUpdater, UpdateInfo } from "electron-updater";
 // import * as os from "os";
 import ElectronStore from "electron-store";
 
-const debug = process.argv.includes("--dbg");
-const pwaUrl = process.env.PL_PWA_URL!.replace(/(\/*)$/, "");
+const debug = true; //process.argv.includes("--dbg");
```
BrunoBernardino, line 6: I'll need to revert this before merging, it was just an easier way to debug.

packages/electron/src/main.ts
```diff
@@ -308,9 +310,30 @@ async function start() {
...
+        console.log("locking screen");
+        ipcMain.emit("electron-lock-app");
+        window.webContents.send("electron-lock-app");
+    });
+
+    // TODO: Remove this
+    setTimeout(() => {
+        console.log("fake lock");
+        // ipcMain.emit("electron-lock-app");
+        window.webContents.send("electron-lock-app");
```
BrunoBernardino, line 334: There are differences in security between these two, and I'm not sure if the `ipcMain` one is enough, but I'll have to go through that _if_ we decide to implement the `ipcRenderer` stuff in the PWA, otherwise, this process can't communicate with the app.

packages/electron/src/platform.ts
```diff
@@ -1,4 +1,13 @@
...
 import { WebPlatform } from "@padloc/app/src/lib/platform";
 
-export class ElectronPlatform extends WebPlatform implements Platform {}
+export class ElectronPlatform extends WebPlatform implements Platform {
+    constructor() {
+        super();
+
+        ipcRenderer.on("electron-lock-app", () => {
+            window.document.dispatchEvent(new CustomEvent("lock-app"));
+        });
```
BrunoBernardino, line 11: This is what will need to be added to the PWA, basically.

packages/electron/webpack.config.js
```diff
@@ -1,5 +1,6 @@
 const { resolve, join } = require("path");
 const { EnvironmentPlugin } = require("webpack");
+const HtmlWebpackPlugin = require("html-webpack-plugin");
```
BrunoBernardino, line 3: This still wasn't enough to get the app to load, there was error after error, and I just gave up at one point.

MaKleSoft:
> 1. When we detect the system is suspending/locking, shutdown/quit the app. This won't require any communication with the PWA.
> 2. We implement something like [electron-safe-ipc](https://www.npmjs.com/package/electron-safe-ipc) directly in the PWA, so electron can send the lock event to it.

Let's go with option 1. If everything goes as planned, the electron version will be phased out soon, anyway, so I don't want to spend too much time on this.

MaKleSoft (approved):
👍

Issue #346, opened 2021-12-23
Email notifications for account activity


MaKleSoft:
- [x] Login from unknown device
- [x] Failed login attempts

MaKleSoft:
@BrunoBernardino This can be closed now, right?

PR #501, opened 2022-07-21
Email Notifications: Failed Login Attempt

BrunoBernardino:
This is a new feature, related to #346 (half of it).

I'm unsure if we should implement a new server method to be called when the client is already authorized but fails to unlock (a trusted device has been compromised, for example), because that could be helpful but also easy to prevent from being called, so not entirely reliable.


assets/email/failed-login-attempt.txt
```diff
@@ -0,0 +1,7 @@
+Hi there!
+
+This is just an email to warn you that there was a failed attempt to login to your account in Padloc.
+
+If this was you, there's no action necessary, otherwise you might want to make sure your trusted devices haven't been compromised, or to remove potentially compromised devices from your trusted list inside the app.
```
BrunoBernardino, line 5: Not sure if there's anything else we should add here.

packages/app/src/elements/settings-security.ts
```diff
@@ -157,6 +157,12 @@ export class SettingsSecurity extends StateMixin(Routing(LitElement)) {
                 this.renderRoot.querySelector("#securityReportCompromisedToggle") as ToggleButton
             ).active,
         });
+        app.account?.setSettings({
+            failedLoginAttemptNotifications: (
+                this.renderRoot.querySelector("#failedLoginAttemptNotificationsToggle") as ToggleButton
+            ).active,
+        });
+        app.save();
```
BrunoBernardino, line 165: I don't think it's important to `await` here.

packages/app/src/elements/settings-security.ts
```diff
@@ -783,7 +811,8 @@ export class SettingsSecurity extends StateMixin(Routing(LitElement)) {
                         </div>
 
                         ${this._renderBiometricUnlock()} ${this._renderMFA()} ${this._renderSessions()}
-                        ${this._renderTrustedDevices()} ${this._rendersecurityReport()}
+                        ${this._renderTrustedDevices()} ${this._renderSecurityReport()}
+                        ${this._renderEmailNotifications()}
```
BrunoBernardino, line 815: While this is an "account setting" (not "app setting"), I think it still belongs in this settings -> security screen.

packages/app/src/elements/unlock.ts
```diff
@@ -215,6 +215,10 @@ export class Unlock extends StartForm {
             if (e.code !== ErrorCode.DECRYPTION_FAILED) {
                 throw e;
             }
+            if (app.account?.settings.failedLoginAttemptNotifications) {
+                // TODO: Send email?
+                console.log("TODO: Sound send failed login attempt email?");
+            }
```
BrunoBernardino, line 221: This is the main question/doubt I've got. I think this could be helpful to identify trusted devices with valid sessions that could've been compromised, _but_ it's something that happens totally client-side, so would need exposing a new endpoint just to send the email, which means it could be exploited and/or bypassed in those situations.

packages/core/src/account.ts
```diff
@@ -27,6 +27,12 @@ export class AccountSecrets extends Serializable {
     favorites = new Set<VaultItemID>();
 }
 
+/** Various application settings */
+export class AccountSettings extends Serializable {
+    /** Enable emails for failed login attempts */
+    failedLoginAttemptNotifications = true;
```
BrunoBernardino, line 33: I wonder if the `securityReport` `app.settings` should actually be here, now. What do you think @MaKleSoft ?

packages/core/src/messenger.ts
```diff
@@ -60,6 +60,15 @@ export class JoinOrgInviteCompletedMessage extends Message<{ orgName: string; op
     }
 }
 
+export class FailedLoginAttemptMessage extends Message<{ srpId: string }> {
+    template = "failed-login-attempt";
+
+    get title() {
+        const appName = process.env.PL_APP_NAME;
+        return `${appName ? appName + " " : ""}Failed Login Attempt (SRP ID: ${this.data.srpId})`;
```
BrunoBernardino, line 68: Do you think this is helpful at all, or should we just not send any info?

MaKleSoft: I'd leave out the SRP id since it'll only confuse users.

packages/app/src/elements/unlock.ts
MaKleSoft: No, I don't think an email notification is appropriate in this case. If we want to add a failsafe for when a device get's stolen, we could add a client-side setting to automatically log out and remove the device from trusted devices after x failed unlock attempts. When **that** happens we can then send an email notification. Generally email notifications should be triggered by events/conditions triggered at the server side.

packages/core/src/server.ts
```diff
@@ -581,6 +582,12 @@ export class Controller extends API {
         // authentication.
         if (!(await getCryptoProvider().timingSafeEqual(M, srp.M1!))) {
             this.log("account.createSession", { success: false });
+            if (acc.settings.failedLoginAttemptNotifications) {
+                try {
+                    // Send invite link to invitees email address
+                    await this.messenger.send(acc.email, new FailedLoginAttemptMessage({ srpId }));
+                } catch (e) {}
+            }
```
MaKleSoft, line 590: Hm, maybe we should add a number of allowed retries? I.e. only send a notification after x failed login attempts?

packages/core/src/account.ts
MaKleSoft: Yeah, I was actually going to do exactly that when implementing that setting but then decided that it wasn't worth including a new property on `Account` just for that purpose. I think there are a few other settings that should probably persist across devices, like the favicon setting and the "reveal masked fields on hover" setting.

packages/core/src/account.ts
```diff
@@ -165,6 +175,12 @@ export class Account extends PBES2Container implements Storable {
         this.favorites.clear();
     }
 
+    /** Update account settings */
+    async setSettings(obj: Partial<AccountSettings>) {
+        Object.assign(this.settings, obj);
+        return this.settings;
+    }
```
MaKleSoft, line 182: That method seems a bit superfluous. Is there any upside to this as opposed to setting the property directly?

packages/core/src/server.ts
BrunoBernardino: Well, then we need to store that. We do it for the auth request, but at this stage you _know_ something or _have access_ to something (like the email, to get the code). I think it's safe to just send the email here because of it.

packages/core/src/account.ts
BrunoBernardino: Good point, there isn't. I'll remove it.

packages/core/src/account.ts
BrunoBernardino: Well, the favicon and masking might make sense on the extension and not the desktop/mobile, or vice-versa, so I'd keep those app-specific. I'll move the security report ones here, though.

packages/app/src/elements/unlock.ts
BrunoBernardino: Sounds good, but out-of-scope. I'll create a separate issue for that after finishing this PR.

packages/core/src/server.ts
MaKleSoft: We could simply store the number of attempts in the `SRPSession` object, which would be trivial. My concern is that people will often mistype their password at least once or twice before getting it right (I know I do) and then those emails will get annoying very quickly, causing them to disable them. On the other hand an attacker who doesn't get the password right on the first try will probably not get it right on the second or third try, either. So I think it's safe to give them a handful of tries, at least.

MaKleSoft: Actually how about this: Give the user a maximum of `x` tries to get the password right (~5 should be a good number, but we can make that configurable). Once they exceed that number of tries, cancel the SRP session, remove the device from trusted devices and send the email notification. Then the notification could say something along the lines of "We've prevented a login attempt from ..." and then the device model and location.

MaKleSoft: The device model and location can be found in `session.deviceInfo`.

packages/core/src/messenger.ts
MaKleSoft: Maybe add the location to the title. E.g.: `Failed Login Attempt from Munich, Germany`.

assets/email/failed-login-attempt.html
```diff
@@ -0,0 +1,275 @@
...
+                                                        font-family: sans-serif;
+                                                        font-size: 14px;
+                                                        font-weight: normal;
+                                                        margin: 0;
+                                                        margin-bottom: 15px;
+                                                    "
+                                                >
+                                                    This is just an email to warn you that there was a
+                                                    <strong>failed</strong> attempt to login to your account in Padloc,
+                                                    from {{ location }}.
```
BrunoBernardino, line 206: The variable's called `location` but it _can_ include the device info as well. I just didn't want to name it `locationAndPotentiallyDeviceToo`.

packages/app/src/elements/settings-security.ts
```diff
@@ -258,7 +268,7 @@ export class SettingsSecurity extends StateMixin(Routing(LitElement)) {
         ) {
             return;
         }
-        await app.api.revokeSession(id);
+        await app.api.revokeSession({ id });
```
BrunoBernardino, line 271: I wasn't aware [we were "limited" by `process` to only send one argument](https://github.com/padloc/padloc/blob/main/packages/core/src/server.ts#L236-L240), so I had to change this into an object, once I realized that.

packages/app/src/elements/toggle-button.ts
```diff
@@ -13,7 +13,7 @@ export class ToggleButton extends LitElement {
     reverse: boolean = false;
 
     @property()
-    label: string | TemplateResult = "";
+    label: string | TemplateResult | TemplateResult<1> = "";
```
BrunoBernardino, line 16: This just avoided a linting error in the IDE.

packages/app/src/elements/unlock.ts
```diff
@@ -215,6 +215,11 @@ export class Unlock extends StartForm {
             if (e.code !== ErrorCode.DECRYPTION_FAILED) {
                 throw e;
             }
+            if (app.account?.settings.failedLoginAttemptNotifications && this._failedCount > 3) {
```
BrunoBernardino, line 218: The number is `3`, but since it starts with 0 and is only incremented _after_  the check, this _actually_ happens at the 5th failed attempt. I considered changing the whole code in this function for the numbers to add up, but deemed it out of scope.

packages/core/src/messenger.ts
```diff
@@ -60,6 +60,15 @@ export class JoinOrgInviteCompletedMessage extends Message<{ orgName: string; op
     }
 }
 
+export class FailedLoginAttemptMessage extends Message<{ location: string }> {
+    template = "failed-login-attempt";
+
+    get title() {
+        const appName = process.env.PL_APP_NAME;
+        return `${appName ? appName + " " : ""}Failed Login Attempt from ${this.data.location})`;
```
BrunoBernardino, line 68: I think that changing this to "we prevented a ..." would sound too alarming, but am open to doing it.

BrunoBernardino:
@MaKleSoft all's implemented now, basically we will send the email:
1. If the non-authorized person's logging in from a new device and fails 5 times (after they either guessed or intercepted the authorization code);
2. If the non-authorized person's logging in from a trusted (or logged in but locked) device and fails 5 times.

This covers half of #346 because what's still missing is sending a different email _after_ logging in successfully, from a new, untrusted device.

BrunoBernardino:
@MaKleSoft I've fixed the conflicts here, if you could please review it! :heart:

packages/core/src/api.ts
```diff
@@ -486,7 +486,7 @@ export class API {
      * Revoke a [[Session]], effectively logging out any client authenticated with it
      */
     @Handler(String, undefined)
-    revokeSession(_id: SessionID): PromiseWithProgress<void> {
+    revokeSession(_params: { id: SessionID; revokedForFailedAttempts?: boolean }): PromiseWithProgress<void> {
```
MaKleSoft, line 489: If you you want to use anything other than a primitive value as an argument, you'll have to define a parameter class for that and adjust the `@Handler` decorator accordingly. See https://github.com/padloc/padloc/blob/052477b6c908092742d602316073b5bb6f53cf5c/packages/core/src/api.ts#L449-L452 for an example.

packages/app/src/elements/settings-security.ts
```diff
@@ -151,12 +151,22 @@ export class SettingsSecurity extends StateMixin(Routing(LitElement)) {
...
+                this.renderRoot.querySelector("#securityReportReusedToggle") as ToggleButton
+            ).active;
+            app.account.settings.securityReportCompromised = (
+                this.renderRoot.querySelector("#securityReportCompromisedToggle") as ToggleButton
+            ).active;
+            app.account.settings.failedLoginAttemptNotifications = (
+                this.renderRoot.querySelector("#failedLoginAttemptNotificationsToggle") as ToggleButton
+            ).active;
+            app.save();
+        }
```
MaKleSoft, line 169: You need to use `app.updateAccount` for this:
```ts
await app.updateAccount(acc => {
            app.account.settings.securityReportWeak = (
                this.renderRoot.querySelector("#securityReportWeakToggle") as ToggleButton
            ).active;
            app.account.settings.securityReportReused = (
                this.renderRoot.querySelector("#securityReportReusedToggle") as ToggleButton
            ).active;
            app.account.settings.securityReportCompromised = (
                this.renderRoot.querySelector("#securityReportCompromisedToggle") as ToggleButton
            ).active;
            app.account.settings.failedLoginAttemptNotifications = (
                this.renderRoot.querySelector("#failedLoginAttemptNotificationsToggle") as ToggleButton
            ).active;
});
```

packages/app/src/elements/toggle-button.ts
MaKleSoft: Honestly I'd rather figure out **why** the type checker was complaining and fix the underlying problem instead. There should be only one type definition for `TemplateResult`. If there are more than one then something is off.

packages/app/src/elements/unlock.ts
MaKleSoft: I think this is a completely distinct feature from failed login notifications and should probably be part of a different PR. Aside from that, we should introduce a separate setting for this, since it really doesn't have anything to do with failed **login** attempts (we're talking about **unlocking** here, and the behaviour goes beyond just a notification). I'd say let's defer this feature to a different PR, which also means the change to the `revokeSession` endpoint is unnecessary.

packages/core/src/account.ts
```diff
@@ -27,6 +27,18 @@ export class AccountSecrets extends Serializable {
...
+    /** Enable checking for weak passwords */
+    securityReportWeak = true;
+    /** Enable checking for reused passwords */
+    securityReportReused = true;
+    /** Enable checking for compromised passwords */
+    securityReportCompromised = true;
+    /** Enable emails for failed login attempts */
+    failedLoginAttemptNotifications = true;
+}
+
```
MaKleSoft, line 41: Just a thought, but do you think it might make sense to group these into subproperties/classes? Something like

```ts
export class SecurityReportSettings extends Serializable {
    weakPasswords = true;
    reusedPasswords = true;
    compromisedPaswords = true;
}

export class NotificationSettings extends Serializable {
    failedLoginAttempts = true;
}

export class AccountSettings extends Serializable {
    @AsSerializable(SecurityReportSettings)
    securityReport = new SecurityReportSettings();

    @AsSerializable(NotificationSettings)
    notifications = new NotificationSettings();
}
```
Yes it's a bit more boilerplate but would keep things a bit tidier which would be nice especially considering that there will probably a bunch more settings in the future.

packages/core/src/server.ts
```diff
@@ -639,6 +669,20 @@ export class Controller extends API {
...
+
+                // Send invite link to invitees email address
+                await this.messenger.send(account.email, new FailedLoginAttemptMessage({ location }));
+
+                // Remove trusted device, if it was (after email as this can throw if the device was already removed)
+                if (session.info.device) {
+                    await this.removeTrustedDevice(session.info.device.id);
+                }
+            } catch (e) {}
+        }
```
MaKleSoft, line 685: Again, I think this is conflating failed login attempts with failed unlock attempts. I know a lot of people don't really understand the difference but it's important for us to make a clear distinction here. I would feel better if we'd do this as part of another PR.

MaKleSoft (changes requested):
As discussed in my comments, I think we shouldn't conflate failed login attempts with failed unlock attempts. Dealing with the latter is also a good idea, but requires a completely different approach and therefore should be done in a separate PR, at least if we want to merge this one in a timely manner. I also think it would be nice to give the `AccountSettings` class more "structure", but I'm open to discussion there.

A different thing that would actually make more sense to include here and would also be much easier to add would be notifications for successful logins on new/untrusted devices. In fact that kind of notification is probably just as important as failed login attempts and could easily be disabled if it get's annoying. Adding it would probably be only a few lines of code, but we can also do that in a different PR.

packages/app/src/elements/settings-security.ts
```diff
@@ -151,11 +151,23 @@ export class SettingsSecurity extends StateMixin(Routing(LitElement)) {
         app.setSettings({
             autoLock: (this.renderRoot.querySelector("#autoLockButton") as ToggleButton).active,
             autoLockDelay: (this.renderRoot.querySelector("#autoLockDelaySlider") as Slider).value,
-            securityReportWeak: (this.renderRoot.querySelector("#securityReportWeakToggle") as ToggleButton).active,
-            securityReportReused: (this.renderRoot.querySelector("#securityReportReusedToggle") as ToggleButton).active,
-            securityReportCompromised: (
+        });
+        app.updateAccount(async (account) => {
```
MaKleSoft, line 155: I think you forgot an `await` here?

packages/app/src/elements/settings-security.ts
```diff
@@ -689,11 +701,11 @@ export class SettingsSecurity extends StateMixin(Routing(LitElement)) {
                     <pl-toggle-button
                         class="transparent"
                         id="securityReportWeakToggle"
-                        .active=${app.settings.securityReportWeak}
+                        .active=${app.account?.settings.securityReport.weakPasswords || false}
                         .label=${html`<div class="horizontal center-aligning spacing layout">
                             <pl-icon icon="weak"></pl-icon>
                             <div>${$l("Weak Passwords")}</div>
-                        </div>`}
+                        </div>` as TemplateResult}
```
MaKleSoft, line 708: Again, why is this cast necessary (it shouldn't be). I can also debug this for you if you're stuck.

packages/app/src/elements/unlock.ts
```diff
@@ -215,6 +215,11 @@ export class Unlock extends StartForm {
             if (e.code !== ErrorCode.DECRYPTION_FAILED) {
                 throw e;
             }
+            if (this._failedCount > 3) {
+                await this.app.logout();
+                router.go("login");
+                return;
+            }
```
MaKleSoft, line 222: Keeping the logout after 5 attempts is a great idea, since it's a simple change. I would add an alert though to inform people of what is happening.

packages/core/src/server.ts
```diff
@@ -602,13 +632,21 @@ export class Controller extends API {
...
+            }
+
+            // Send new login notification (it's a new or untrusted device)
+            if (acc.settings.notifications.newLogins) {
+                try {
+                    const location = this._buildLocationAndDeviceString(this.context.location, this.context.device);
+
+                    await this.messenger.send(acc.email, new NewLoginMessage({ location }));
+                } catch (e) {}
+            }
```
MaKleSoft, line 649: I think we can/should do this asynchronously, since it's not critical if it fails and waiting for the message to be sent will add a significant delay.

packages/core/src/server.ts
```diff
@@ -582,6 +598,20 @@ export class Controller extends API {
...
+            if (srpState.failedAttempts >= 5 && acc.settings.notifications.failedLoginAttempts) {
+                try {
+                    const location = this._buildLocationAndDeviceString(this.context.location, this.context.device);
+
+                    await this.messenger.send(acc.email, new FailedLoginAttemptMessage({ location }));
+
+                    // Remove trusted device, if it was (after email as this can throw if the device was already removed)
+                    if (this.context.device) {
+                        await this.removeTrustedDevice(this.context.device.id);
+                    }
```
MaKleSoft, line 612: This should be outside of the conditional block, since we should probably do this regardless of the notification setting. We can also add a separate setting for this but I think simply doing this as a default security precaution is totally justifiable and doesn't necessarily need to be disable-able.

assets/email/new-login.txt
```diff
@@ -0,0 +1,7 @@
+Hi there!
+
+This is just an email to warn you that there was a new successful login to your account in Padloc, from {{ location }}.
+
+If this was you, there's no action necessary, otherwise you might want to make sure your trusted devices haven't been compromised, or to remove potentially compromised devices from your trusted list inside the app.
```
MaKleSoft, line 5: You might also want to add a note about the device in question having been removed from the list of trusted devices.

packages/core/src/server.ts
```diff
@@ -582,6 +598,20 @@ export class Controller extends API {
         // authentication.
         if (!(await getCryptoProvider().timingSafeEqual(M, srp.M1!))) {
             this.log("account.createSession", { success: false });
+            ++srpState.failedAttempts;
+            await this.storage.save(auth);
+            if (srpState.failedAttempts >= 5 && acc.settings.notifications.failedLoginAttempts) {
+                try {
+                    const location = this._buildLocationAndDeviceString(this.context.location, this.context.device);
+
+                    await this.messenger.send(acc.email, new FailedLoginAttemptMessage({ location }));
```
MaKleSoft, line 607: Again, I don't think we need to wait for this.

MaKleSoft (changes requested):
Had a couple more comments. Sorry ;)

packages/app/src/elements/settings-security.ts
BrunoBernardino: Because it has a variable inside, it's reported as `TemplateResult<1 | 2>` instead of `TemplateResult`. The app doesn't break without this, but `tslint` complains.

MaKleSoft: > Because it has a variable inside, it's reported as TemplateResult<1 | 2> instead of TemplateResult

Not sure what you mean by that. Is there any documentation around this?

BrunoBernardino: I think the problem _might_ be related to https://github.com/runem/lit-analyzer/issues/154

This is what I'm talking about:

Screenshot from 2022-08-02 07-31-02

MaKleSoft: Okay, let's table this for now, then. Maybe it'll get fixed upstream.

assets/email/failed-login-attempt.txt
```diff
@@ -0,0 +1,9 @@
+Hi there!
+
+This is just an email to warn you that there was a failed attempt to login to your account in Padloc, from {{ location }}.
+
+If this was you, there's no action necessary, otherwise you might want to make sure your trusted devices haven't been compromised, or to remove potentially compromised devices from your trusted list inside the app.
+
+Note that if this happened on a trusted device, it was already removed from the trusted devices list, automatically.
+
+This email was sent to you by Padloc (https://padloc.app). If you have any questions, please don't hesitate to contact us at support@padloc.app!
```
MaKleSoft, line 9: Just a minor thing, but can we add a

> Have a great day!
> 
> Your Padloc Team

packages/app/src/elements/settings-security.ts
```diff
@@ -730,6 +742,42 @@ export class SettingsSecurity extends StateMixin(Routing(LitElement)) {
...
+            <div class="box">
+                <h2 class="padded uppercase bg-dark border-bottom semibold">${$l("Email Notifications")}</h2>
+
+                <div>
+                    <pl-toggle-button
+                        class="transparent"
+                        id="failedLoginAttemptsNotificationsToggle"
+                        .active=${app.account?.settings.notifications.failedLoginAttempts || false}
+                        .label=${html`<div class="horizontal center-aligning spacing layout">
+                            <pl-icon icon="weak"></pl-icon>
```
MaKleSoft, line 756: Can we find a better icon for this please? 🙃 The envelope one should be fine if you don't want to get creative.

BrunoBernardino:
@MaKleSoft ready for review again, and BTW, with the new login email notification I added yesterday, this will fully close #346

packages/core/src/server.ts
```diff
@@ -582,6 +598,23 @@ export class Controller extends API {
...
+                }
+
+                if (acc.settings.notifications.failedLoginAttempts) {
+                    try {
+                        const location = this._buildLocationAndDeviceString(this.context.location, this.context.device);
+
+                        this.messenger.send(acc.email, new FailedLoginAttemptMessage({ location }));
+                    } catch (e) {}
+                }
+            }
```
MaKleSoft, line 617: While testing this I noticed that you can keep trying indefinitely after 5 failed attempts and it will sent a notification for every single attempt. Instead, we should cancel the SRP session (by deleting it) so they have to at least go through the the email verification again before they can try again. You might need to add some code on the client to handle this though.

MaKleSoft (changes requested):

packages/core/src/server.ts
BrunoBernardino: I assume we'll delete the SRP session _even_ if we don't send the email, right?

packages/app/src/elements/settings-security.ts
BrunoBernardino: Oops, sorry! I forgot about that.

assets/email/failed-login-attempt.txt
BrunoBernardino: I intentionally removed that because this didn't seem like a "feel good" email, but I can add it back.

packages/core/src/server.ts
MaKleSoft: Yes, correct.

assets/email/failed-login-attempt.txt
MaKleSoft: True. Just "Your Padloc Team" would be fine too I guess.

MaKleSoft:
@BrunoBernardino Oh, almost forgot: I guess now would be a good time to start updating the change log along with PRs :)

MaKleSoft:
@BrunoBernardino Btw, I haven't forgotten about this PR. Just too many other things to deal with right now. I'll try to get to it later this week.

MaKleSoft (approved):
Ok, finally got around to properly reviewing and testing this. I did find a problem with the account settings not being saved on the server side which I fixed and made a couple of other minor changes. Should be good merge now!

PR #502, opened 2022-07-25
Update support page with proper content and styling

MaKleSoft:
Last minute addition before the v4.0.0 release. Totally forgot to add proper links to the support page 🤦‍♂️


BrunoBernardino (approved):
I didn't pull down, but the code looks good! Nice catch!

PR #505, opened 2022-07-26
Allow selectively disabling all billing functionality on certain platforms

MaKleSoft:
Title says it all. This will allow us to disable all billing-related functionality and references on certain platforms. For people on the free plan, using premium features will simply display a message saying the feature is not available on their platform yet and to use the web app instead. Once they've upgraded to premium, those features will be available on those platforms as well (but without a way to upgrade/downgrade etc).

### DIFF
--- a/packages/tauri/src/platform.ts
+++ b/packages/tauri/src/platform.ts
@@ -1,7 +1,10 @@
-import { Platform } from "@padloc/core/src/platform";
+import { DeviceInfo, Platform } from "@padloc/core/src/platform";
 import { WebPlatform } from "@padloc/app/src/lib/platform";
-// import { MemoryStorage } from "@padloc/core/src/storage";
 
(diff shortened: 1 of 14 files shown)


packages/tauri/src/platform.ts
```diff
@@ -1,7 +1,10 @@
-import { Platform } from "@padloc/core/src/platform";
+import { DeviceInfo, Platform } from "@padloc/core/src/platform";
```
BrunoBernardino, line 1: This is causing a linting error because it's not used.

BrunoBernardino (changes requested):
Only one minor thing that's causing the tauri builds to fail. Other than that, the code looks good (I didn't pull it down).

BrunoBernardino (approved):
Perfect!

PR #506, opened 2022-07-26
More helpful error reports via email

MaKleSoft:
This adds some additional information to the error reports sent by email, which should make debugging problems easier.

I also forgot disabling the billing page in the last PR, so included that here.


BrunoBernardino (approved):
Changes look good!

PR #511, opened 2022-07-28
Add flag to completely disable multi-factor authentication for a given account

MaKleSoft:
We especially need this specifically for apple reviewers, since they sometimes insist on using an existing account and getting the mfa code to them is a pita. Might be useful for other purposes, too. Disabled by default obviously.


BrunoBernardino (approved):
Looks good!

PR #513, opened 2022-07-28
Force-logout older clients and reject any future requests

MaKleSoft:
This will effectively disable access for all clients < 4.0.0. Not happy about doing this but it's necessary to prevent further issues.

### DIFF
--- a/packages/core/src/server.ts
+++ b/packages/core/src/server.ts
@@ -52,7 +52,7 @@ import {
 } from "./messenger";
 import { Server as SRPServer, SRPSession } from "./srp";
 import { DeviceInfo, getCryptoProvider } from "./platform";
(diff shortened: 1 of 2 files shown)


BrunoBernardino (changes requested):
It looks like we need to fix a couple of tests about the legacy UI and server, but looks good apart from that!

BrunoBernardino:
@MaKleSoft let me know if you'd like me to fix this.

MaKleSoft:
@BrunoBernardino I decided to table this for now. Other than I though originally, use of the v3 client was not responsible for any data loss, so I think this would cause more trouble than it would prevent at this point. It might still make sense to do this later, though, so let's just keep this open for now.

PR #515, opened 2022-07-29
Provide better information in error reports...

MaKleSoft:
... by printing the stack trace and message of the original error instead of the `Err` wrapper


MaKleSoft:
Also added a postgres logger so we have an easy way to log events in production

packages/core/src/util.ts
```diff
@@ -21,6 +21,18 @@ export async function uuid(): Promise<string> {
...
+ * NOT CRYPTOGRAPHICALLY SAFE!
+ */
+export function unsafeUUID(): string {
+    return "xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx".replace(/[xy]/g, function (c) {
+        var r = (Math.random() * 16) | 0,
+            v = c == "x" ? r : (r & 0x3) | 0x8;
+        return v.toString(16);
+    });
+}
+
```
MaKleSoft, line 35: I was using this initially for the event ids before I decided to got with time stamp + random number instead (for faster sorting). Figured this might be useful in the future so I kept it around

PR #521, opened 2022-08-02
Update SimpeWebAuthn dependency to latest

MaKleSoft:


BrunoBernardino:
I looked at the changelog in https://github.com/MasterKale/SimpleWebAuthn/blob/master/CHANGELOG.md and it seems we shouldn't be affected by the breaking changes there.

PR #526, opened 2022-08-03
Fix problem with OrgProvisioning::owner property

MaKleSoft:
While looking through the database, I noticed that in the `orgprovisioning` table, the `owner` property was stored as a string instead of a json object. Turns out the problem was that [here](https://github.com/padloc/padloc/blob/main/packages/core/src/provisioning.ts#L424) the owner property was set to a reference of the `OrgMember` class instead of a plain object, which caused problems during serialisation. This PR should fix that.

### DIFF
--- a/packages/core/src/org.ts
+++ b/packages/core/src/org.ts
@@ -309,17 +309,20 @@ export class Org extends SharedContainer implements Storable {
 
     /** [[Account]] which created this organization */
     get owner() {
-        return this.members.find((member) => member.role === OrgRole.Owner);
+        const owner = this.members.find((member) => member.role === OrgRole.Owner);
+        return owner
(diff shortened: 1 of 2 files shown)


BrunoBernardino (approved):
Nice catch! I didn't pull this down as it seems simple enough.

Do we need to run some script to cleanup data? Or is the auto-fixing of serializing/updating enough?

MaKleSoft:
> Do we need to run some script to cleanup data? Or is the auto-fixing of serializing/updating enough?

Already created an issue :) https://github.com/padloc/devops/issues/15

MaKleSoft:
@BrunoBernardino FYI: Fix is deployed

PR #530, opened 2022-08-05
Add missing translations check

### LINKED ISSUE #523: Add check for missing translation entries
It would be nice to have a test that checks whether there are any new translatable strings in the app that haven't been added to the translation files yet. Would probably be as simple as:

1. Run `npm run extract` inside `packages/locale`
2. Check if `packages/locale/res/translations/_template.json` has changed.
3. If it has, fail the test.

@BrunoBernardino Not a priority, but if you think you can knock it out quickly feel free to do it right away.

BrunoBernardino:
Also send params for the authenticator endpoint when it fails, so we can try to see if there's any helpful information there (related to the Windows errors mentioned in https://github.com/MasterKale/SimpleWebAuthn/issues/238 )

Fixes #523


packages/locale/package.json
```diff
@@ -8,7 +8,7 @@
         "res"
     ],
     "scripts": {
-        "extract": "ts-node src/extract.ts ../*/src/{**,.}/*.ts"
+        "extract": "ts-node src/extract.ts `ls ../*/src/**/*.ts`"
```
MaKleSoft, line 11: Is the `ls` required? I though simply fixing the glob would do it.

MaKleSoft:
Looks like the order of the strings has changes, which is probably due to the changes in how the files are listed. Going forward we should probably order the strings alphabetically to avoid random changes in the order.

BrunoBernardino:
Looks like a lot of translations are missing with the "fixed" glob.

packages/locale/res/translations/de.json
```diff
@@ -2005,23 +2013,23 @@
   ],
   [
     "Date",
-    "Datum"
```
BrunoBernardino, line 2008: hmm @MaKleSoft why would it remove the existing translations?

BrunoBernardino: Was just because locally they had been removed. I just brought them back from `main` and re-ran `npm run locale:extract` (no need to run it from the package directory), and it's all fine now.

BrunoBernardino:
Alright @MaKleSoft sorry, finally ready for a review.

MaKleSoft (approved):

MaKleSoft:
Good to merge from my side! Maybe consider sorting the translation strings alphabetically to guarantee consistent ordering and making it easier to find a given string.

BrunoBernardino:
@MaKleSoft I never had that problem again, so I think we'll be fine for a while.

PR #531, opened 2022-08-05
Make release assets consistent

BrunoBernardino:
Unfortunately [it's not yet possible for Tauri](https://github.com/tauri-apps/tauri-action/issues/215), nor there's any GH action for that, _but_ [there's an API endpoint](https://docs.github.com/en/rest/releases/assets#update-a-release-asset) we could use.

Since it's only the Tauri assets that are not according to the new standard, I think it's fine to keep it something we manually update, otherwise we can try to build a script that taps into that endpoint, later.

### DIFF
--- a/packages/electron/prepare-build.js
+++ b/packages/electron/prepare-build.js
@@ -33,13 +33,13 @@ async function main() {
     const buildConfig = {
         appId,
         buildVersion,
-        artifactName: "${name}_${version}_${os}_electron_${arch}.${ext}",
+        artifactName: `${name.toLowerCase()}_\${version}_\${os}_electron_\${arch}.\${ext}`,
         directories: {
             app: "app",
             buildResources: "build",
         },
         mac: {
-            artifactName: "${name}_${version}_macos_electron_${arch}.${ext}",
+            artifactName: `${name.toLowerCase()}_\${version}_macos_electron_\${arch}.\${ext}`,
             hardenedRuntime: true,
             darkModeSupport: true,
             gatekeeperAssess: false,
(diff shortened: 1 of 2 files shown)


packages/electron/prepare-build.js
```diff
@@ -33,11 +33,13 @@ async function main() {
     const buildConfig = {
         appId,
         buildVersion,
+        artifactName: "${name}_${version}_${os}_electron_${arch}.${ext}",
         directories: {
             app: "app",
             buildResources: "build",
         },
         mac: {
+            artifactName: "${name}_${version}_macos_electron_${arch}.${ext}",
```
BrunoBernardino, line 42: `${os}` for `mac` is `mac` and not `macos`, which we're using in the assets, thus the need for this platform-specific `artifactName`.

BrunoBernardino:
Release is being built at https://github.com/padloc/padloc/actions/runs/2803282247 and currently visible at https://github.com/padloc/padloc/releases/tag/v4.0.1 (but I'll remove that in a bit)

BrunoBernardino:
I didn't want to update `tauri.conf.json`'s version as this will be merged to `main`, but seems like it generates 10 files that we'll need to manually update to match the naming system. I'll leave the release up (it was built against Beta, not Production), in case you want to test it, and I'll delete it and the tag after this is merged.

MaKleSoft (commented):
So you're saying we still have to do this manually for the Tauri releases? Would it not be possible to rename the files in the CI script after the build is completed? Or would that be too complicated?

w3bdesign (approved):

BrunoBernardino:
@MaKleSoft I'm saying we'll have to do this manually for the Tauri releases. I could add another script in CI to fix that, but it's not trivial (probably a day or two to get it right), and I'd rather do it in a separate PR, if you think it's worth it.

MaKleSoft (approved):

MaKleSoft:
Ok, I can live with that :)

Issue #507, opened 2022-07-26
Fullscreen notes editing is broken on iOS


MaKleSoft:
Main problem is that the "safe area" behind the notch is not respected, but it get's worse due to the focus behavior on iOS, which slide the whole screen up a bit. This makes it impossible to leave fullscreen edit mode.

PR #534, opened 2022-08-07
Improve fullscreen notes on iOS

MaKleSoft:
- Added appropriate padding to top of fullscreen notes editor to accommodate for the notch on iPhone.
- Switched from [ionic-plugin-keyboard](https://github.com/ionic-team/cordova-plugin-ionic-keyboard) to [cordova-plugin-keyboard](https://github.com/cjpearson/cordova-plugin-keyboard) because of https://github.com/ionic-team/cordova-plugin-ionic-keyboard/issues/35. Both are unmaintained, but cordova-plugin-keyboard seems to be doing the job nicely!
- Disabled `HideKeyboardFormAccessoryBar` setting. I had enabled this originally for aesthetic reasons but I now think that was silly, especially since without the accessory bar there is no way to close the keyboard on iOS without clicking outside an editable area (which can sometimes be tricky).

Fixes https://github.com/padloc/padloc/issues/507


packages/cordova/package.json
```diff
@@ -56,14 +56,14 @@
...
         "es6-promise-plugin": "4.2.2",
-        "ionic-plugin-keyboard": "2.2.1",
         "typescript": "4.4.3"
     },
     "devDependencies": {
         "@types/cordova-plugin-qrscanner": "1.0.31",
         "clean-webpack-plugin": "3.0.0",
         "cordova": "11.0.0",
         "cordova-android": "10.1.2",
+        "cordova-plugin-keyboard": "^1.2.0",
```
BrunoBernardino, line 66: Can you please fix this version? 
```suggestion
        "cordova-plugin-keyboard": "1.2.0",
```

Also, shouldn't this have been a normal dependency instead of a dev one?

packages/cordova/config.xml
```diff
@@ -84,8 +82,8 @@
...
             </array>
         </config-file>
+        <splash src="res/ios/splash/Default@3x~universal~anyany.png"/>
     </platform>
-    <plugin name="cordova-plugin-app-version" spec="~0.1.9" />
-    <plugin name="cordova-plugin-backbutton" spec="~0.3.0" />
-    <plugin name="ionic-plugin-keyboard" spec="~2.2.1" />
-</widget>
+    <plugin name="cordova-plugin-app-version" spec="~0.1.9"/>
+    <plugin name="cordova-plugin-backbutton" spec="~0.3.0"/>
```
BrunoBernardino, line 88: Shouldn't `cordova-plugin-keyboard` be here?

BrunoBernardino (commented):
Nice work! I've added some questions since I'm not sure if they're accurate or not, as I didn't pull this down to test.

packages/cordova/config.xml
MaKleSoft: Good question. I'm not sure if it's need but we can add it just to be sure. Would you take care of that please?

BrunoBernardino: Since there are many other plugins that don't show here and I assume this already worked, I'd rather not add a change without being able to test it through again.

BrunoBernardino (approved):

Issue #517, opened 2022-07-30
Add better device descriptions for Tauri apps


MaKleSoft:
Currently the Tauri app shows up as "Edge on Windows" (or something similar, depending on the platform). We should try to get a better device description. For reference, here is how it's done for the Cordova apps:  https://github.com/padloc/padloc/blob/main/packages/cordova/src/platform.ts#L51

escapedraccoon:
Maybe also allow for changing the description, making it easier to distinguish between different installations on the Tauri platform (additionally keeping the standard-setting in brackets* to avoid malicious access renaming their app to something familiar)?

*(or at least show it somewhere)

MaKleSoft:
@escapedraccoon Not a bad idea! I'll create a new ticket for it, but I can't promise when (or if) we'll implement this. But if enough people ask for it we'll consider it.

PR #536, opened 2022-08-07
Generate better device descriptions for Tauri runtime

MaKleSoft:
Fixes https://github.com/padloc/padloc/issues/517

Unfortunately we can't do better that "Windows Device", "MacOS Device" etc. right now because we don't have access to the computer name yet. Specifically having an equivalent of Nodes `os.hostname` would be helpful, which is planned but not ready yet. See https://github.com/tauri-apps/tauri/issues/2233


BrunoBernardino (approved):
Nice and simple.

Issue #532, opened 2022-08-05
Keep previous options when editing in Android


Terkyz:
I'm not sure if this also happens in the Web/Desktop version, but if I try to edit two things at once without saving the previous change, automatically the first change is deleted

Example: When editing the name of my Vault, I don't save and I try to edit the mail, directly the name of the Vault will be deleted and I will have to edit it again

BrunoBernardino:
@MaKleSoft is this a limitation of how we're using `lit`? I've experienced this before as well. If I want to add say 3 new fields but already have data in, I have to add and save each one individually, which isn't a very good UX. I've also lost data like @Terkyz mentions here, by editing something, then trying to edit something else on the same item before saving.

MaKleSoft:
I can reproduce this, but **only** by clicking on the `+ Add Field` button while in edit mode. @MaKleSoft @Terkyz can you confirm this?

BrunoBernardino:
That's usually what happens to me, yeah. I can edit anything that already exists, but adding or _removing_ a field will revert everything else (or uploading a file).

MaKleSoft:
Hm, removing fields or uploading files does not result in the same behavior. Only clicking on `+ Add Field` (but not clicking on the `+` button in the top right corner.

BrunoBernardino:
The top right corner only seems to cause the issue when creating a new item, I believe.

MaKleSoft:
@BrunoBernardino Can you check if this PR fixes your issues? https://github.com/padloc/padloc/pull/538

PR #538, opened 2022-08-08
Various fixes and improvements

MaKleSoft:
See commit comments. (Hopefully) fixes https://github.com/padloc/padloc/issues/532


BrunoBernardino (approved):
This looks great! I've pulled it down and tested it too. It definitely fixes #532

PR #540, opened 2022-08-08
FEATURE: add missing german strings & improve translations

crydotsnake:


MaKleSoft:
Hi @crydotsnake, thank you so much for your contribution! Before I start reviewing, could you please update the translation file to be consistent with the original formatting? It's two spaces for indentation, but your file has a mix of 2 and 4 spaces. I'm assuming this is due to auto formatting in your editor?

BrunoBernardino:
@crydotsnake Note you should be able to do it by simply running `npm run prettier` from the root of the repo.

crydotsnake:
> @crydotsnake Note you should be able to do it by simply running `npm run prettier` from the root of the repo.

`npm run prettier` did nothing. I also run `npm run format:check` before i commit my changes and ther was no error message about bad formating.

BrunoBernardino:
Oops, sorry about that. It seems we indeed ignore the translations in `.prettierignore`. Should we add those in @MaKleSoft ?

BrunoBernardino:
Sorry @crydotsnake I would definitely recommend you revert that. I suppose we don't have prettier there because of how it tweaks the code makes it harder for a non-technical person to contribute a translation. Could you simply make your translations consistent with the 4 space-indentation instead? That should make things easier. The conflict is about a translation that was removed, by the way.

MaKleSoft:
@crydotsnake @BrunoBernardino Sorry for the confusion... yes, translation files are excluded from prettier, because prettier uses inconsistent formatting for shorter translation strings, which makes the file less readable imo. Basically the idea is to let the extraction script take care of the formatting and the simply fill in the translations inside of the quotes **without making any changes to the formatting**.

MaKleSoft:
@crydotsnake Did you close the PR by mistake? We can take care of fixing the formatting for you if you want - after all, you've already done most of the work!

PR #550, opened 2022-08-10
Update tw.json

rhuba8324:
Changed Notes, Note, Security Reports to more accurate words.

I misunderstood the meaning for these lines, now fixed with more accurate and understandable words.


packages/locale/res/translations/tw.json
```diff
@@ -2155,4 +2155,4 @@
     "Custom",
     "自訂"
   ]
-]
\ No newline at end of file
+]
```
BrunoBernardino, line 2158: @rhuba8324 can you try not adding this EOL, or running `npm run locale:extract` from the root of the repo to automatically fix it? It's fine if you can't, I'll just have to fix it after merging.

rhuba8324: It's now fixed, at least I think so.
Sorry for my mistake.

rhuba8324:
EOL fixed, sorry for my mistake.
Still learning about these things.

packages/locale/res/translations/tw.json
BrunoBernardino: No worries, not something you should know _a priori_!

PR #557, opened 2022-08-13
Fix typos in `_template.json`

halacoglu:
- `organisation` > `organization` to keep consistency
- Possessive apostrophe
- `synching` > `syncing` as the latter is more commonly used
- Other little fixes

PR #559, opened 2022-08-16
Fix typos

BrunoBernardino:
Found via #557 (thanks @halacoglu!)

### DIFF
--- a/packages/app/src/mixins/error-handling.ts
+++ b/packages/app/src/mixins/error-handling.ts
@@ -66,8 +66,7 @@ export function ErrorHandling<B extends Constructor<Object>>(baseClass: B) {
                     await alert(
                         error.message ||
                             $l(
-                                "Some data associated with your account was saved with a newer version of " +
-                                    "Padloc and cannot be decoded. Please install the latest version Padloc!"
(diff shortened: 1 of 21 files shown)


packages/app/src/mixins/error-handling.ts
```diff
@@ -66,8 +66,7 @@ export function ErrorHandling<B extends Constructor<Object>>(baseClass: B) {
                     await alert(
                         error.message ||
                             $l(
-                                "Some data associated with your account was saved with a newer version of " +
-                                    "Padloc and cannot be decoded. Please install the latest version Padloc!"
+                                `Some data associated with your account was saved with a newer version of Padloc and cannot be decoded. Please install the latest version of Padloc!`
```
MaKleSoft, line 69: I'm not sure if I mentioned this before, but if you use template literals inside the `translate` function the extraction script won't be able to find them. So either we keep using regular string literals or we update the parser: https://github.com/padloc/padloc/blob/main/packages/locale/src/extract.ts#L7-L17

BrunoBernardino: I'm working on updating the parser if it's not hard, though first I'm bringing back the lost translations because of it (including the ones with the fixed typos).

Issue #433, opened 2022-04-20
Audit Logs


MaKleSoft:
At this point it's not 100% clear what this would mean specifically, but it's a feature frequently requested by large organisations. Let's figure out to what degree this is possible considering Padloc's architecture and security/privacy constraints.

PR #560, opened 2022-08-17
Basic Audit Logs

BrunoBernardino:
Initial setup and work, a few questions still need answering, and only working for some auth-related actions right now.

This is a very rough note on some things I wanted to achieve/remember/attempt with this:

- separate database and potentially database connection, logs table
    - table/audit log per data type:
      - auth
        - auth_id
        - modifier_account_id
        - date
        - type (created, updated, deleted)
        - old_data (encrypted)
        - new_data (encrypted)
        - (account_id can be found via old_data -> account_id or new_data -> account_id)
      - account
        - account_id
        - modifier_account_id
        - date
        - type (created, updated, deleted)
        - old_data (encrypted)
        - new_data (encrypted)
        - (auth_id can be found via old_data -> auth_id or new_data -> auth_id)
      - org
        - org_id
        - modifier_account_id
        - date
        - type (created, updated, deleted)
        - old_data (encrypted)
        - new_data (encrypted)
        - (auth_id or account_id can be found via old_data -> account_id or new_data -> account_id)
      - vault
        - vault_id
        - modifier_account_id
        - date
        - type (created, updated, deleted)
        - old_data (encrypted)
        - new_data (encrypted)
        - (org_id or account_id can be found via old_data -> account_id or new_data -> account_id)
      - item
        - item_id
        - modifier_account_id
        - date
        - type (created, updated, deleted)
        - old_data (encrypted)
        - new_data (encrypted)
        - (vault_id or org_id or account_id can be found via old_data -> account_id or new_data -> account_id)
- [x] store when changes happen
- [x] new class: auditor, new Auditor, server.auditor.log(kind, id, modifier_account_id, date, type, old_data, new_data)
- [x] when a password changes, decrypt and encrypt all audit logs for that account

This is related to #433 but in order to close that, we'll need a way to display these logs, which I'd prefer to do in a separate PR, but would be amenable to implementing it here too.

### DIFF
--- a/packages/app/src/elements/item-icon.ts
+++ b/packages/app/src/elements/item-icon.ts
@@ -48,7 +48,10 @@ export class ItemIcon extends LitElement {
             } catch (e) {}
         }
 
-        const favIcon = app.settings.favicons && url ? `https://icons.duckduckgo.com/ip3/${url}.ico` : undefined;
+        const faviconUrl = url?.startsWith("*.") ? url.replace("*.", "") : url;
+
+        const favIcon =
+            app.settings.favicons && faviconUrl ? `https://icons.duckduckgo.com/ip3/${faviconUrl}.ico` : undefined;
 
         // const dataUrl = favIcon && (await this._loadAsDataUrl(favIcon));
         // console.log(dataUrl);
--- a/packages/core/src/auditor.ts
+++ b/packages/core/src/auditor.ts
@@ -1,0 +1,42 @@
+import { Storable } from "./storage";
+
+export type AuditKind = "auth" | "account" | "org" | "vault" | "item";
+export type AuditChangeType = "created" | "updated" | "deleted";
+
+export class LogChange extends Storable {
+    id: string = "";
+
+    date: Date = new Date();
+
+    get kind(): string {
+        return `audit_log_${this.objectKind}`;
+    }
+
+    constructor(
+        public objectKind: AuditKind,
+        public objectId: string,
+        public modifierAccountId: string,
+        public type: AuditChangeType,
+        public oldData: any,
+        public newData: any
+    ) {
+        super();
+    }
+}
+
+export interface Auditor {
+    log(
+        kind: AuditKind,
+        id: string,
+        modifierAccountId: string,
+        type: AuditChangeType,
+        oldData: any,
+        newData: any
+    ): LogChange;
+}
+
+export class VoidAuditor implements Auditor {
+    log(kind: AuditKind, id: string, modifierAccountId: string, type: AuditChangeType, oldData: any, newData: any) {
+        return new LogChange(kind, id, modifierAccountId, type, oldData, newData);
+    }
+}
--- a/packages/core/src/auditor.ts
+++ b/packages/core/src/auditor.ts
@@ -17,26 +17,18 @@ export class LogChange extends Storable {
         public objectId: string,
         public modifierAccountId: string,
         public type: AuditChangeType,
-        public oldData: any,
-        public newData: any
+        public metadata?: any
     ) {
         super();
     }
 }
 
 export interface Auditor {
-    log(
-        kind: AuditKind,
-        id: string,
-        modifierAccountId: string,
-        type: AuditChangeType,
-        oldData: any,
-        newData: any
-    ): LogChange;
+    log(kind: AuditKind, id: string, modifierAccountId: string, type: AuditChangeType, metadata?: any): LogChange;
 }
 
 export class VoidAuditor implements Auditor {
-    log(kind: AuditKind, id: string, modifierAccountId: string, type: AuditChangeType, oldData: any, newData: any) {
-        return new LogChange(kind, id, modifierAccountId, type, oldData, newData);
+    log(kind: AuditKind, id: string, modifierAccountId: string, type: AuditChangeType, metadata?: any) {
+        return new LogChange(kind, id, modifierAccountId, type, metadata);
     }
 }
--- a/packages/core/src/directory.ts
+++ b/packages/core/src/directory.ts
@@ -57,7 +57,7 @@ export class DirectorySync implements DirectorySubscriber {
 
             org.members.push(newOrgMember);
 
-            await this.server.updateMetaData(org);
+            await this.server.updateMetaData(org, "directory-sync");
             await this.server.storage.save(org);
         }
     }
@@ -89,7 +89,7 @@ export class DirectorySync implements DirectorySubscriber {
 
             existingMember.updated = new Date();
 
-            await this.server.updateMetaData(org);
+            await this.server.updateMetaData(org, "directory-sync");
             await this.server.storage.save(org);
         }
     }
@@ -119,7 +119,7 @@ export class DirectorySync implements DirectorySubscriber {
                 }
             }
 
-            await this.server.updateMetaData(org);
+            await this.server.updateMetaData(org, "directory-sync");
             await this.server.storage.save(org);
         }
     }
@@ -149,7 +149,7 @@ export class DirectorySync implements DirectorySubscriber {
 
             org.groups.push(newGroup);
 
-            await this.server.updateMetaData(org);
+            await this.server.updateMetaData(org, "directory-sync");
             await this.server.storage.save(org);
         }
     }
@@ -180,7 +180,7 @@ export class DirectorySync implements DirectorySubscriber {
 
             existingGroup.members = members;
 
-            await this.server.updateMetaData(org);
+            await this.server.updateMetaData(org, "directory-sync");
             await this.server.storage.save(org);
         }
     }
@@ -194,7 +194,7 @@ export class DirectorySync implements DirectorySubscriber {
             }
             org.groups.splice(existingGroupIndex, 1);
 
-            await this.server.updateMetaData(org);
+            await this.server.updateMetaData(org, "directory-sync");
             await this.server.storage.save(org);
         }
     }
--- a/packages/core/src/server.ts
+++ b/packages/core/src/server.ts
@@ -52,7 +52,7 @@ import {
 } from "./messenger";
 import { Server as SRPServer, SRPSession } from "./srp";
 import { DeviceInfo, getCryptoProvider } from "./platform";
-import { getIdFromEmail, uuid, removeTrailingSlash } from "./util";
+import { getIdFromEmail, uuid, removeTrailingSlash, deepClone } from "./util";
 import { loadLanguage } from "@padloc/locale/src/translate";
 import { Logger, VoidLogger } from "./logging";
 import { PBES2Container } from "./container";
@@ -60,6 +60,7 @@ import { KeyStoreEntry } from "./key-store";
 import { Config, ConfigParam } from "./config";
 import { Provisioner, Provisioning, ProvisioningStatus, StubProvisioner } from "./provisioning";
 import { V3Compat } from "./v3-compat";
+import { Auditor, VoidAuditor } from "./auditor";
 
 /** Server configuration */
 export class ServerConfig extends Config {
@@ -136,6 +137,10 @@ export class Controller extends API {
 
     get storage() {
         return this.server.storage;
+    }
+
+    get auditor() {
+        return this.server.auditor;
     }
 
     get messenger() {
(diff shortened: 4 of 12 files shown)


packages/core/src/auditor.ts
```diff
@@ -0,0 +1,42 @@
+import { Storable } from "./storage";
+
+export type AuditKind = "auth" | "account" | "org" | "vault" | "item";
+export type AuditChangeType = "created" | "updated" | "deleted";
```
BrunoBernardino, line 4: Anything critical you think I missed for this initial version, @MaKleSoft ?

packages/core/src/server.ts
```diff
@@ -252,10 +257,12 @@ export class Controller extends API {
         const { auth } = this._requireAuth();
         const authenticator = new Authenticator({ type, purposes, device });
         await authenticator.init();
+        const originalAuth = deepClone(auth);
         const provider = this._getAuthServer(type);
         const responseData = await provider.initAuthenticator(authenticator, auth, data);
         auth.authenticators.push(authenticator);
         await this.storage.save(auth);
+        this.auditor.log("auth", auth.id, auth.accountId!, "updated", originalAuth, auth);
```
BrunoBernardino, line 265: So, we can do this, or, an alternative that's not uncommon, would be to pass the `auditor` to the `storage`, and every time we run `storage.save()`, we'd try to run `auditor.log` (knowing that some changes don't need to be individually audited -- sessions and provisioning, for example), but I wasn't sure if you liked that.

packages/server/src/config.ts
```diff
@@ -97,6 +97,22 @@ export class LoggingConfig extends Config {
...
+    @ConfigParam()
+    backend: "void" | "mongodb" | "postgres" = "void";
+
+    @ConfigParam(MongoDBStorageConfig)
+    mongodb?: MongoDBStorageConfig;
+
+    @ConfigParam(PostgresConfig)
+    postgres?: PostgresConfig;
+}
+
```
BrunoBernardino, line 115: Anything critical missing here for a v1 of the audit logs?

packages/core/src/auditor.ts
MaKleSoft: Instead of restricting ourselves to specific CRUD actions, I think we should go for something more generic, analogous to what we have in [the logging module](https://github.com/padloc/padloc/blob/main/packages/core/src/logging.ts#L16-L24). In fact, now that I think of it, the logging module already pretty much does what we're trying to do, no? All we need is an endpoint to send events from the client (for those events that we can't catch on the server, like editing/viewing items, plus some logic around storing encrypted data in events.

MaKleSoft:
Ok, after thinking about this for a bit, I do think that instead of reinventing the wheel, we should instead extends the scope of [`Logger`](https://github.com/padloc/padloc/blob/main/packages/core/src/logging.ts#L37). With the current implementation, we're already getting a lot of events for free: https://docs.google.com/spreadsheets/d/1Sma3Nf5IRdS5Vl-hUSMhFUxJYnwgPOMrqB6zBlBKvaQ/edit#gid=1870207771

Really the only things left to do are:

1. Create a server endpoint for events that can't be caught by the server (like viewing/editing items):
    ```ts
    @Handler(LogEvent, undefined)
    startRegisterAuthenticator(
        event: LogEvent
    ): PromiseWithProgress<void> {
        ...
    }
    ```
2. Add tracking for client side events. E.g.: For editing an item:
    ```ts
    app.api.log(new LogEvent("vaultitem.edit", {
        account: {
            name: app.account.name,
            email: app.account.email,
            id: app.account.id
        },
        time: new Date().toISOString(),
        item: {
            name: this._item.name,
            id: this._item.id
        },
        vault: {
            name: this._vault.name,
            id: this._vault.id
        }
        ...
    }));
    ```
3. Display audit logs in admin portal

packages/core/src/auditor.ts
BrunoBernardino: You can see in my latest commit that I tried the endpoint approach and it was becoming a problem.

MaKleSoft: Ok, so maybe we should first define what those need are. I think we should have a brief call about this.

BrunoBernardino:
Thanks for taking a look at this @MaKleSoft !

> 1. Create a server endpoint for events that can't be caught by the server (like viewing/editing items):

I've tried this and it has a few problems:

1. Less reliable
2. Opens an opportunity for exploitation (fake logging)
3. I'm not sure we want to log viewing individual items. I understand some enterprises might want to have that information, but it can be easily abused and I'm not comfortable doing anything more than logging Create/Update/Delete here (since we do for auth, we know when someone logs in, for example, as a way to know if the app's being used/compromised)

> 2. Add tracking for client side events. E.g.: For editing an item: 

Check my latest commit message on that, similar to number one. We can easily infer (even if it's not cheap or the fastest) changes to items and log those on the server, without having to "trust the client" to tell us that information.

> 3. Display audit logs in admin portal

I was thinking we'd have a trail for this for the app itself. Yes, there's a good use case for better filtering and whatnot in the admin portal, but what I wasn't sure was if the displaying in the app itself would be done in this PR or a separate one.

MaKleSoft:
> I'm not sure we want to log viewing individual items. I understand some enterprises might want to have that information, but it can be easily abused and I'm not comfortable doing anything more than logging Create/Update/Delete here (since we do for auth, we know when someone logs in, for example, as a way to know if the app's being used/compromised)

> I was thinking we'd have a trail for this for the app itself. Yes, there's a good use case for better filtering and whatnot in the admin portal, but what I wasn't sure was if the displaying in the app itself would be done in this PR or a separate one.

Audit logs are purely an enterprise feature, and would be useless if they didn't include things like viewing and editing items (which I believe it the main think customers will want to check on). Regarding your concerns about privacy/abuse: This will **not** be used in any form in the enterprise offering, and is strictly opt-in.

> Check my latest commit message on that, similar to number one. We can easily infer (even if it's not cheap or the fastest) changes to items and log those on the server, without having to "trust the client" to tell us that information.

How? Anything encrypted inside vaults is completely opaque to us, so we have zero chance of detecting which items have been accessed, changed or deleted.

BrunoBernardino:
> Audit logs are purely an enterprise feature, and would be useless if they didn't include things like viewing and editing items (which I believe is the main thing customers will want to check on). Regarding your concerns about privacy/abuse: This will not be used in any form in our SaaS offering, and is strictly opt-in.

I'm not saying we shouldn't include editing, but I'm very concerned about viewing. This falls into very-easy-to-infer-real-data-kind of metadata. I know it won't be used in our SaaS offering, but I'm still concerned about abuses in enterprise context.

> How? Anything encrypted inside vaults is completely opaque to us, so we have zero chance of detecting which items have been accessed, changed or deleted.

Welp. I was going to run into that tomorrow, I forgot we encrypt the entire vault, and not items individually, so I can't see if something is different in a given item. Seems I'll have to build in that API endpoint and frontend logic anyway (I still have the backend piece cached in my IDE, will push that up, so I can continue tomorrow).

MaKleSoft:
> I'm not saying we shouldn't include editing, but I'm very concerned about viewing. This falls into very-easy-to-infer-real-data-kind of metadata. I know it won't be used in our SaaS offering, but I'm still concerned about abuses in enterprise context.

I fully agree with you about the privacy implications and potential for abuse, but the fact is that many (if not most) large enterprises **need** this level of tracking for compliance reasons, and by not offering it we'd be effectively cutting ourselves off from that market. Let me emphasise that we'll **never** make this level of tracking the default, nor will it ever by used in our SaaS product, but the reality is we need to give enterprise customers the choice, or they'll simply go for a different solution.

packages/app/src/elements/item-icon.ts
```diff
@@ -48,7 +48,10 @@ export class ItemIcon extends LitElement {
             } catch (e) {}
         }
 
-        const favIcon = app.settings.favicons && url ? `https://icons.duckduckgo.com/ip3/${url}.ico` : undefined;
+        const faviconUrl = url?.startsWith("*.") ? url.replace("*.", "") : url;
```
BrunoBernardino, line 51: This is an unrelated fix, for URLs with the new `*.` format.

packages/core/src/directory.ts
```diff
@@ -57,7 +57,7 @@ export class DirectorySync implements DirectorySubscriber {
 
             org.members.push(newOrgMember);
 
-            await this.server.updateMetaData(org);
+            await this.server.updateMetaData(org, "directory-sync");
```
BrunoBernardino, line 60: There's no real `accountId` making this change, so we use this special string. Seemed fine as it is, but I can make it a constant if you prefer.

packages/core/src/server.ts
```diff
@@ -1489,18 +1573,22 @@ export class Controller extends API {
...
         await this.storage.save(org);
 
+        this.auditor.log("org", org.id, account.id, "updated", { accountId: account.id });
+
         // Delete vault
         await this.storage.delete(vault);
 
         this.log("vault.delete", {
             vault: { id: vault.id, name: vault.name, owner: vault.owner },
-            org: (org && { id: org.id, name: org.name, owner: org.owner }) || undefined,
```
BrunoBernardino, line 1502: This was unnecessary (probably old copy+paste) because we're already tapping into the `org` before.

packages/core/src/server.ts
```diff
@@ -1708,8 +1798,8 @@ export class Controller extends API {
         this.log("account.deleteLegacyAccount");
     }
 
-    updateMetaData(org: Org) {
-        return this.server.updateMetaData(org);
+    updateMetaData(org: Org, modifierAccountId: string) {
```
BrunoBernardino, line 1801: This was necessary so we could pass on whoever was making a modification.

packages/core/src/server.ts
```diff
@@ -1773,11 +1865,28 @@ export class Controller extends API {
...
             keyStoreEntry: { id: entry.id },
         });
     }
 
+    async itemAuditorLog({ itemId, auditChangeType, metadata }: ItemAuditorLogParams): Promise<void> {
+        const { account } = this._requireAuth();
+
+        // TODO: Remove this
+        console.log("======== server.itemAuditorLog");
+        console.log(JSON.stringify({ itemId, auditChangeType, metadata }, null, 2));
```
BrunoBernardino, line 1880: So, I don't know what I'm missing in the deserialization, but it's always an empty object (it looks like a properly serialized object being sent in the request). Help @MaKleSoft

packages/core/src/util.ts
```diff
@@ -243,3 +243,23 @@ export function setPath(obj: any, path: string, value: any) {
...
     }
 }
+
+// Simple (as in a plain object), not shallow
+export function simpleClone(object: any) {
+    return JSON.parse(JSON.stringify(object));
+}
+
+// Copies everything recursively, including class, getters and setters
+export function deepClone(object: any) {
```
BrunoBernardino, line 253: I ended up not needing these two, so if you'd prefer, I can delete them.

BrunoBernardino:
@MaKleSoft this is ready for a review (just pending the deserialization issue for the item logs, which will hopefully be something quick and then there'll be nothing left "to do" here).

packages/app/src/elements/item-icon.ts
MaKleSoft: Can we move this outside of this PR, please? I think it may be a bit longer before we can merge this one.

MaKleSoft:
@BrunoBernardino So, I've reviewed your PR carefully and thought a lot about what we talked about, and in the end decided to go for a different approach. I still haven't worked out all the details, but roughly it will be something like this:

- Audit logs will use [`Logger`](https://github.com/padloc/padloc/blob/main/packages/core/src/logging.ts#L37-L39) as the underlying logging mechanism, instead of implementing something new. The reason for this is that even though the purpose of audit logs is different from "regular" logs, the underlying requirements and challenges do have a large overlap, and it feels like using two different implementations would duplicate a lot of effort.
- Since we discussed that audit logs will be logging CRUD operations exclusively, it makes sense to have them be handled by the storage layer (I think you had the same idea at one point). This will require a lot less code and make it significantly more reliable, since it will work automatically for all storables and there is less chance to miss a spot where logs should be written. The way that I plan to implement this is to allow (optionally) passing a `Logger` instance to certain storage backends. I'll also have to figure out how to pass the context without complicating the storage api, but I think I already have a good idea on how to do that.
- Audit log events will live in a single table. This makes querying/filtering etc. a lot easier and removes a lot of unnecessary complexity.
- In a way, audit logs can be seen as a specialised version of the existing logging mechanism. However, even though they use most of the same code, they'll be stored and displayed separately.
- I've decided to exclude CRUD events for vault items form audit logs for now. We **may** include them in the "regular" logs though.

I'll be implementing my own version of audit logs as part of the [initial admin portal implementation](https://github.com/padloc/padloc/pull/570). I haven't started on the audit logs yet, but you can already see some of the work I've been doing on the `logging` module to solve some of the challenges related to the points above.

Sorry for not merging this implementation! Even though we didn't end up using the code, it certainly helped nail down the requirements and challenges and gave me a couple of good ideas on how to implement this in a more robust way. 

You mentioned that were were a couple of unrelated fixed in this PR. Can you separate those out and put them into a new PR please? Thanks!

BrunoBernardino:
@MaKleSoft Roger. I'm assuming you got clarification and clearance from the enterprise clients asking for this implementation for these changes.

PR #564, opened 2022-08-22
Add expiring or expired items to Security Report

### LINKED ISSUE #345: Expiration Dates / Renewal intervals for Vault Items

BrunoBernardino:
Closes #345

Still WIP, I need to figure out the right UX for making this as hands-off as possible.

On the "renewal" aspect, my idea is to check when an item's updated, if it has the `AuditType.ExpiredItem` in `auditResults`, and `expiryFrequency` is not `manual`, ask for a confirmation if the `expiresBy` date should be updated, and if so, automatically update with the given frequency option.

When editing, I'm planning to show the button as I have for automatically adding 1 year (see discussion below), and an option for the renewal (manual, trimester, semester, year).

@MaKleSoft let me know your thoughts.


packages/app/src/elements/item-view.ts
```diff
@@ -226,6 +236,28 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
...
+    private _removeExpiration() {
+        if (!this._item) {
+            return;
+        }
+
+        this._item.expiresBy = undefined;
+        this._item.expiryFrequency = undefined;
+
+        this.save();
+    }
```
BrunoBernardino, line 259: @MaKleSoft I made this to test out the report quickly, but I think there might be something to it, on making it as easy as clicking a button to have an item expire yearly, which is the most common frequency. Let me know your thoughts.

packages/app/src/lib/audit.ts
```diff
@@ -51,6 +51,16 @@ async function hasPasswordBeenCompromised(passwordHash: string) {
     return false;
 }
 
+async function isItemExpiringOrExpired(expiryDate: Date) {
```
MaKleSoft, line 54: Why is this function async?

packages/app/src/lib/audit.ts
```diff
@@ -51,6 +51,16 @@ async function hasPasswordBeenCompromised(passwordHash: string) {
...
+    const thirtyDaysFromNow = add(new Date(), { days: 30 });
```
MaKleSoft, line 55: I assume this is for testing?

packages/core/src/app.ts
```diff
@@ -1520,6 +1520,8 @@ export class App {
             attachments?: AttachmentInfo[];
             auditResults?: AuditResult[];
             lastAudited?: Date;
+            expiresBy?: Date;
+            expiryFrequency?: ExpiryFrequencyOption;
```
MaKleSoft, line 1524: Better use a generic interval, like number of days. In the UI there can still be a list of predefined intervals but in the data layer there is no reason to restrict ourselves.

packages/core/src/app.ts
```diff
@@ -1520,6 +1520,8 @@ export class App {
             attachments?: AttachmentInfo[];
             auditResults?: AuditResult[];
             lastAudited?: Date;
+            expiresBy?: Date;
```
MaKleSoft, line 1523: *Very* minor nitpick: `expiryDate` would be slightly more consistent with `expiryFrequency`.

packages/core/src/item.ts
```diff
@@ -269,13 +269,21 @@ export enum AuditType {
...
     fieldIndex: number;
 }
 
+export enum ExpiryFrequencyOption {
+    Manual = "manual",
+    Trimestrially = "trimestrially",
+    Semestrially = "semestrially",
+    Yearly = "yearly",
+}
+
```
MaKleSoft, line 286: Those options seem weird to me. I think more sensible defaults would be `weekly`, `monthly`, `quarterly`, `yearly`. But since `monthly` and `quarterly` and `yearly` can't be accurately expressed in days, it's probably better to use `7 days`, `30 days`, `90 days`, `365 days`.

MaKleSoft:
> On the "renewal" aspect, my idea is to check when an item's updated, if it has the AuditType.ExpiredItem in auditResults, and expiryFrequency is not manual, ask for a confirmation if the expiresBy date should be updated, and if so, automatically update with the given frequency option.

Yep, sounds good to me. Except I would also ask if the `expiryFrequency` is set to "manual" (or is undefined), but ask if the expiration date should be removed instead.

MaKleSoft:
> > On the "renewal" aspect, my idea is to check when an item's updated, if it has the AuditType.ExpiredItem in auditResults, and expiryFrequency is not manual, ask for a confirmation if the expiresBy date should be updated, and if so, automatically update with the given frequency option.
>
> Yep, sounds good to me. Except I would also ask if the expiryFrequency is set to "manual" (or is undefined), but ask if the expiration date should be removed instead.

Actually, if we're asking after each update, I'm wondering if we need the `expiryInterval` property at all. We could also just always give the following options:

- Remove expiration
- Add new expiration (1 Year from now)
- Add new expiration (30 days from now)
- Add new expiration (7 days from now)
- Add new expiration (custom)

Would be the same number of clicks (one) and make the whole process and implementation a bit simpler.

MaKleSoft:
> When editing, I'm planning to show the button as I have for automatically adding 1 year (see discussion below), and an option for the renewal (manual, trimester, semester, year).

I like the section and button as they are currently, but instead of "Have this item expire next year" I would have the button simply say "Add Expiration Date". Upon clicking the button, we'd show a popover menu with the following options:

- 1 Year from now
- 30 days from now
- 7 days from now
- custom date

MaKleSoft:
Once an expiration date has been added, we simply show a date input in edit mode so they can adjust the date manually if they want.

packages/app/src/lib/audit.ts
BrunoBernardino: Oops, typo.

packages/app/src/lib/audit.ts
BrunoBernardino: Not really, I thought it made sense to warn in the UI when an item is expiring too, so people know it needs to be acted on, rather than waiting for it to be "past" its date.

packages/core/src/app.ts
BrunoBernardino: That's a good idea. I'll implement that tomorrow.

packages/core/src/app.ts
BrunoBernardino: True, but I don't like how it reads or sounds. :sweat_smile:

packages/core/src/item.ts
BrunoBernardino: I've never seen rotation policies shorter than 3 months, and even those were for very extreme cases. The most common is yearly, but I'm happy to go with 30, 90, 120, 365 days as options in the UI, instead.

BrunoBernardino:
> Except I would also ask if the `expiryFrequency` is set to "manual" (or is undefined), but ask if the expiration date should be removed instead.

@MaKleSoft could you elaborate? I don't understand why we'd ask to remove the expiration date.

> Actually, if we're asking after each update, I'm wondering if we need the `expiryInterval` property at all.

We only ask _when_ it's automatic. I wouldn't ask when it's manual, otherwise it can get quite annoying, though it's not super common for items to be constantly updated, I guess. It's also less of a cognitive load to say "yes/no" than showing so many options.

> I like the section and button as they are currently, but instead of "Have this item expire next year" I would have the button simply say "Add Expiration Date". Upon clicking the button, we'd show a popover menu with the following options:

Roger that!

> Once an expiration date has been added, we simply show a date input in edit mode so they can adjust the date manually if they want.

Check out what I just pushed. I added a "formatted" date like "Expires in 6 days. Renews Yearly." which switches to the input + dropdown choice when editing.

MaKleSoft:
> We only ask when it's automatic. I wouldn't ask when it's manual, otherwise it can get quite annoying, though it's not super common for items to be constantly updated, I guess. It's also less of a cognitive load to say "yes/no" than showing so many options.

Yeah I don't think items aren't updated that frequently, and items with expiration dates even more rarely. So I wouldn't worry about the popup getting annoying.

MaKleSoft:
>  It's also less of a cognitive load to say "yes/no" than showing so many options.

I would argue the opposite. Having a popup that simply gives you the options "Yes" and "No" puts a burden on the user to figure out what those options actually represent, meaning they'll have to read the prompt, which as we both now people rarely do. Having the choices spelled out explicitly makes it more obvious what they are being asked for and they can't just click through without thinking. I guarantee you if it's just standard yes/no prompt a lot of users will simply think it's just a "are you sure you want to save" kind of question and will simply click "Yes" without thinking.

I'm honestly not sure if the "automatic renewal" makes that much sense in practice, since the user still has to confirm every time they update the item. That's why I'm saying let's just let the choose every time. As a compromise, we can also simply "remember" the last option and highlight that as the default option in the "renewal" prompt.

MaKleSoft:
> > Except I would also ask if the expiryFrequency is set to "manual" (or is undefined), but ask if the expiration date should be removed instead.
> 
> @MaKleSoft could you elaborate? I don't understand why we'd ask to remove the expiration date.

Well if the item has been updated, that means that the previous expiration date likely isn't accurate anymore, even if it's in the future. But like I said above, I think the distinction between "manual" and "automatic" is unnecessarily confusing and I'd rather just get rid of the "automatic" option altogehter.

packages/app/src/elements/item-view.ts
```diff
@@ -563,7 +674,7 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
                     class="animated padded spacing evenly stretching horizontal layout save-cancel"
                     ?hidden=${!this._editing}
                 >
-                    <pl-button class="primary spacing horizontal layout" @click=${this.save}>
+                    <pl-button class="primary spacing horizontal layout" @click=${() => this.save()}>
```
BrunoBernardino, line 677: This was sending an object `{ isTrusted: true }`, but it didn't seem necessary or used, and it messed with my new param.

packages/app/src/elements/item-view.ts
```diff
@@ -601,16 +712,28 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
         }
     }
 
-    save() {
+    async save(skipExpiresPrompt = false) {
         if (!this._nameInput.reportValidity()) {
             return;
         }
+
+        if (!skipExpiresPrompt && this._item!.expiresAt) {
```
BrunoBernardino, line 720: I need this param to not ask this question when setting a new date. The alternative would be to store any "input" value in a state, which seemed more complex and unnecessary.

packages/app/src/elements/item-view.ts
```diff
@@ -601,16 +712,28 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
...
+            }
+        }
+
         app.updateItem(this._item!, {
             name: this._nameInput.value,
             fields: [...this._fieldInputs].map((fieldEl: FieldElement) => fieldEl.field),
             tags: this._tagsInput.tags,
             auditResults: [],
             lastAudited: undefined,
+            expiresAt: this._item!.expiresAt,
```
BrunoBernardino, line 736: I hope this name is better.

packages/app/src/elements/item-view.ts
```diff
@@ -77,6 +77,9 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
     @state()
     private _fields: Field[] = [];
 
+    @state()
+    private _expiresAt?: Date;
+
```
MaKleSoft, line 82: This is in line with the rest of the editing logic. Basically we remember the updated property and don't apply it before the user hits save. This way the user can add/remove the expiration date in edit mode and hit cancel without the change actually being applied.

packages/app/src/elements/item-view.ts
```diff
@@ -234,24 +237,24 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
         }
     }
 
-    private _removeExpiration(skipSaving = false) {
+    private _removeExpiration(skipSaving = this._editing) {
```
MaKleSoft, line 240: We don't want to leave edit mode or save if they add/remove an expiration date while in edit mode.

packages/app/src/elements/renew-expiration-dialog.ts
```diff
@@ -41,23 +43,22 @@ export class RenewExpirationDialog extends Dialog<boolean, number> {
...
-            placeholder: $l("Enter New Expiration Date"),
-            confirmLabel: $l("Update"),
-            type: "date",
-            pattern: FIELD_DEFS[FieldType.Date].pattern.toString(),
-            message: await message(),
-        });
-
-        const days = Math.abs(differenceInDays(new Date(), new Date(newDate))) + 1;
+        this.open = false;
+        const newDate: string = await prompt(
```
MaKleSoft, line 47: You way works, too, but this is the intended way of using the prompt dialog.

BrunoBernardino:
@MaKleSoft I've simplified this further so we just store the number of days and calculate `updated + days` as you suggested. Check it out.

packages/app/src/lib/audit.ts
```diff
@@ -51,16 +51,6 @@ async function hasPasswordBeenCompromised(passwordHash: string) {
...
-function isItemExpiringOrExpired(expiryDate: Date) {
-    const thirtyDaysFromNow = add(new Date(), { days: 30 });
-
-    if (expiryDate <= thirtyDaysFromNow) {
-        return true;
-    }
-
-    return false;
-}
-
```
MaKleSoft: I decided that only items should be reported that are actually expired, the reason being that it's possible to add an expiration period of less than 30 days (may be rare, but we don't know how people are going to use it) and in that case items would always appear in the security report, making it kind of pointless. If people want items to appear in the security report sooner, they can simply enter a shorter expiration period.

packages/core/src/item.ts
```diff
@@ -323,14 +323,16 @@ export class VaultItem extends Serializable {
     @AsDate()
     lastAudited?: Date = undefined;
 
-    expiresEveryPeriodOfDays?: number = undefined;
+    /** Number of days after which the item expires */
+    expiresAfter?: number = undefined;
```
MaKleSoft, line 327: I know adding the unit to the property name can be considered good practice, but since we're not doing anywhere else (see https://github.com/padloc/padloc/blob/main/packages/core/src/app.ts#L49-L53 for example), I'd rather keep things consistent.

packages/app/src/elements/item-view.ts
```diff
@@ -632,49 +588,46 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
...
+                                                          10
+                                                      );
+                                                  }}
+                                                  .value=${this._expiresAfter?.toString() || ""}
+                                                  select-on-focus
+                                                  style="width: 4em"
+                                              >
+                                              </pl-input>
+                                              ${$l("days after being updated.")}
+                                          </div>
```
MaKleSoft, line 615: This will make translations a bit harder, but I think it's worth it for the cleaner look.

packages/app/src/elements/item-view.ts
```diff
@@ -234,64 +237,17 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
...
-            this._removeExpiration();
-        } else if (daysToAdd > 0) {
-            this._renewItemExpiration(daysToAdd);
+    private async _addExpiresAfter() {
+        if (this._editing) {
+            this._expiresAfter = 30;
+            await this.updateComplete;
+            this._expiresAfterInput.selectAll();
+        } else {
+            this.edit("addExpiration");
```
MaKleSoft, line 246: Given the new approach, I think this UX is a bit nicer and more consistent with the rest of the editing experience.

PR #569, opened 2022-08-26
Implement Bitwarden Import

### LINKED ISSUE #561: Dedicated Bitwarden Import

BrunoBernardino:
This should support all fields and data (except attachments, which aren't exported), for the JSON (unencrypted) format.

[You can get the sample Bitwarden provides](https://bitwarden.com/help/condition-bitwarden-import/#condition-a-json) to test (their totp example doesn't seem to match their expectation on their clients' source code because they still look for a `secret` search param).

Fixes #561

### DIFF
--- a/packages/app/src/lib/audit.ts
+++ b/packages/app/src/lib/audit.ts
@@ -251,7 +251,7 @@ export function noItemsTextForAudit(type: AuditType) {
         case AuditType.CompromisedPassword:
             return $l("You don't have any items with compromised passwords!");
         case AuditType.ExpiredItem:
-            return $l("You don't have any expiring or expired items!");
+            return $l("You don't have any expired items!");
         default:
             return $l("You don't have any insecure items!");
     }
@@ -303,7 +303,7 @@ export function descriptionForAudit(type: AuditType) {
             );
         case AuditType.ExpiredItem:
             return $l(
-                "Expiring or expired items are those that have been identified as being close to or past their set expiry date, which haven't been updated in a manually set number of days. These items should be rotated as soon as possible."
+                "Expired items are those that have been identified as being past their set expiry date, which haven't been updated in a given number of days. These items should be rotated as soon as possible."
             );
         default:
             "";
--- a/packages/app/src/lib/import.ts
+++ b/packages/app/src/lib/import.ts
@@ -8,9 +8,10 @@ import { translate as $l } from "@padloc/locale/src/translate";
 import { readFileAsText, readFileAsArrayBuffer } from "@padloc/core/src/attachment";
 
 import { OnePuxItem } from "./1pux-parser";
+import { BitwardenExport, BitwardenItem } from "./bitwarden-parser";
 
 export interface ImportFormat {
-    value: "csv" | "padlock-legacy" | "lastpass" | "padloc" | "1pux";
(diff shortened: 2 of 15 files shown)


packages/app/src/lib/audit.ts
```diff
@@ -251,7 +251,7 @@ export function noItemsTextForAudit(type: AuditType) {
         case AuditType.CompromisedPassword:
             return $l("You don't have any items with compromised passwords!");
         case AuditType.ExpiredItem:
-            return $l("You don't have any expiring or expired items!");
+            return $l("You don't have any expired items!");
```
BrunoBernardino, line 254: Just a few text fixes since you removed "expiring" from the list.

packages/app/src/lib/import.ts
```diff
@@ -425,11 +431,110 @@ export function is1Pux(file: File): boolean {
...
+    } catch (error) {
+        throw new Err(ErrorCode.INVALID_BITWARDEN, "Failed to parse Bitwarden .json file.");
+    }
+}
+
+export async function isBitwarden(file: File): Promise<boolean> {
+    try {
+        const content = await readFileAsText(file);
+        const data = JSON.parse(content);
+        return Object.prototype.hasOwnProperty.call(data, "items");
```
MaKleSoft, line 524: Just curious: Why not simply `data.hasOwnProperty("items")` or `Boolean(data.items)`?

MaKleSoft: Also, is there no better/more reliable name to identify a bitwarden file? It doesn't seem too unlikely that there are other export formats that might include an `items` property.

BrunoBernardino: You can read https://stackoverflow.com/questions/12017693/why-use-object-prototype-hasownproperty-callmyobj-prop-instead-of-myobj-hasow/ or https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwnProperty#using_hasownproperty_as_a_property_name for more thorough explanations, but basically, there's no guarantee `object.hasOwnProperty(property)` will work as expected (especially in client-side code), whereas `Object.prototype.hasOwnProperty.call(object, property)` will.

As for a more reliable name, there isn't, no, and if we run into problems, we can simply try to confirm there are some more required properties, like `items[x].type` and so on.

MaKleSoft: Alright! Thanks for the explantation :)

packages/app/src/lib/import.ts
```diff
@@ -425,11 +431,110 @@ export function is1Pux(file: File): boolean {
...
+            }
+        } else {
+            fields.push(
+                new Field({ name: extraField.name, value: extraField.value, type: extraField.type as FieldType })
+            );
+        }
+    }
+
+    return createVaultItem({ name: itemName, fields, tags });
+}
```
MaKleSoft, line 491: Why is this not in `lib/bitwarden-parser`?

MaKleSoft (approved):
Tested it and it works great! Had a couple of questions/comments but overall this looks good to merge!

packages/app/src/lib/import.ts
BrunoBernardino: It can be moved later, no real reason, it was just consistency with the 1pux importer, but that can be moved too.

PR #570, opened 2022-08-29
Admin Portal

MaKleSoft:
First pass at admin portal. This initial version will likely contain:

- [x] Display, search and delete Accounts
- [x] Display, search and delete Orgs
- [x] Display logs (including audit logs)


packages/core/src/server.ts
```diff
@@ -223,7 +237,7 @@ export class Controller extends API {
 
         ctx.provisioning = await this.provisioner.getProvisioning(auth, session);
 
-        await Promise.all([this.storage.save(session), this.storage.save(account), this.storage.save(auth)]);
+        await Promise.all([this.storage.save(session), this.storage.save(auth)]);
```
MaKleSoft, line 240: Saving the account object here wasn't actually necessary.

MaKleSoft:
Phew, this turned out to be a pretty big PR, because in order to enable some of the functionality required by the admin portal, I had to add quite a few capabilities to `core`:

- The `Storage` interface now also supports simple querying via the `StorageQuery` object.
- Add `count` method to `Storage` interface
- Added ability to log in as admin, which unlocks extended access to all endpoints
- Add list endpoints for accounts, orgs, and logs
- Other refactors that I won't list here

I startet building the audit logs on top of the existing logging mechanism but then decided to build dedicated mechanisms for both change logs and request logs (@BrunoBernardino you got that part right the first time). We'll probably deprecate the existing logging mechanism at some point or rename it to something else (like "tracking" or "analytics") since they don't really provide a lot of value on top of change and request logs. For now we'll leave it more or less as-is though (with a few optimisations).

@BrunoBernardino Since the big is quite big I don't expect you to look through all the code, but maybe pull it down and test it locally to see everything works as expected and maybe provide feedback about the functionality and UX. Note that the the Admin portal should be considered early alpha at this point and is obviously far from finished. But let's get this early version merged first and then gradually build on it.

BrunoBernardino:
I'll need some time to go through this, is it OK if I give you a review by EOW?

MaKleSoft:
> I'll need some time to go through this, is it OK if I give you a review by EOW?
Yep, np.

BrunoBernardino:
@MaKleSoft I've played around with this today for a while and it looked pretty good! Nice job! I only made a few small tweaks to the table displays, because I had a user with a really long name and email (from previous testing), and it looked terrible. Then I just added the `PL_SERVER_ADMINS` variable example to `example.env` and updated the branch. Should be ready to merge whenever you want!

Here's how it looks after my tweaks, with the automatic truncation via CSS:

Screenshot from 2022-09-29 18-27-59

BrunoBernardino (approved):

gautamsi:
Do you have plans to create a docker image for this? or it should be run separately?

BrunoBernardino:
> Do you have plans to create a docker image for this? or it should be run separately?

Sorry for the absurdly late reply here, it can run as part of the same docker image, just a separate service/package of this repo.

PR #572, opened 2022-08-30
Implement Item History

### LINKED ISSUE #434: Store item change history, allow "rewinding"

BrunoBernardino:
This is the initial version, still some issues creating history items and only displaying in console.

Also not storing vault changes yet, nor allowing rewind/delete, but will be a part of this PR.

Fixes #434


packages/app/src/elements/item-view.ts
```diff
@@ -654,6 +671,47 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
...
+                                <pl-icon icon="history" class="inline small light"></pl-icon>
+                                ${$l("History")}
+                            </h2>
+
+                            <pl-list class="border-top block" ?hidden=${!history.length}>
+                                ${history.map((historyEntry) => {
+                                    const historyEntryUpdater =
+                                        org && org.getMember({ accountId: historyEntry.updatedBy });
+                                    // TODO: Remove this
+                                    console.log(historyEntry);
```
BrunoBernardino, line 688: This showed me that on save, the `.history[]` entries have everything (including `.name` or `.tags` when I changed those), but right after sync (< 2 seconds, locally), those disappear. I still have `.fields`, for some reason. Ideas @MaKleSoft ?

packages/app/src/elements/item-view.ts
```diff
@@ -654,6 +671,47 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
...
+
+                                    return html`
+                                        <div
+                                            class="double-padded small hover click border-bottom horizontal center-aligning text-centering layout"
+                                            @click=${() => this._showHistoryEntry(historyEntry)}
+                                        >
+                                            ${$l("Changed")} ${historyEntry.name && " " + $l("name")}
```
BrunoBernardino, line 695: This looks a bit ugly so that we can keep this translatable.

packages/core/src/app.ts
```diff
@@ -1524,11 +1536,36 @@ export class App {
         }
     ) {
         const { vault } = this.getItem(item.id)!;
-        vault.items.update(new VaultItem({ ...item, ...upd, updatedBy: this.account!.id }));
+        const newItem = new VaultItem({ ...item, ...upd, updatedBy: this.account!.id });
+        this._updateHistory(item, newItem);
+        vault.items.update(newItem);
```
BrunoBernardino, line 1541: I assume/believe something here is removing the `.name` and `.tags` properties from `item.history[]` entries. I suppose it's serialization, but haven't found where @MaKleSoft .

packages/core/src/item.ts
```diff
@@ -278,6 +278,28 @@ export interface AuditResult {
...
+    constructor(vals: Partial<ItemHistory> = {}) {
+        super();
+        Object.assign(this, vals);
+    }
+
+    date: Date = new Date();
+
+    updatedBy: AccountID = "";
+
+    name?: string;
```
MaKleSoft, line 291: All class properties require an initialiser to be able to deserialised properly, even if they're optional. This is probably why these weren't stored. I.e. this needs to be:

```
name?: string = undefined;
```

packages/core/src/item.ts
```diff
@@ -278,6 +278,28 @@ export interface AuditResult {
...
+
+    vaultId?: string;
+
+    @AsSerializable(Field)
+    fields?: Field[];
```
MaKleSoft, line 296: FYI, properties with a serialisation decorator such as `@AsSerializable`, `@AsDate` are an exception and don't necessarily need to be initialized.

packages/core/src/item.ts
BrunoBernardino: Ah, TIL!

packages/core/src/app.ts
```diff
@@ -1524,11 +1536,36 @@ export class App {
...
+
+        if (JSON.stringify(oldItem.fields) !== JSON.stringify(newItem.fields)) {
+            newHistoryEntry.fields = oldItem.fields;
+        }
+
+        if (newHistoryEntry.name || newHistoryEntry.tags || newHistoryEntry.fields) {
+            // Prepend (first is the most recent change) and cap
+            newItem.history = [newHistoryEntry, ...newItem.history].slice(0, ITEM_HISTORY_ENTRIES_LIMIT);
+        }
+    }
```
MaKleSoft, line 1567: I think every history entry should contain every all of these properties, even if they haven't changed. This may take up a bit more space but makes restoring a state much simpler. In fact, I would argue that it's necessary. Consider the following scenario:

An item was changed > 10 times. Now the user wants to restore the last item the history. No fields had changed in that particular instance, but they did change later. Because the fields are not stored in the last history entry, end no older entries are available, the fields can not be restored to that point in time. So we either need a before and after for each of these properties or we'll simply store all properties in every history entry. I would argue that the latter option is much simpler and more practical, even if it potentially takes up a bit more space.

BrunoBernardino:
@MaKleSoft this now has the ability for you to view and restore history entries, for all relevant fields, with a nice UI/UX.

I started implementing the delete option, but it wasn't looking very clean (passing around a `historyEntryIndex` to `updateItem`), so I ditched it for now, as you can simply make small edits to "make it disappear".

Right now we're only missing the `vaultId` change (logging and restoring), but as I was thinking about implementing its restore, I'm also wondering if we need it, since restoring to a previous vault requires a bit more work (it might no longer exist, or you may no longer have access to the previous vault), and it seems different than the other changes we're logging and restoring to, here. What do you think?

packages/core/src/item.ts
```diff
@@ -278,6 +278,32 @@ export interface AuditResult {
...
+    constructor(vals: Partial<ItemHistory> = {}) {
+        super();
+        Object.assign(this, vals);
+    }
+
+    date: Date = new Date();
+
+    updatedBy: AccountID = "";
+
+    fieldsChanged: ItemHistoryFieldsChangedOption[] = [];
```
MaKleSoft, line 293: I don't think we need to store this here as we can easily do a diff later.

packages/core/src/item.ts
```diff
@@ -278,6 +278,32 @@ export interface AuditResult {
     fieldIndex: number;
 }
 
+export type ItemHistoryFieldsChangedOption = "name" | "vaultId" | "fields" | "tags";
+
+export class ItemHistory extends Serializable {
```
MaKleSoft, line 283: I think `ItemHistoryEntry` would be a slightly more appropriate name.

packages/app/src/elements/history-entry-dialog.ts
```diff
@@ -0,0 +1,121 @@
...
+                ${fieldsChanged.includes("fields") ? this._showChangedFieldsContent() : null}
+
+                <div class="horizontal evenly stretching spacing layout top-margined">
+                    <pl-button class="primary" @click=${() => this._restore()}>
+                        ${$l("Restore item to this version")}
+                    </pl-button>
+                    <pl-button @click=${this.dismiss}> ${$l("Cancel")} </pl-button>
+                </div>
+            </div>
+        `;
```
MaKleSoft, line 49: I would just display the entire item as it was at the time. In general in would view history entries as a snapshot of the entire item, rather than a change set. I think it's much easier for users to digest that way.

packages/app/src/elements/history-entry-dialog.ts
```diff
@@ -0,0 +1,121 @@
...
+                    ${$l("Name")}
+                </h2>
+                <div class="bottom-margined">${this._historyEntry.name || $l("<None>")}</div>
+            </div>
+        `;
+    }
+
+    private _showChangedTagsContent() {
+        return html`
+            <div class="tags border-bottom">
```
MaKleSoft, line 72: `tags` seems to be the wrong class for this. Better use `horizontal layout`.

packages/app/src/elements/history-entry-dialog.ts
```diff
@@ -0,0 +1,121 @@
...
+
+    async show(historyEntry: ItemHistory) {
+        await this.updateComplete;
+        this._historyEntry = historyEntry;
+        return super.show();
+    }
+
+    private _showChangedNameContent() {
+        return html`
+            <div class="tags border-bottom">
```
MaKleSoft, line 60: Again, please use `horizontal layout` instead of `tags`.

packages/app/src/elements/history-entry-dialog.ts
MaKleSoft: Actually, better to display tags below the header, same as in the item view.

packages/app/src/elements/history-entry-dialog.ts
```diff
@@ -0,0 +1,121 @@
...
+                <h2 class="subtle horizontally-double-margined bottom-margined animated section-header">
+                    <pl-icon icon="field" class="inline small light"></pl-icon>
+                    ${$l("Fields")}
+                </h2>
+                <pl-list class="border-top block">
+                    ${repeat(
+                        this._historyEntry.fields,
+                        (field) => `${field.name}_${field.type}`,
+                        (field: Field) => html`
+                            <pl-field class="padded list-item" .field=${field} .editing=${false} .auditResults=${[]}>
```
MaKleSoft, line 101: I would set `pointer-events: none` on this to disable field actions and dragging.

packages/core/src/item.ts
```diff
@@ -278,6 +278,32 @@ export interface AuditResult {
...
+    constructor(vals: Partial<ItemHistory> = {}) {
+        super();
+        Object.assign(this, vals);
+    }
+
+    date: Date = new Date();
```
MaKleSoft, line 289: This should be called `time`, since it contains the full date and time.

packages/locale/res/translations/_template.json
```diff
@@ -451,6 +451,46 @@
     "This group does not have access to any vaults yet.",
```
MaKleSoft, line 451: Can we agree to wait with updating the translations until right before merging? Makes the PR a little less noisy.

packages/core/src/item.ts
MaKleSoft: Actually, let's call it `updated`, which should reflect the `updated` property of the item before the change.

MaKleSoft: Aaactually, let's keep this property and rename it to `created`, as well as an additional `updated` property as described above.

packages/app/src/elements/history-entry-dialog.ts
```diff
@@ -0,0 +1,121 @@
...
+
+                <div class="stretch"></div>
+
+                ${fieldsChanged.includes("name") ? this._showChangedNameContent() : null}
+                ${fieldsChanged.includes("tags") ? this._showChangedTagsContent() : null}
+                ${fieldsChanged.includes("fields") ? this._showChangedFieldsContent() : null}
+
+                <div class="horizontal evenly stretching spacing layout top-margined">
+                    <pl-button class="primary" @click=${() => this._restore()}>
+                        ${$l("Restore item to this version")}
```
MaKleSoft, line 44: Let's keep this button label a bit shorter. Currently this looks a it weird on a narrow screen.

packages/app/src/elements/item-view.ts
```diff
@@ -654,6 +687,41 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
...
+                                            ${$l("Changed")} ${historyEntry.fieldsChanged.join(", ")} ${" "}
+                                            ${until(formatDateFromNow(historyEntry.date))}
+                                            ${historyEntryUpdater && " " + $l("by {0}", historyEntryUpdater.email)}
+                                        </div>
+                                    `;
+                                })}
+                            </pl-list>
+
+                            <div
+                                class="border-top border-bottom double-padded small hover center-aligning text-centering"
```
MaKleSoft, line 718:
```suggestion
                                class="border-top border-bottom double-padded small center-aligning text-centering"
```

packages/core/src/item.ts
BrunoBernardino: We wouldn't be able to tell what changed for the "oldest" change, though, because there's no older reference.

packages/app/src/elements/history-entry-dialog.ts
BrunoBernardino: I think that's harder to grasp at what changed, though. If you have an item with 10 fields, and only the tags or name changed, it'll be harder to understand that if you're seeing everything.

I wish we could just display the fields that changed, but that's harder to do.

packages/app/src/elements/history-entry-dialog.ts
BrunoBernardino: I think this way makes it more obvious that's what changed (related to the above).

packages/core/src/item.ts
BrunoBernardino: I'd rather keep it as `date` because it is a `Date` object. I'm happy to change it to `time` or `created` but not really `updated` (nor 2 date properties, because when would they be different? You're not going to change history entries)

packages/locale/res/translations/_template.json
BrunoBernardino: Ok, I just like to see the tests passing, but can live with that.

packages/app/src/elements/item-view.ts
```diff
@@ -654,6 +687,41 @@ export class ItemView extends Routing(StateMixin(LitElement)) {
...
+                                        org && org.getMember({ accountId: historyEntry.updatedBy });
+
+                                    return html`
+                                        <div
+                                            class="double-padded small hover click border-bottom horizontal center-aligning text-centering"
+                                            @click=${() => this._showHistoryEntry(historyEntry)}
+                                        >
+                                            ${$l("Changed")} ${historyEntry.fieldsChanged.join(", ")} ${" "}
+                                            ${until(formatDateFromNow(historyEntry.date))}
+                                            ${historyEntryUpdater && " " + $l("by {0}", historyEntryUpdater.email)}
```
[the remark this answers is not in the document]
BrunoBernardino: Could you elaborate? Show the actual value changes here? I was trying to prevent that because of potentially showing passwords.

packages/core/src/app.ts
```diff
@@ -1524,11 +1536,39 @@ export class App {
...
+        this._updateHistory(item, newItem);
+        vault.items.update(newItem);
         await this.saveVault(vault);
         await this.syncVault(vault);
     }
 
+    protected _updateHistory(oldItem: VaultItem, newItem: VaultItem) {
+        const newHistoryEntry = new ItemHistory();
+
+        newHistoryEntry.updatedBy = newItem.updatedBy;
```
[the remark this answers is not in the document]
BrunoBernardino: And here is revealed the problem of a field that has the same name on both classes. This `updatedBy` is for the history entry, not the item itself. This isn't a value that's restored, so we want to know who changed the history (who created this entry), not who had made it that way before.

That being said, we could still infer one from the current `updatedBy`, but we would lose that reference for the oldest entry.

BrunoBernardino:
@MaKleSoft you didn't answer on the vault change so I'm assuming it can be deferred and will prepare this PR without that. If we want to store/save that, we can add that in later.

packages/app/src/elements/history-entry-dialog.ts
BrunoBernardino: This makes it impossible to reveal/see password fields, so is there a better/different way to remove dragging that doesn't disable that important functionality?

packages/core/src/item.ts
MaKleSoft: Good point! But honestly, only knowing that the fields changed an not which fields and what from isn't really that helpful either. Also, we can work around that problem easily by simply storing 11 history items but only displaying 10.

packages/app/src/elements/history-entry-dialog.ts
MaKleSoft: I think you are focusing too much on displaying changes. Trust me, people care much less about what has changed and much more about simply restoring an item to a certain point in time. In fact, I would probably not display changes at all but if you think it would make for a better UX we can highlight the changed properties somehow.

packages/app/src/elements/history-entry-dialog.ts
MaKleSoft: Not sure what you mean? The `tags` class is meant to be used explicitly as a container for tags (`<div class="tag"></div>` ) and is simply out of place here. Or are you talking about horizontal vs vertical layout?

packages/app/src/elements/history-entry-dialog.ts
MaKleSoft: Good point! In that case I guess you'll either need to add something like a `readonly` property or (probably easier) simply recreate the display portion of the UI. I'd actually prefer the latter approach I think since I'd like to avoid adding more complexity to `<pl-field>`.

packages/app/src/elements/history-entry-dialog.ts
BrunoBernardino: I changed the class. Was referring mostly to the 

> Actually, better to display tags below the header, same as in the item view.

comment.

packages/app/src/elements/history-entry-dialog.ts
BrunoBernardino: I added a simple `hideEdit` to hide that button, but keep the rest as it'll be familiar UX.

packages/core/src/item.ts
MaKleSoft: I think property names should be based on the information the contain, rather than the name of their type. My point is that `created` is more descriptive because it contains the information of when the history item was created. `updated` on the other hand is *supposed* to mirror the same property from the original item (although I'm open to a different name). We don't want to restore that value but I'd like to potentially use it for displaying the time line. I.e. show the time that version of the item was created rather than the time it was changed.

Again, I'd like the whole thing to be a list of snapshots, rather than a list of changes. I.e.: "What did the item look like at a certain point in time" vs "What has changed and when". Remember the main point of this feature is **not** to serve as an audit log but simply to allow people to restore a previous version of an item.

packages/app/src/elements/item-view.ts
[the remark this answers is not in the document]
MaKleSoft: As explained in another comment:

> Again, I'd like the whole thing to be a list of snapshots, rather than a list of changes. I.e.: "What did the item look like at a certain point in time" vs "What has changed and when". Remember the main point of this feature is not to serve as an audit log but simply to allow people to restore a previous version of an item.

I **don't** want to display a list of changes (or at least not focus on them) but instead simply show that a different version of this item existed at a certain point in time. They can then "travel back" to that time and see how the item looked back then and, if desired, restore that point in time. Very similar to how time machine works on MacOS.

packages/core/src/app.ts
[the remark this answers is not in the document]
MaKleSoft: It's just a matter of perspective. **You** want to display who has made the change (and caused the history item to be created). **I** want to display who had created that previous version. We already know who has triggered the change by looking at the `updatedBy` property of the subsequent history entry (or the current version).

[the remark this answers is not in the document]
MaKleSoft: This is directly connected to my proposed `updated` property on the history item btw. Both should reflect who and when had created that **version** of the item **not** who and when the change was made. Does that make sense?

MaKleSoft:
> @MaKleSoft you didn't answer on the vault change so I'm assuming it can be deferred and will prepare this PR without that. If we want to store/save that, we can add that in later.

Sorry, forgot to address that. Yeah, I agree that tracking and restoring the history beyond moving the item to a different vault is way more complicated and probably not worth it. In fact, what actually happens when an item is removed is that it is "deleted" from the old vault and added to the new vault. This is also how it should be show in the history I think. (Once we have soft deletes we can still keep the history of the old item around for a while).

MaKleSoft:
@BrunoBernardino Sorry for being so fastidious, but I really want to make sure we get the UX right, and we really only have one shot at this, since changing it later will be very hard without breaking backward compatibility.

I'd like to reiterate my main gripe with the current implementation: We're basically talking about two subtly different things here. I'm talking about an item **history**, you're talking about a **change log**. The former is simply a list of **snapshots** of previous version of the item (how did the item look at a certain point in time), while the latter documents and highlights **changes** (what was changed and when). Both have their own advantages and justifications, but unfortunately we have to decide one one, and I'm strongly in favor of the "history" or "snapshots" approach.

packages/app/src/elements/history-entry-dialog.ts
MaKleSoft: Is it still draggable though?

BrunoBernardino:
> I'd like to reiterate my main gripe with the current implementation: We're basically talking about two subtly different things here. I'm talking about an item history, you're talking about a change log. The former is simply a list of snapshots of previous version of the item (how did the item look at a certain point in time), while the latter documents and highlights changes (what was changed and when). Both have their own advantages and justifications, but unfortunately we have to decide one one, and I'm strongly in favor of the "history" or "snapshots" approach.

Apart from some minor things (`date` vs `created`, adding `updated`, and changing the `updatedBy` value), I don't see how the implementation here is different. I agree the displaying is, though, and I can accommodate that, no problem.

Am I missing something, or do you agree with that?

MaKleSoft:
> @MaKleSoft you didn't answer on the vault change so I'm assuming it can be deferred and will prepare this PR without that. If we want to store/save that, we can add that in later.

Yep, agreed :) Btw, if what I'm saying is not making sense to you or you don't quite understand where I'm going with this, I'm happy to take over for the rest of the UI work (I'll probably make some final adjustments, anyway). Sorry, but when it comes to the UI/UX I have a very specific vision and It's hard for me to let that stuff go 😅

BrunoBernardino:
Well, if you want to, you can take over right now, then, and I'll start working on some other unrelated stuff.

MaKleSoft:
Yeah that'd probably be easier / faster than trying to convey to you what I have in mind exactly 😅. Just to be clear - I really like what you've done so far! It's just a little different from what I am envisioning.

MaKleSoft:
Again ready for review. My changes:

- History is now displayed as list of versions instead of changes, with the current version displayed at the top of the list (not clickable, replaces the note on the bottom).
- History entry dialog now always display the item name, tags and fields with diff displayed for each (compared to current version).
- Made viewing / restoring from item history a premium feature in SaaS version


Screenshot 2022-09-03 at 18 01 19
Screenshot 2022-09-03 at 18 01 42

MaKleSoft (commented):
@BrunoBernardino Back to you ;)

MaKleSoft (approved):

MaKleSoft:
@BrunoBernardino Can't request your review for some reason. Let me know if you have any comments!

BrunoBernardino:
I've reviewed it, pulled it down, played around with it a bit, and all looks good! Not my personal preference, but I think it works well!

PR #587, opened 2022-09-07
Allow org premium features in account

### LINKED ISSUE #579: Premium Features should be available for all members of an organisation with a paid plan
Currently premium features like TOTP, item history etc. are not available inside the private vault of organisation members. Let's change this and make sure that all members of an organisation with a paid plan get to enjoy all premium features in their private vault as well 😊

BrunoBernardino:
Fixes #579

### DIFF
--- a/packages/core/src/app.ts
+++ b/packages/core/src/app.ts
@@ -2073,7 +2073,27 @@ export class App {
     }
 
     getAccountFeatures() {
-        return this.getAccountProvisioning()?.features || new AccountFeatures();
+        const accountFeatures = this.getAccountProvisioning()?.features || new AccountFeatures();
+
+        // If there's a "premium feature" disabled, try to get an org's features instead
+        if (this._hasPremiumFeaturesDisabled(accountFeatures)) {
+            for (const orgId of this.getAccountProvisioning()?.orgs) {
+                const org = this.getOrg(orgId);
+                if (org) {
+                    const orgFeatures = this.getOrgFeatures(org.info);
+                    if (!this._hasPremiumFeaturesDisabled(orgFeatures)) {
+                        accountFeatures.attachments = orgFeatures.attachments;
+                        accountFeatures.securityReport = orgFeatures.securityReport;
+                        accountFeatures.totpField = orgFeatures.totpField;
+                        accountFeatures.notesField = orgFeatures.notesField;
+                        accountFeatures.itemHistory = orgFeatures.itemHistory;
+                        return accountFeatures;
+                    }
+                }
+            }
+        }
+
+        return accountFeatures;
     }
 
     getOrgFeatures(org: OrgInfo) {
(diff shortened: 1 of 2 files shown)


packages/core/src/app.ts
```diff
@@ -2086,6 +2106,10 @@ export class App {
      * ================
      */
 
+    private _hasPremiumFeaturesDisabled(features: AccountFeatures | OrgFeatures) {
```
BrunoBernardino, line 2109: I wasn't sure if you wanted to do something different here (or check for more features like `attachments`).

packages/core/src/app.ts
```diff
@@ -2073,7 +2073,27 @@ export class App {
...
+                const org = this.getOrg(orgId);
+                if (org) {
+                    const orgFeatures = this.getOrgFeatures(org.info);
+                    if (!this._hasPremiumFeaturesDisabled(orgFeatures)) {
+                        accountFeatures.attachments = orgFeatures.attachments;
+                        accountFeatures.securityReport = orgFeatures.securityReport;
+                        accountFeatures.totpField = orgFeatures.totpField;
+                        accountFeatures.notesField = orgFeatures.notesField;
+                        accountFeatures.itemHistory = orgFeatures.itemHistory;
+                        return accountFeatures;
```
BrunoBernardino, line 2090: I intentionally return without saving here so that if the user leaves the org, their account won't "magically" keep the features enabled.

packages/server/src/provisioning/stripe.ts
```diff
@@ -493,7 +493,7 @@ export class StripeProvisioner extends BasicProvisioner {
                 undefined,
                 undefined,
                 true,
-                "history"
+                "History"
```
BrunoBernardino, line 496: This just seemed like a typo.

MaKleSoft: That argument is case-insensitive. So either way is fine :)

MaKleSoft (changes requested):
I don't think we should be doing this on the client side. In fact, this should be a concern of the Stripe provisioner only. Specifically, [this](https://github.com/padloc/padloc/blob/main/packages/server/src/provisioning/stripe.ts#L237) would be a good place to put that logic. Basically, enable a feature if at least one org the user is a member of has it enabled. I.e.:

```ts
for (const orgProvisioning of provisioning.orgs) {
    provisioning.account.features.attachments.disabled = 
        provisioning.account.features.attachments.disabled && orgProvisioning.features.attachments.disabled;
    ...
}
```

We'll also have to make sure that the storage quota is > 0 or they won't be able to upload anything. Let's set it to 1GB:

```ts
if (!provisioning.account.features.attachments.disabled && !provisioning.account.quota.storage) {
    provisioning.account.quota.storage = 1000;
}
```

BrunoBernardino:
Good points. I chose to do it on the client side since that's where we were checking anyway, and it seemed "too much" to move it to the abstract provisioning layer, and seemed too "our-use-case-focused" to put it in Stripe, but I'll add it to Stripe, since anyone who self-hosts/customizes Padloc (that don't use the Stripe provisioner) will likely enable everything for everyone anyway.

Issue #347, opened 2021-12-23
Support for more import formats


MaKleSoft:
- [x] Apple Keychain
- [x] Chrome
- [x] Firefox
- [x] Edge
- [x] Brave
- [x] Lastpass (check current implementation, support newer formats)
- [x] Dashlane
- [x] KeePass
- [x] Bitwarden
- [x] NordPass

Any requests?

PR #590, opened 2022-09-12
Dashlane, KeePass, and NordPass import support

BrunoBernardino:
This adds support to more easily import Dashlane, KeePass, and NordPass CSV export files.

Related to #347

Sample files:
- [dashlane_credentials_sample.csv](https://github.com/padloc/padloc/files/9547443/dashlane_credentials_sample.csv)
- [keepass_sample.csv](https://github.com/padloc/padloc/files/9547444/keepass_sample.csv)
- [nordpass_sample.csv](https://github.com/padloc/padloc/files/9547446/nordpass_sample.csv)

@MaKleSoft not sure if you want to merge this to `main` for a `v4.1.1` with these "soon" or `v4.2.0`.


packages/app/src/lib/import.ts
```diff
@@ -536,6 +570,229 @@ export async function isBitwarden(file: File): Promise<boolean> {
...
+    const zipCodeIndex = 9;
+    const folderIndex = 10;
+    const fullNameIndex = 11;
+    const phoneIndex = 12;
+    const emailIndex = 13;
+    const address1Index = 14;
+    const address2Index = 15;
+    const cityIndex = 16;
+    const countryIndex = 17;
+    const stateIndex = 18;
```
BrunoBernardino, line 697: This could've been a map with the index, field name, and field type, but it was only saving 10-20 lines and this is very readable, just long.

MaKleSoft (approved):
Looks good! Let's just merge this into main.

I'm thinking that with all these different import options, we should probably change the UX a bit. Like I think it would be nicer if people could select the desired import format first, and then select the file. This would have the advantage that it's more obvious what kind of import files are supported and we wouldn't have to try every option to figure out what kind of file it is. This would be part of a different PR of course. What do you think?

BrunoBernardino:
> I'm thinking that with all these different import options, we should probably change the UX a bit. Like I think it would be nicer if people could select the desired import format first, and then select the file. This would have the advantage that it's more obvious what kind of import files are supported and we wouldn't have to try every option to figure out what kind of file it is. This would be part of a different PR of course. What do you think?

I'm not sure. I like that we can detect and show the proper choice based on the file (because otherwise the import might fail in spectacularly vague ways), _but_ I'd argue we should let customers choose CSV (if it is CSV file) for the advanced mapping, for example, as they might want to name fields differently, or add more tags, etc.

PR #601, opened 2022-09-24
Add configurable healthcheck endpoint

### LINKED ISSUE #279: [Feature Request] Endpoint for Health Check using GET Method by LoadBalancer
I was tried to build padloc on Google Kubernetes Engine using Google Cloud Load Balancer(GCLB)
But, GKE Ingress has require passing health check using GET method.

padloc is returning 405(method not allowed) error except for CONNECT/POST method.
So, It will be didn't work well.

I want you add health check endpoint using GET method.

BrunoBernardino:
Defaults to `/healthcheck` and can be configured via `PL_SERVER_HEALTHCHECK_PATH`, which will return a 200 status code response for a `GET` request.

Fixes #279

### DIFF
--- a/packages/core/src/server.ts
+++ b/packages/core/src/server.ts
@@ -88,6 +88,10 @@ export class ServerConfig extends Config {
     @ConfigParam()
     scimServerUrl = "http://localhost:5000";
 
+    /** Path on the server for responding with 200, to be used in health checks (e.g. load balancers) */
+    @ConfigParam()
+    healthcheckPath = "/healthcheck";
+
     constructor(init: Partial<ServerConfig> = {}) {
         super();
         Object.assign(this, init);
@@ -2003,6 +2007,17 @@ export class Server {
         const context: Context = {};
 
         const start = Date.now();
+
+        if (req.method.startsWith("GET:")) {
+            if (req.method === `GET:${this.config.healthcheckPath}`) {
+                res.result = 200;
+                return res;
+            }
+
+            // Reject non-healthcheck GET requests with legacy response
+            res.result = 405;
+            return res;
+        }
 
         try {
             context.device = req.device;
--- a/packages/core/src/server.ts
+++ b/packages/core/src/server.ts
@@ -88,10 +88,6 @@ export class ServerConfig extends Config {
     @ConfigParam()
     scimServerUrl = "http://localhost:5000";
 
-    /** Path on the server for responding with 200, to be used in health checks (e.g. load balancers) */
-    @ConfigParam()
-    healthcheckPath = "/healthcheck";
-
     constructor(init: Partial<ServerConfig> = {}) {
         super();
         Object.assign(this, init);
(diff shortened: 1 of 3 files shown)


packages/core/src/server.ts
```diff
@@ -2004,6 +2008,17 @@ export class Server {
 
         const start = Date.now();
 
+        if (req.method.startsWith("GET:")) {
+            if (req.method === `GET:${this.config.healthcheckPath}`) {
```
BrunoBernardino, line 2012: The alternative to this weird-looking thing would be to make it an `HTTPTransport` config setting instead, but that also felt wrong, as this is really a "server" thing, not a "transport" one.

packages/server/src/transport/http.ts
```diff
@@ -59,6 +59,20 @@ export class HTTPReceiver implements Receiver {
                 case "OPTIONS":
                     httpRes.end();
                     break;
+                case "GET":
+                    try {
+                        const req = new Request();
+                        // httpReq.url will include searchParams, so we parse the pathname properly
+                        const url = new URL(`http://localhost${httpReq.url || ""}`);
```
BrunoBernardino, line 66: The host won't be used, so I didn't want to inject/pass anything around.

packages/core/src/server.ts
MaKleSoft: Actually, now that I think about it I think it **should** be a concern of the transport layer. In fact, this whole code block seems kind of pointless, since `HTTPReceiver` is already setting the `200` status code and the response body is irrelevant in this case, anyway. So yeah, let's make it a setting of `HTTPTransport`.

MaKleSoft (changes requested):
As mentioned in my comment, let's make this a setting of `HTTPReceiver` and leave the `Server` component out of it for now. In the future, if we decide to implement a more elaborate check (like also verifying database connectivity etc), we should add a second handler to `Receiver::listen` instead which can then be bound to a `Server::healthCheck` method. I.e.:

```ts
// packages/core/src/transport.ts
export interface Receiver {
    listen(handler: (req: Request) => Promise<Response>, healthCheck: () => Promise<boolean>): void;
}

// packages/server/src/init.ts
new HTTPReceiver(config.transport.http)
    .listen((req) => server.handle(req), () => server.healthCheck());
```

packages/server/src/transport/http.ts
```diff
@@ -39,6 +39,10 @@ export class HTTPReceiverConfig extends Config {
 
     @ConfigParam()
     allowOrigin: string = "*";
+
+    /** Path on the HTTP server for responding with 200, to be used in health checks (e.g. load balancers) */
+    @ConfigParam()
+    healthcheckPath = "/healthcheck";
```
MaKleSoft, line 45: Minor nitpick: I think I'd prefer `healthCheckPath`

MaKleSoft (changes requested):
Nice and simple - I like it. I think `healthcheckPath` should be `healthCheckPath` but other than that it looks good!

MaKleSoft (approved):

PR #606, opened 2022-10-01
Allow overriding CSV import format

### LINKED ISSUE #591: Allow overriding chosen import file type, when it's a CSV
When importing passwords we try to determine if a CSV is of a specific app (LastPass, Dashlane, KeePass, NordPass, iCloud, Chrome, Firefox), and while that should result in a great experience, we should allow customers to choose a CSV for our own advanced CSV import, so they can make better mappings of the fields themselves.

BrunoBernardino:
Fixes #591

### DIFF
--- a/packages/app/src/elements/import-dialog.ts
+++ b/packages/app/src/elements/import-dialog.ts
@@ -22,6 +22,8 @@ fieldTypeOptions.push({
     value: "skip",
 });
 
+const csvSupportedFormatValues = imp.csvSupportedFormats.map((importFormat) => importFormat.value);
+
 @customElement("pl-import-dialog")
 export class ImportDialog extends Dialog<File, void> {
     @state()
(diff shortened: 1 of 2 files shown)


packages/app/src/elements/import-dialog.ts
```diff
@@ -65,16 +67,24 @@ export class ImportDialog extends Dialog<File, void> {
 
     renderContent() {
         const csvHasColumnsOnFirstRow = this._csvHasColumnsOnFirstRowButton?.active;
+        const hasChosenCsvSupportedFormat = Boolean(
+            this._formatSelect?.value && csvSupportedFormatValues.includes(this._formatSelect?.value)
+        );
+        const formatOptions = imp.supportedFormats.map((importFormat) => ({
+            ...importFormat,
+            disabled: hasChosenCsvSupportedFormat && !csvSupportedFormatValues.includes(importFormat.value),
```
BrunoBernardino, line 75: I ran into a bunch of problems with changing the options array (removing non-CSV options when choosing a CSV-supported format), until I found this gem was already supported in our `pl-select`. Yay!

MaKleSoft (approved):

PR #609, opened 2022-10-05
Fix update deployment GitHub Action

BrunoBernardino:
Apparently _some_ variables require the syntax in the `if:` bit, though it's not exactly clear which.


BrunoBernardino:
What do you know, it's properly documented: https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-using-secrets

MaKleSoft (approved):

PR #626, opened 2022-11-13
Caddy Docker Example

BrunoBernardino:
This adds a working example for using Padloc with Caddy, which automatically handles SSL/HTTPS.

### DIFF
--- a/docs/examples/hosting/docker/caddy/Caddyfile
+++ b/docs/examples/hosting/docker/caddy/Caddyfile
@@ -1,0 +1,28 @@
+localhost {
+    encode gzip zstd
+
+    handle_path /server* {
+        reverse_proxy * server:3000 {
+            header_up Host {host}
+            header_up X-Real-IP {remote}
+            # header_up X-Forwarded-For {remote}
+            header_up X-Forwarded-Host {upstream_hostport}
+        }
+    }
+
+    handle {
+        root * /pwa
+    
+        file_server
+
+        try_files {path} /index.html
+
+        header {
+            X-Frame-Options DENY
+        }
+
+        request_body {
+            max_size 10MB
+        }
+    }
+}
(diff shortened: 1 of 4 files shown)


docs/examples/hosting/docker/caddy/Caddyfile
```diff
@@ -0,0 +1,28 @@
+localhost {
+    encode gzip zstd
+
+    handle_path /server* {
```
BrunoBernardino, line 4: `handle_path` instead of `handle` so the `/server` path isn't sent with the request to the proxy.

docs/examples/hosting/docker/caddy/Caddyfile
```diff
@@ -0,0 +1,28 @@
...
+        reverse_proxy * server:3000 {
+            header_up Host {host}
+            header_up X-Real-IP {remote}
+            # header_up X-Forwarded-For {remote}
+            header_up X-Forwarded-Host {upstream_hostport}
+        }
+    }
+
+    handle {
```
BrunoBernardino, line 13: A separate `handle` block is required in order to have the SPA-routing not interfere with the proxy routing.

docs/examples/hosting/docker/caddy/Caddyfile
```diff
@@ -0,0 +1,28 @@
...
+        root * /pwa
+    
+        file_server
```
BrunoBernardino, line 16: This will serve all files inside the root above `/pwa`

docs/examples/hosting/docker/caddy/Caddyfile
```diff
@@ -0,0 +1,28 @@
+localhost {
+    encode gzip zstd
+
+    handle_path /server* {
+        reverse_proxy * server:3000 {
+            header_up Host {host}
+            header_up X-Real-IP {remote}
+            # header_up X-Forwarded-For {remote}
```
BrunoBernardino, line 8: This header is commented because it generates a warning/notice, since it's sent by default.

MaKleSoft (commented):
Nice, looks super straightforward! Just one question. Once ready to deploy in production, how does one switch from a self-signed certificate to something like letsencrypt?

MaKleSoft (approved):

PR #628, opened 2022-11-20
Add oauth authenticator and provisioner

MaKleSoft:
This adds support for using oauth as a default mechanism in place of email. Currently only works with PWA and electron app.

### DIFF
--- a/packages/server/src/provisioning/oauth.ts
+++ b/packages/server/src/provisioning/oauth.ts
@@ -1,0 +1,145 @@
+import {
+    AccountProvisioning,
+    BasicProvisioner,
+    BasicProvisionerConfig,
+    Provisioning,
+    ProvisioningStatus,
+} from "@padloc/core/src/provisioning";
+import { getIdFromEmail } from "@padloc/core/src/util";
+import { Storage } from "@padloc/core/src/storage";
+import { ConfigParam } from "@padloc/core/src/config";
+import { request } from "../transport/http";
+import { AccountID } from "@padloc/core/src/account";
+import { Auth } from "@padloc/core/src/auth";
+import { OauthConfig, OauthUserInfo } from "../auth/oauth";
+
+export class OauthProvisionerConfig extends BasicProvisionerConfig {
+    @ConfigParam("number")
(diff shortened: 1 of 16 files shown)


packages/server/src/provisioning/oauth.ts
```diff
@@ -0,0 +1,145 @@
...
+    async accountDeleted(_params: { email: string; accountId?: string }): Promise<void> {}
+
+    private async _sync({ account }: Provisioning) {
+        if (!this.oauthConfig) {
+            throw "No oauth configuration found!";
+        }
+
+        if (!account.metaData?.oauth?.tokens) {
+            account.status = ProvisioningStatus.Unprovisioned;
+            account.statusLabel = "Access Denied";
```
BrunoBernardino, line 75: Is there a reason for the strings in this file to not have been translated?

packages/server/src/provisioning/oauth.ts
```diff
@@ -0,0 +1,145 @@
...
+            account.actionUrl = undefined;
+            await this.storage.save(account);
+            return;
+        }
+    }
+
+    protected _getAccountProvisioningFromUserInfo(userInfo: OauthUserInfo): Partial<AccountProvisioning> {
+        return {
+            status: ProvisioningStatus.Active,
+            statusLabel: "Aktive",
```
BrunoBernardino, line 134: Ist dies ein Fehler?

BrunoBernardino (approved):
I didn't pull this down, but reviewed the code; it looks good!

I only have a couple of questions/comments which aren't blocking.

packages/server/src/provisioning/oauth.ts
MaKleSoft: Mostly because I don't expect this class to actually be used in production. It's more of a base class for more advanced  oauth-based provisioning.

packages/server/src/provisioning/oauth.ts
MaKleSoft: Haha, yeah that's a typo. Not a big deal though for the reason mentioned above.

PR #630, opened 2022-11-20
Fix admin path deployment

BrunoBernardino:
This is a partial fix to deploying admin in a `/path`. Routing, however, doesn't work (redirects to `/` automatically on login), and probably needs some fundamental change.


BrunoBernardino:
FYI @MaKleSoft

BrunoBernardino:
@MaKleSoft even if this doesn't "fix everything" I'd still like to merge it as it makes it easier to distinguish when one thing was loaded vs another. Is that ok?

PR #635, opened 2022-11-25
Fix long tag display

### LINKED ISSUE #616: Display issues when using very long tag names
As seen here: https://github.com/padloc/padloc/discussions/615

BrunoBernardino:
Fixes #616

I didn't implement the limit on input because the only time people complained about long names was with automatic imports, which wouldn't be fixed by that new limitation.

I also tried adding a tooltip, but failed, and in the end, clicking on the item allows viewing the tag without it being truncated, so it should be fine.

Screenshot from 2022-11-25 15-07-11

### DIFF
--- a/packages/app/src/styles/misc.ts
+++ b/packages/app/src/styles/misc.ts
@@ -64,6 +64,10 @@ export const misc = css`
         color: var(--color-negative);
     }
 
+    .tag.ellipsis {
+        max-width: 100px;
+    }
+
     .input-wrapper {
         display: flex;
         align-items: center;
--- a/packages/app/src/styles/misc.ts
+++ b/packages/app/src/styles/misc.ts
@@ -64,10 +64,6 @@ export const misc = css`
         color: var(--color-negative);
     }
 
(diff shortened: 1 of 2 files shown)


packages/app/src/styles/misc.ts
```diff
@@ -64,6 +64,10 @@ export const misc = css`
         color: var(--color-negative);
     }
 
+    .tag.ellipsis {
+        max-width: 100px;
+    }
+
```
MaKleSoft, line 70: I think this rule is much too broad. Can we add this [here](https://github.com/padloc/padloc/blob/main/packages/app/src/elements/items-list.ts#L163) instead? Also, please try to use `em` as sizing units wherever possible.

MaKleSoft (changes requested):

packages/app/src/styles/misc.ts
BrunoBernardino: Done.

MaKleSoft (approved):

PR #636, opened 2022-11-25
Add DigitalOcean "one-click-install" button

BrunoBernardino:
This adds the button to the README and fixes the deploy template to actually work (still to be tested).

URL to test: https://cloud.digitalocean.com/apps/new?repo=https://github.com/padloc/padloc/tree/feature/digitalocean-one-click-install

PR #637, opened 2022-11-25
Add DigitalOcean "one-click-install" button

BrunoBernardino:
This adds the button to the README and fixes the deploy template to actually work (still to be tested).

URL to test: https://cloud.digitalocean.com/apps/new?repo=https://github.com/padloc/padloc/tree/feature-digitalocean-one-click-install (I had to do this PR instead of https://github.com/padloc/padloc/pull/636 because the `/` in the branch name broke DO...)

### DIFF
--- a/.do/deploy.template.yaml
+++ b/.do/deploy.template.yaml
@@ -66,7 +66,7 @@ spec:
     databases:
         - name: db
           engine: PG
-          version: "v14"
+          version: "14.x"
           production: false
           db_name: padloc
           db_user: padloc
--- a/.do/deploy.template.yaml
+++ b/.do/deploy.template.yaml
@@ -15,7 +15,7 @@ spec:
           build_command: npm ci
           environment_slug: node-js
           git:
-              branch: feature-digitalocean-one-click-install
+              branch: main
               repo_clone_url: "https://github.com/padloc/padloc"
           http_port: 3000
           instance_count: 1
@@ -76,7 +76,7 @@ spec:
           catchall_document: index.html
           environment_slug: node-js
           git:
-              branch: feature-digitalocean-one-click-install
+              branch: main
               repo_clone_url: "https://github.com/padloc/padloc"
           output_dir: packages/pwa/dist
           routes:
--- a/.do/deploy.template.yaml
+++ b/.do/deploy.template.yaml
@@ -12,7 +12,7 @@ spec:
           value: ${pwa.PUBLIC_URL}
     services:
         - name: server
-          build_command: npm ci
+          build_command: echo $PL_DATA_POSTGRES_TLS_CAFILE_CONTENTS > $PL_DATA_POSTGRES_TLS_CAFILE && npm ci
           environment_slug: node-js
           git:
               branch: feature-digitalocean-one-click-install
@@ -56,10 +56,13 @@ spec:
                 value: ${db.PASSWORD}
               - key: PL_DATA_POSTGRES_TLS
                 scope: RUN_TIME
-                value: "false"
+                value: "true"
               - key: PL_DATA_POSTGRES_TLS_CAFILE
-                scope: RUN_TIME
-                value: ""
+                scope: RUN_AND_BUILD_TIME
+                value: "db-ca.crt"
+              - key: PL_DATA_POSTGRES_TLS_CAFILE_CONTENTS
+                scope: RUN_AND_BUILD_TIME
+                value: ${db.CA_CERT}
               - key: PL_DATA_POSTGRES_TLS_REJECT_UNAUTHORIZED
                 scope: RUN_TIME
                 value: "false"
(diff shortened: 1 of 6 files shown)


BrunoBernardino:
@MaKleSoft I'm thinking of adding an env var to disable CSP, since that makes the initial setup harder and isn't necessarily increasing the security in those situations. It causes a problem using this as well, so I'd do it as part of this PR. What do you think?

MaKleSoft:
> I'm thinking of adding an env var to disable CSP, since that makes the initial setup harder and isn't necessarily increasing the security in those situations. It causes a problem using this as well, so I'd do it as part of this PR. What do you think?

Sure, why not! As long as it's optional.

.do/deploy.template.yaml
```diff
@@ -1,81 +1,49 @@
...
+          scope: RUN_AND_BUILD_TIME
+          value: ${server.PUBLIC_URL}
+        - key: PL_PWA_URL
+          scope: RUN_AND_BUILD_TIME
+          value: ${pwa.PUBLIC_URL}
+    services:
+        - build_command: npm ci
+          environment_slug: node-js
+          git:
+              branch: feature-digitalocean-one-click-install
```
BrunoBernardino, line 26: This needs to be reverted to `main` before merging this, but it's necessary to keep testing.

.do/deploy.template.yaml
```diff
@@ -1,81 +1,49 @@
...
+          health_check:
+              http_path: /healthcheck
+          run_command: cd packages/server && npm start
+          source_dir: /
+    static_sites:
+        - build_command: npm ci && cd packages/pwa && npm run build
+          catchall_document: index.html
+          environment_slug: node-js
+          git:
+              branch: feature-digitalocean-one-click-install
```
BrunoBernardino, line 43: This needs to be reverted to `main` before merging this, but it's necessary to keep testing.

BrunoBernardino:
The app now installs and loads, but doesn't get through the verification code (it validates, but keeps requesting it). I'll need to investigate to see what's wrong/missing, but there was no error in the logs. I suspect something with the automatically-built URL env vars (including/not-including a trailing slash, for example). Thoughts welcome if you have any @MaKleSoft

BrunoBernardino:
@MaKleSoft I'm still getting an issue about the verification code not matching: https://github.com/padloc/padloc/blob/98a73dc43aee9c40347cc25e6071801c3c4c7bc6/packages/core/src/auth/email.ts#L61-L73

Do you think it can be something to do with the fact we're using `leveldb` here? I'm not 100% sure if the App Platform has the instances with the disks in write mode. If you're not sure I'll try setting it up to a `postgres` db, but that's unnecessarily expensive for small/personal use. Any other ideas you have will help.

MaKleSoft:
> I'm not 100% sure if the App Platform has the instances with the disks in write mode.

Yep, I think that's exactly it. AFAIK app platform "services" are all supposed to be stateless, so we either need to mount a "Volume" or provision a DB.

BrunoBernardino:
Alright, seems I'll have to provision a DB because [they don't support "volumes" yet](https://www.digitalocean.com/community/questions/can-app-platform-persist-data-on-a-volume-between-releases), and an object storage won't really work, here.

BrunoBernardino:
@MaKleSoft I'm leaving the `branch` set to `feature-digitalocean-one-click-install` so you can test, if you want.

packages/server/src/storage/postgres.ts
```diff
@@ -73,9 +76,10 @@ export class PostgresStorage implements Storage {
     private _ensuredTables = new Map<string, Promise<void>>();
 
     constructor(public config: PostgresConfig) {
-        const { host, user, password, port, database, tls, tlsCAFile, tlsRejectUnauthorized } = config;
+        const { host, user, password, port, database, tls, tlsCAFile, tlsCAFileContents, tlsRejectUnauthorized } =
+            config;
         const tlsCAFilePath = tlsCAFile && resolve(process.cwd(), tlsCAFile);
-        const ca = tlsCAFilePath && readFileSync(tlsCAFilePath).toString();
+        const ca = (tlsCAFilePath && readFileSync(tlsCAFilePath).toString()) || tlsCAFileContents;
```
MaKleSoft, line 82: I think `tlsCAFileContents` should take precedence of `tsCAFilePath`, so:

```suggestion
        const ca = tlsCAFileContents || (tlsCAFilePath && readFileSync(tlsCAFilePath).toString());
```

MaKleSoft:
Getting this error when clicking the button:

Screenshot 2022-11-29 at 14 37 36

BrunoBernardino:
@MaKleSoft are you using this URL: https://cloud.digitalocean.com/apps/new?repo=https://github.com/padloc/padloc/tree/feature-digitalocean-one-click-install ?

MaKleSoft (approved):
Alright, good stuff! Even set up a test instance and it works!

PR #641, opened 2022-12-06
Fix race condition related to routing

### LINKED ISSUE #638: Can't login when hosting with docker compose
I use the `docker-compose.yml` from https://github.com/padloc/padloc/blob/main/docs/examples/hosting/docker/basic/docker-compose.yml and being redirected to the signup page after entering the 6 digits verification.

MaKleSoft:
This should also fix #638


BrunoBernardino (approved):
Looks good. :shipit: !

PR #646, opened 2022-12-13
Fix dependency issue with Mozilla

BrunoBernardino:
They can't load from git, only URLs.

We had fixed the dependency, but not force-updated the lock resolution. Unfortunately a few other sub-sub-dependencies also updated.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "padloc",
-    "version": "4.1.0",
+    "version": "4.2.0",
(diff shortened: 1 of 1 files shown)


BrunoBernardino:
Production build running at https://github.com/padloc/padloc/actions/runs/3687686925

BrunoBernardino:
@MaKleSoft you can get the latest XPI from https://github.com/padloc/padloc/releases/tag/v4.2.0 or the GitHub action above.

PR #648, opened 2022-12-14
Upgrade min SDK for Android Google Play

BrunoBernardino:
I can't build this locally anymore, so I'll have to debug in GH Actions.


BrunoBernardino:
I'm trying to go through https://ionic.zendesk.com/hc/en-us/articles/7891143965975-Migrating-to-Cordova-Android-11 but not really done yet. I've already spent over 4h on this, so pausing for today.

BrunoBernardino:
It seems we won't be able to push 4.2.0 to the Play Store, but the APK we have still works well.

Since this is a fairly new problem (November 2022) we'll close this for now and hopefully by the time we need to push something again it'll have a better solution.

PR #650, opened 2022-12-18
Fix unknown location in login/failed login emails

BrunoBernardino:
Also fixes a couple of typos in those emails' subjects.

@MaKleSoft classical case of a "one-line fix" after a couple of hours of debugging... :facepalm:

### DIFF
--- a/packages/core/src/server.ts
+++ b/packages/core/src/server.ts
@@ -2165,6 +2165,7 @@ export class Server {
 
         try {
             context.device = req.device;
+            context.location = req.location;
             try {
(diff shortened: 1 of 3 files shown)


packages/core/src/server.ts
```diff
@@ -2165,6 +2165,7 @@ export class Server {
 
         try {
             context.device = req.device;
+            context.location = req.location;
```
BrunoBernardino, line 2168: We were setting the location only when authenticating, not before.

PR #651, opened 2022-12-18
Add a confirmation for recovery of account

### LINKED ISSUE #645: Add new confirmation dialog for password reset
There have been enough instances in the past few months of people accidentally resetting their password (without realizing that deletes all data) that we should add an additional confirmation prompt, making it clear the data will be gone and it'll be irreversible.

BrunoBernardino:
Basically make it clearer the data will be deleted.

Fixes #645

Also fix typo.

### DIFF
--- a/packages/app/src/elements/recover.ts
+++ b/packages/app/src/elements/recover.ts
@@ -4,7 +4,7 @@ import { app, router } from "../globals";
 import { StartForm } from "./start-form";
 import { Input } from "./input";
 import { Button } from "./button";
-import { alert, choose } from "../lib/dialog";
+import { alert, choose, confirm } from "../lib/dialog";
 import { passwordStrength } from "../lib/util";
 import { customElement, query, state } from "lit/decorators.js";
 import { html } from "lit";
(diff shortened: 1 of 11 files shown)


packages/app/src/elements/recover.ts
```diff
@@ -185,7 +185,18 @@ export class Recover extends StartForm {
             }
         }
 
-        return this._recover(email, password);
+        const doRecover = await confirm(
+            $l("Your data **WILL BE DELETED** and cannot be recovered!"),
```
MaKleSoft, line 189: Minor change in wording considering that shared vaults are not affected.

```suggestion
            $l("All data in your personal vault **WILL BE DELETED** and cannot be recovered!"),
```

packages/app/src/elements/recover.ts
```diff
@@ -185,7 +185,18 @@ export class Recover extends StartForm {
...
+            $l("Delete data and recover account"),
+            $l("Cancel"),
+            {
+                icon: "warning",
```
MaKleSoft, line 193:
```suggestion
                icon: "warning",
                type: "destructive",
```

MaKleSoft (changes requested):
Had some minor changes but looks good otherwise.

packages/app/src/elements/recover.ts
BrunoBernardino: While technically more correct, I think this becomes too wordy and prone to be ignored as the first warnings. Do you really prefer your version?

MaKleSoft: It's a bit more wordy, but I think it's fine, since they still have to click a big red button (`type: "destructive"` will turn the button red) with the words "Delete data" on it.

PR #654, opened 2022-12-26
Allow instant-lock on background for mobile devices

### LINKED ISSUE #558: Allow Auto-Lock to be instant when the app is "backgrounded"
When using a mobile device, it's not uncommon for people to want the app to be instantly locked when it's "backgrounded" (instead of exited). Right now that's not possible to achieve, as the minimum time to set is 1 minute.

BrunoBernardino:
Fixes #558


BrunoBernardino:
Here's an Android build for Production with this code (which I've used for testing): https://github.com/padloc/padloc/actions/runs/3780988995

MaKleSoft (approved):

PR #655, opened 2022-12-26
Increase password generator limits

### LINKED ISSUE #577: A suggestion regarding the password generator
I was wondering why you chose to generate passwords with a maximum of 50 characters? Is it possible to increase them to 128 characters like in the BitWarden generator? I use 128-character passwords on almost all services and would like to have this possibility in the Padloc generator too ✌️

BrunoBernardino:
Also introduce a toggle to capitalize a word from a passphrase.

Fixes #577

### DIFF
--- a/packages/app/src/elements/generator.ts
+++ b/packages/app/src/elements/generator.ts
@@ -1,4 +1,4 @@
-import { randomNumber, randomString, chars } from "@padloc/core/src/util";
+import { randomString, chars } from "@padloc/core/src/util";
 import { generatePassphrase, AVAILABLE_LANGUAGES } from "@padloc/core/src/diceware";
 import { translate as $l } from "@padloc/locale/src/translate";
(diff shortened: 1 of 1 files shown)


MaKleSoft:
What is the point of capitalising a random, single word? To increase entropy? To satisfy password constraints? Doesn't seem to be a very good solution for either.

BrunoBernardino:
> What is the point of capitalising a random, single word? To increase entropy? To satisfy password constraints? Doesn't seem to be a very good solution for either.

Both, really, but mostly the second. I frequently have to do this manually because a few systems/apps require a capital letter and don't recognize the passphrase without it as a good one (debatable, but still). It's off by default.

If you don't think that's enough, I could add one more toggle for adding a "word" (random set, but within the word separators) of numbers?

MaKleSoft (approved):

PR #676, opened 2023-01-31
Updated Polish translation and Fixed English template typos

lukaszslupski:


packages/locale/res/translations/_template.json
```diff
@@ -444,7 +444,7 @@
     ""
   ],
   [
-    "Allow Edititing",
+    "Allow Editing",
```
MaKleSoft, line 447: Thank you for fixing these typos! Unfortunately changing the translation key also requires changing in in the source code. Do you know how to do that?

MaKleSoft (changes requested):

packages/locale/res/translations/_template.json
lukaszslupski: I'm not a programmer and that's all I can do to help.

MaKleSoft: Ok, got it. I'll take care of that then :)

lukaszslupski: Could you please also add a Polish wordlist to the password generator?

MaKleSoft: > Could you please also add a Polish wordlist to the password generator?

Unfortunately that's not quite as straightforward, since for that we need a source for the top / most common ~ 10.000
 words from the Polish dictionary.

MaKleSoft:
Merged. Thanks for your contribution!

PR #681, opened 2023-02-11
Improved search

BrunoBernardino:
This makes it possible to search for things like "Google - Padloc" with a term like "google padloc" instead of "google - padloc".

@MaKleSoft I'd love to push this on a 4.2.1, but I know we haven't really discussed it. Let me know your thoughts!


BrunoBernardino:
The android build error seems to be caused by https://github.com/android-actions/setup-android/issues/378

BrunoBernardino:
Now cypress is failing in CI due to https://github.com/cypress-io/cypress/issues/19299 (though locally all is fine). Will fix that too.

BrunoBernardino:
Small proof I'm not crazy :sweat_smile: 

Screenshot from 2023-02-11 11-30-49

packages/app/src/elements/items-list.ts
```diff
@@ -52,7 +52,19 @@ function filterByString(fs: string, rec: VaultItem) {
...
+        return content.search(escapeRegex(fs.toLowerCase())) !== -1;
+    }
+
+    for (const searchTerm of searchTerms) {
+        if (!content.includes(searchTerm)) {
+            return false;
+        }
+    }
+
+    return true;
```
MaKleSoft, line 67: I think this should do the same thing in fewer lines (plus it's more readable):

```suggestion
    const searchTerms = escapeRegex(fs.toLowerCase()).split(" ").filter(Boolean);
    
    return searchTerms.every(searchTerm => content.includes(searchTerm));
```

MaKleSoft (changes requested):
I like the idea! Just had a suggestion on how to make it more terse/readable. Other than that it should be good to merge.

BrunoBernardino:
Now it's https://github.com/GitSquared/node-geolite2-redist/issues/26 ! I'll update `geolite2-redist`...

packages/app/src/elements/items-list.ts
BrunoBernardino: Ah, good one!

BrunoBernardino:
I give up. The server isn't booting in CI for some weird reason and I can't figure that out. I'll just disable e2e tests in CI for now, but they're passing locally. FYI @MaKleSoft

cypress.config.ts
```diff
@@ -0,0 +1,18 @@
+import { defineConfig } from "cypress";
```
BrunoBernardino, line 1: Cypress 10 introduced a different config and file names. These changes are related to that.

MaKleSoft (approved):

PR #728, opened 2023-10-23
Updated german translation

RoyalCrafter:
Added the missing phrases for the german translation

Sorry in advance if something is not working, but my computer crashed everytime i tried to install the dependencies so i was not able to run the test

Issue #248


MaKleSoft:
Merged! Thank you!

PR #734, opened 2023-12-02
Add Additional Potential LastPass Export Format

arimgibson:
Closes #732 

Checks for a new heading format of an exported LastPass CSV file while still supporting previous formats.

PR #735, opened 2023-12-02
Alter LastPass Parsing for New and Legacy Formats

### LINKED ISSUE #733: Incorrect LastPass CSV Parsing to Padloc Format
See related issue #732, which suggests LastPass has changed their export format.

When importing a LastPass export, the file is incorrectly parsed to Padloc's format, resulting in the LastPass item name as the Padloc tag name opposed to the Padloc item name.

image

The likely culprit is the `lpParseRow` function, which uses defined indexes to alter the row format into Padloc's format. Because of items LastPass's format re-arranging, these indexes are no longer current.

https://github.com/padloc/padloc/blob/6e3cf04fa5fc699c334a8ef85704e26a80f224eb/packages/app/src/lib/import.ts#L357-L391

I'd be happy to PR a fix for this issue but would appreciate input from package maintainers beforehand to find the most appropriate solution. If supporting the legacy format is required (which it probably should be), then a new parameter could be passed to `lpParseRow` to indicate which format the CSV follows, which is then used to set the indexes appropriately. This could be a boolean `legacyFormat`, or something else.

```ts
/*
 * Parses a single row in a LastPass CSV file. Apart from extracting the default fields, we also parse
 * the 'extra' column for 'special notes' and remove any special fields that are not needed outside of
 * LastPass
 */
async function lpParseRow(row: string[], legacyFormat: boolean): Promise<VaultItem> {
    let nameIndex, categoryIndex, urlIndex, usernameIndex, passwordIndex, notesIndex, totpIndex;

    if (legacyFormat) {
        nameIndex = 4;
        categoryIndex = 5;
        urlIndex = 0;
        usernameIndex = 1;
        passwordIndex = 2;
        notesIndex = 3;
        totpIndex = 7;
    } else {
        nameIndex = 5;
        categoryIndex = 6;
        urlIndex = 0;
        usernameIndex = 1;
        passwordIndex = 2;
        notesIndex = 4;
        totpIndex = 3;
    }

    let fields: Field[] = [
        new Field({ name: $l("Username"), value: row[usernameIndex], type: FieldType.Username }),
        new Field({ name: $l("Password"), value: row[passwordIndex], type: FieldType.Password }),
        new Field({ name: $l("URL"), value: row[urlIndex], type: FieldType.Url }),
    ];
    const notes = row[notesIndex];

    if (row[urlIndex] === "http://sn") {
        // The 'http://sn' url indicates that this line represents a 'secure note', which means
        // we'll have to parse the 'extra' column to retrieve the individual fields
        fields.push(...lpParseNotes(notes));
        // In case of 'secure notes' we don't want the url and NoteType field
        fields = fields.filter((f) => f.name != "url" && f.name != "NoteType");
    } else {
        // We've got a regular 'site' item, so the 'extra' column simply contains notes
        fields.push(new Field({ name: $l("Notes"), value: notes, type: FieldType.Note }));
    }

    if (row[totpIndex]) {
        fields.push(new Field({ name: $l("One-Time Password"), value: row[totpIndex], type: FieldType.Totp }));
    }

    const dir = row[categoryIndex];
    // Create a basic item using the standard fields
    return createVaultItem({ name: row[nameIndex], fields, tags: dir ? [dir] : [] });
}
```

arimgibson:
Closes #733

Allows for different CSV row parsing from a LastPass CSV file depending on the heading row format.

- add new boolean parameter `legacyFormat` to `lpParseRow` function
- alter `asLastPass` function to detect modern vs legacy format
- alter `asLastPass` function to provide `lpParseRow` with new `legacyFormat` parameter

Video shows importing process with correct item name imported in Padloc, along with the tag working correctly.

https://github.com/padloc/padloc/assets/43756494/07ac8a47-e392-4219-9681-f3ff000a4f7c

### DIFF
--- a/packages/app/src/lib/import.ts
+++ b/packages/app/src/lib/import.ts
@@ -354,14 +354,26 @@ function lpParseNotes(str: string): Field[] {
  * the 'extra' column for 'special notes' and remove any special fields that are not needed outside of
  * LastPass
  */
-async function lpParseRow(row: string[]): Promise<VaultItem> {
-    const nameIndex = 4;
-    const categoryIndex = 5;
-    const urlIndex = 0;
-    const usernameIndex = 1;
-    const passwordIndex = 2;
-    const notesIndex = 3;
-    const totpIndex = 7;
+async function lpParseRow(row: string[], legacyFormat: boolean): Promise<VaultItem> {
+    let nameIndex, categoryIndex, urlIndex, usernameIndex, passwordIndex, notesIndex, totpIndex;
+
+    if (legacyFormat) {
+        nameIndex = 4;
+        categoryIndex = 5;
+        urlIndex = 0;
+        usernameIndex = 1;
+        passwordIndex = 2;
+        notesIndex = 3;
+        totpIndex = 7;
+    } else {
+        nameIndex = 5;
+        categoryIndex = 6;
+        urlIndex = 0;
+        usernameIndex = 1;
+        passwordIndex = 2;
+        notesIndex = 4;
+        totpIndex = 3;
+    }
 
     let fields: Field[] = [
         new Field({ name: $l("Username"), value: row[usernameIndex], type: FieldType.Username }),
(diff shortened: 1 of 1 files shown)


packages/app/src/lib/import.ts
```diff
@@ -393,13 +405,15 @@ async function lpParseRow(row: string[]): Promise<VaultItem> {
 export async function asLastPass(file: File): Promise<VaultItem[]> {
     const data = await readFileAsText(file);
     const papa = await loadPapa();
+    const isLegacyFormat =
+        data.split("\n")[0].trim() === "url,username,password,totp,extra,name,grouping,fav" ? false : true;
```
MaKleSoft, line 409:
```suggestion
    const isLegacyFormat = data.split("\n")[0].trim() === "url,username,password,totp,extra,name,grouping,fav";
```

`data.split("\n")[0].trim() === "url,username,password,totp,extra,name,grouping,fav"` already evaluates to a boolean, so the ternary expression is redundant here.

MaKleSoft (changes requested):
Looks good! Just accept the one change suggestion that I had and add the changes from #734, and this should be good to merge.

packages/app/src/lib/import.ts
MaKleSoft: Ah, that makes sense. In that case let's keep it as-is then :)

MaKleSoft (approved):
All good to merge!

PR #761, opened 2024-09-03
Button item-view wait until save

Zahkthar:
When there are many passwords (1000+), the application becomes slow and quitting causes a rollback. The button now wait until save and start a sync in background (even if it cannot terminate, the localstorage will be pushed on the next app startup)


Zahkthar:
Note that it's only a UX upgrade. In a normal use case, the user will switch off the application just after changing his password, even though it is not yet saved in the application (and even less synced if there are many passwords in the vaults), which will cause a rollback.

